PR #175, opened 2019-05-07
[Oedax] greatly simplify bounding curves

dong77:


dong77:
@Brechtpd  @autumn84 

I've changed the curve design and how prices merge into the target price. check this out - file:///Users/d/Projects/loopring/protocols/docs/index.html

PR #180, opened 2019-05-07
[Oedax] disable and remove queue

dong77:


dong77:
@Brechtpd  to make it simpler, I disabled and completely removed the queue support for now.

PR #181, opened 2019-05-07
[Oedax] Settlement refactor

Brechtpd:
Probably still contains some issues.
Not sure yet how to best reward the withdrawer so this is not yet implemented.
Maybe we can even fine the owner for not withdrawing for it's users if he can set some kind of withdrawing fee some way.

### DIFF
--- a/packages/oedax_v1/contracts/impl/libauction/AuctionSettlement.sol
+++ b/packages/oedax_v1/contracts/impl/libauction/AuctionSettlement.sol
@@ -61,21 +61,21 @@ library AuctionSettlement
         // update state
         s.closeTime = s.startTime + i.duration;
         s.settlementTime = block.timestamp;
-        uint rebate = s.fees.creatorEtherStake;
+        uint rebatePaid = s.fees.creatorEtherStake;
 
         if (!i.isBounded) {
-            rebate /= 2;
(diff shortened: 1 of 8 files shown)


packages/oedax_v1/contracts/impl/libauction/AuctionSettlement.sol
```diff
@@ -89,58 +87,84 @@ library AuctionSettlement
...
-        if (s.bidToken != address(0x0)) {
-            payToken(
-                feeRecipient,
-                s.bidToken,
-                ERC20(s.bidToken).balanceOf(address(this))
-            );
+        IAuctionData.Account storage a = s.accounts[user];
+        if (a.withdrawFee > 0) {
+            // Reward the withdrawer
+            msg.sender.transfer(a.withdrawFee);
```
dong77: msg.sender.transfer(a.withdrawFee);

So this will transfer ether from where?

dong77:
I think we should drop the withdrawal fee, and require the owner to withdraw for all the users before he can get back his stake. We need to give a percentage of trading volume to the owner as a reward. What do you think?

In general, it's not a good user experience to ask each user to withdraw.

we can make the withdrawal smart by using gasleft() method.

PR #184, opened 2019-05-09
[Oedax] Settlement unification

Brechtpd:
@dong77 I removed `settlementTime`, which may or may not be correct because it changes the logic in `getAuctionStatus` slightly. Though I think once the auction has closed it will remain closed so we shouldn't need it there.

### DIFF
--- a/packages/oedax_v1/contracts/impl/libauction/AuctionStatus.sol
+++ b/packages/oedax_v1/contracts/impl/libauction/AuctionStatus.sol
@@ -61,51 +61,58 @@ library AuctionStatus
             i.askPrice = s.P;
             i.bidPrice = s.P;
 
-            if (s.settlementTime == 0) {
-                // price unbounded and not settled yet
-                i.duration = yToX(s, s.P); // the earliest end elapsed
+            // price unbounded
+            i.duration = yToX(s, s.P); // the earliest end elapsed
 
-                if (i.duration > elapsed) {
-                    // the auction is open
-                    i.askPrice = xToY(s, elapsed);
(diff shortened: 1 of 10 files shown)


packages/oedax_v1/contracts/impl/libauction/AuctionStatus.sol
```diff
@@ -61,53 +61,60 @@ library AuctionStatus
             i.askPrice = s.P;
             i.bidPrice = s.P;
 
-            if (s.settlementTime == 0) {
```
Brechtpd: Not sure if this can be removed.

packages/oedax_v1/contracts/impl/libauction/AuctionStatus.sol
```diff
@@ -61,53 +61,60 @@ library AuctionStatus
...
-                    i.bidAllowed = ~uint256(0); // = uint.MAX
-                }
+                i.askAllowed = ~uint256(0); // = uint.MAX
+                i.bidAllowed = ~uint256(0); // = uint.MAX
             }
         } else {
             i.askPrice = i.actualPrice;
             i.bidPrice = i.actualPrice;
 
-            if (s.settlementTime == 0) {
```
Brechtpd: Not sure if this check can be removed.

packages/oedax_v1/contracts/impl/libauction/AuctionStatus.sol
dong77: Maybe should remove it.

dong77:
Great

PR #198, opened 2019-05-27
update results after optimization

dong77:


dong77:
please correct the numbers if they are incorrect.

Brechtpd:
I updated the numbers in the branch with the new fee model in #199.
It would be a bit confusing updating the throughput numbers on the master branch without the the design/implementation that produces the better numbers.

PR #199, opened 2019-05-27
[Protocol3] Implemented 3.0 fee model

Brechtpd:


dong77:
@kongliangzhong make sure you review this and ask a lot of questions :)

packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -48,7 +48,7 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
     using ExchangeTokens        for ExchangeData.State;
     using ExchangeWithdrawals   for ExchangeData.State;
 
-    ExchangeData.State public state;
+    ExchangeData.State private state;
```
dong77: why this is private?

Brechtpd: When it's public solidity will automatically try to generate a getter for the field `ProtocolFeeData protocolFeeData`, and because it's a struct it cannot do it without having `pragma experimental ABIEncoderV2`. We could not use a `struct` or enable `ABIEncoderV2` but I thought there was no need to make it public because we have getters for the things we want to expose onchain.

dong77:
Thanks for the clarification.

PR #203, opened 2019-05-31
[Protocol3-DRAFT] Make it easy to support offchain data-availability (test will fail)

dong77:


dong77:
with this proposed implementation, the operator contract cannot make any withdrawal request, correct?

packages/loopring_v3/contracts/operator/IPFSBasedDataAvailabilityOperator.sol
```diff
@@ -0,0 +1,49 @@
...
+        external
+        returns (bytes32 merkleRootAfter)
+    {
+        require(data.length > 0, "empty ipfs multihash");
+        merkleRootAfter = IExchange(exchange).commitBlock(
+            blockType,
+            blockSize,
+            bytes("0")
+        );
+        ipfsHashMap[merkleRootAfter] = data;
```
Brechtpd: Is the IFPS hash in some way easily related to the data that's stored there? If so, maybe it's possible to input the IPFS hash into the circuit so the data can be validated to be correct.

dong77: Its really depends on how the file is formatted. I guess if this is achievable, the file is not guaranteed to be available on IPFS, so maybe we don't want to to that far.

packages/loopring_v3/contracts/operator/Operator.sol
```diff
@@ -0,0 +1,94 @@
...
+        )
+        external
+        payable
+        onlyOwner
+    {
+        IExchange(exchange).withdraw(token, amount);
+    }
+
+    /// @dev Withdrawal the tokens/ether held in this contract.
+    function withdrawTo(
```
Brechtpd: Maybe we should allow setting a recipient address on an Account. The recipient address can only be set by the account owner. This recipient address would get all tokens from all withdrawals from that point on.  I remember Dolomite asking for this once. Could also be useful to make the contract that distributes the tokens another contract so it can be easily uprgraded/...

dong77: will do. This PR is still very draft. Will have to think about it.

dong77:
@Brechtpd I'm having some hard time figure out a clean way to make a simple contract for Operator and the Ring-matcher. Seems a lot of calls to IExchange will have to be proxied. Do you think we should simply use a Proxy?

packages/loopring_v3/contracts/operator/Operator.sol
Brechtpd: We would need to add this to the exchange contract e.g. a `IExchange.setRecipient(address)` function that sets a `recipient` field in `Account`. If set, this recipient address would receive the ETH/tokens instead of the Account owner.

dong77:
Can we just change the withdrawal request to have an optional parameter called ‘recipient’? This will save us a new request type. 

发自我的 iPhone

Brechtpd:
> @Brechtpd I'm having some hard time figure out a clean way to make a simple contract for Operator and the Ring-matcher. Seems a lot of calls to IExchange will have to be proxied. Do you think we should simply use a Proxy?

The operator allows multiple operators to work together without trust. If that's not needed and a single address has full control over all Operator related things we could just check against `tx.origin` instead of `msg.sender`. This would allow `tx.origin` to use any intermediate contract for any function he likes, so e.g. only `commitBlock` needs to be implemented or even an existing contract can be reused for all exchanges. For a 'real' Operator contract we need still check against `msg.sender` to ensure everything passes through that contract.

Using `tx.origin` however does not guarantee in any way all blocks are passed through the contract, which I think makes it worse than the method used in this PR.

Brechtpd:
> Can we just change the withdrawal request to have an optional parameter called ‘recipient’? This will save us a new request type.

The method I proposed wouldn't introduce a new request type. All withdrawals after calling `IExchange.setRecipient(address)` would use the Account.recipient address instead of the Account.owner address. This is completely onchain. 

Doing a recipient address for every withdrawal independently would be more difficult and more expensive.

dong77:
I see, I like it.

Issue #139, opened 2019-04-14
[Protocol3] Same circuit with/without data-availability


Brechtpd:
Currently we will use different circuits, but it's possible to re-use the data-availability circuit, which is a bit more expensive (~10%) because it needs to hash the data-availability data. Once we have optimized the data-availability data the performance difference may not be significant any more. 

It would work like this. Instead of hashing all calldata at once, we hash the data that's actually needed on-chain and the data-availability data separately. Both hashes are then hashed together to create the single public input for the circuit. This allows an exchange that doesn't want to use data-availability to send just the hash on-chain. Exchanges that want to use data-availability will still send all the data and the hash is still calculated on-chain. The exchange contract simply also has a `commitBlock` function that takes the data-availability as a hash directly.

This may also be useful to reduce the cost of sending all this data from the Operator contract to the Exchange contract (not sure how expensive this is). It's up to the Operator contract anyway to ensure that the data-availability data can easily be extracted from the Ethereum block data.

dong77:
@Brechtpd This is done on the protocol level, isn't it?

Brechtpd:
Yes, it changes the input value of the circuit because the hash is calculated differently in the Exchange contract.

Brechtpd:
If the circuits get more efficient by using a cheaper hash function for the Merkle tree it can still make sense to use a different circuit without DA if it's still significantly cheaper (the DA data needs to be hashed using sha256 which is still expensive in circuit). Currently the difference is only 5%.

The same circuit can be used with and without onchain DA as follows:

```
function commitBlockWithDA(
  bytes memory data,
  bytes memory dataAvailabilityData)
{
  // The operator passes in the DA data and we hash it onchain, ensuring the DA is onchain
  bytes32 dataAvailabilityDataHash = sha256(dataAvailabilityData);
  commitBlockInternal(data, dataAvailabilityDataHash)
}

function commitBlockWithoutDA(
  bytes memory data,
  bytes32 dataAvailabilityDataHash)
{
  // The operator passes in the DA hash directly, the DA is not guaranteed to be onchain
  require(!S.onchainDataAvailability, "ONCHAIN_DA_CANNOT_BE_ENABLED");
  commitBlockInternal(data, dataAvailabilityDataHash)
}

function commitBlockInternal(
  bytes memory data,
  bytes32 dataAvailabilityDataHash)
{
  bytes32 dataHash = sha256(data);
[...]

PR #204, opened 2019-05-31
[Protocol3] Add basic support for offchain data-availability

dong77:


packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -54,10 +54,15 @@ library ExchangeBlocks
...
         uint16 blockSize,
-        bytes memory data
+        bytes memory data,
+        bytes memory offchainData
         )
         internal  // inline call
     {
+        if (S.onchainDataAvailability) {
+            require(offchainData.length == 0, "INVALID_OFFCHAIN_DATA");
+        }
```
dong77: when onchainDataAvailabile == false, we do not enforce offchainData.length > 0

packages/loopring_v3/contracts/test/Operator.sol
```diff
@@ -38,7 +38,7 @@ contract Operator {
         )
         external
     {
-        exchange.commitBlock(blockType, numElements, data);
+        exchange.commitBlock(blockType, numElements, data, bytes("0"));
```
dong77: not sure if bytes("0").length == 0.

Brechtpd:
Related to #139. Will need some minor modifications to the circuits.

dong77:
@Brechtpd I’m merging this PR and leave the modification of circuits to you.

PR #211, opened 2019-06-07
[Protocol3] Implemented skipping of expensive token withdrawals while distributing

### LINKED ISSUE #202: [Protocol3] Skip expensive token withdrawals while distributing
Both ETH transfers as well as ERC-20 token transfers can cost a lot of gas. This can be used as an attack because the withdrawals need to be automatically distributed by the operator. Withdrawals that are more expensive than expected need to be skipped while the operator is distributing. The user can always manually withdraw without any gas limit.

For ETH we currently use `address.transfer` which has a 2300 gas limit and throws if more gas is used. We need to do it in a way that doesn't throw but simply returns true/false so we can skip it.

For ERC-20 we need to set a gas limit on the token transfer.

Brechtpd:
Fixes #202


packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
```diff
@@ -283,44 +284,56 @@ library ExchangeWithdrawals
...
-                assembly {
-                    mstore(add(mWithdrawals, offset), data)
-                }
-                withdrawBlock.withdrawals = mWithdrawals;
+                success = transferTokens(
+                    S,
+                    recipient,
+                    tokenID,
+                    amount
+                );
```
dong77: should we create a wrapping method for code between line 289 and 302, then inside the method emit a TokenDistributionFailure event if the `success` is false?

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
```diff
@@ -283,44 +284,56 @@ library ExchangeWithdrawals
...
-            );
+                emit WithdrawalCompleted(
+                    accountID,
+                    tokenID,
+                    recipient,
+                    uint96(amount)
+                );
+            }
+        } else {
+            success = true;
```
dong77: should we also emit WithdrawalCompleted event if the amount is zero?

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
```diff
@@ -450,25 +471,66 @@ library ExchangeWithdrawals
...
-        address payable owner = address(uint160(accountOwner));
+        // Limit the amount of gas that can be used
+        uint gasLimit = ExchangeData.GAS_LIMIT_SEND_TOKENS();
+        return sendTokensInternal(S, to, tokenID, amount, gasLimit);
+    }
+
+    // The transfer is guaranteed to succeed using as much gas as needed,
+    // otherwise it throws. Always returns true.
+    // Works similar to address.transfer() for transfering ETH.
+    function transferTokens(
```
dong77: I would suggest to merge `sendTokens`, `transferTokens` and `sendTokensInternal ` as one function.

dong77:
I suggest adding a failure event for failed distributions.

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
Brechtpd: Merged the functions into one.

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
Brechtpd: WithdrawalCompleted is now emitted when the withdrawal data hasn't been reset yet by a previous successful withdrawal (either the account, token or amount is non-zero). Otherwise we don't have any actual information about the withdrawal onchain.

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
Brechtpd: Added a `WithdrawalFailed`, which can only be emitted while distributing. When a user manually withdraws the transaction will fail with `TRANSFER_FAILURE` when something goes wrong with the token transfer.

Issue #206, opened 2019-06-03
[Protocol3] Better maintenance mode implementation [including tests]


dong77:
The current implementation lacks some flexibility. If the operator purchased some down time, he has to wait for all time to be used before the DEX can restore to the normal mode. 

We should enable the DEX operator to exit the maintenance mode whenever they want, and all unused downtime can be used in the future. The cost per unit of downtime should be a parameter at the protocol level. 

The unit of the downtime should be probably a minute instead of a second.

Brechtpd:
Done In #212

PR #212, opened 2019-06-08
[Protocol3] Improved maintenance mode

Brechtpd:
Implements #206.

Tests still need to be updated.


packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
```diff
@@ -96,27 +96,59 @@ library ExchangeAdmins
...
         require(!S.isShutdown(), "INVALID_MODE");
+        require(durationMinutes > 0, "INVALID_DURATION");
+
+        // Purchased downtime from a previous maintenance period or a previous call
+        // to startOrContinueMaintenanceMode can be re-used, so we need to calculate
+        // how many additional minutes we need to purchase
+        uint numMinutesLeft = S.getNumDowntimeMinutesLeft();
+        if (numMinutesLeft < durationMinutes) {
+            uint numMinutesToPurchase = durationMinutes.sub(numMinutesLeft);
+            uint costLRC = getDowntimeCostLRC(S, numMinutesToPurchase);
```
dong77: can this be written as:
`uint costLRC = S.getDowntimeCostLRC(numMinutesToPurchase);`

packages/loopring_v3/contracts/impl/libexchange/ExchangeMode.sol
```diff
@@ -117,4 +127,24 @@ library ExchangeMode
...
+        )
+        internal // inline call
+        view
+        returns (uint)
+    {
+        if (S.downtimeStart == 0) {
+            return S.numDowntimeMinutes;
+        } else {
+            // Calculate how long (in minutes) the exchange is in maintenance
+            uint numDowntimeMinutesUsed = now.sub(S.downtimeStart) / 60;
```
dong77: use block.timestamp instead of now

dong77:
@Brechtpd can we add a totalDowntime per exchange. Maybe also have a view method called

`getTotalDowntimeBips` to return `totalDowntime * 100 / totalExchangeLifetime`, this will be a good indicator.

In future versions, we may want to give a different downtime price per minute, for example:

```javascript
function getDowntimePricePerMinute() {
  if (getTotalDowntimeBips < 100) {
    return downtimePriceLRCPerMinute
  } else if (getTotalDowntimeBips < 500) {
    return downtimePriceLRCPerMinute * 2;
  } else {
   return downtimePriceLRCPerMinute * 5;
  }
}
```

packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
Brechtpd: That doesn't seem possible because it's a function in the same library. If you do something like this: 

```
library ExchangeAdmins
{
    using ExchangeAdmins for ExchangeData.State;
    ...
}
```

then this strangely compiles, but cannot be deployed because it references itself.

packages/loopring_v3/contracts/impl/libexchange/ExchangeMode.sol
Brechtpd: Any reason for this? `now` is exactly the same as `block.timestamp`:
https://solidity.readthedocs.io/en/develop/units-and-global-variables.html?highlight=block.timestamp#block-and-transaction-properties

All other code uses `now`.

Brechtpd:
> @Brechtpd can we add a totalDowntime per exchange. Maybe also have a view method called
> 
> `getTotalDowntimeBips` to return `totalDowntime * 100 / totalExchangeLifetime`, this will be a good indicator of the exchange's quality of service.

Good idea. I exposed `getTotalTimeInMaintenanceSeconds` and `getExchangeCreationTimestamp` which allows doing this.

packages/loopring_v3/contracts/impl/libexchange/ExchangeMode.sol
dong77: Ok, there is a lint warning we should use `block.timestamp` instead of `now`. But if you don't see it, it means we disabled that rule. Then stick to `now` for consistence then.

packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
dong77: Fair enough.

dong77:
LGTM.

dong77:
please merge once tests pass.

Brechtpd:
> After a second thought, maybe we should still use constant price for the downtime. We should not set up any barrier for a DEX to operate? @Brechtpd ?

Maintenance mode needs to be expensive because users cannot withdraw anything in maintenance. If it's cheap the exchange can hold user funds hostage for a long time. Making maintenance mode even more expensive if it's used a lot could make sense, other parameters like total user funds stored in the exchange, ... could also be used.

PR #218, opened 2019-06-19
[Protocol3] Implementation more admin-functions for UserStakingPool

dong77:


dong77:
I'll merge this into master. But I want to get user staking contracts implemented differently. So there will be a big refactor.

PR #220, opened 2019-06-19
add a tool for generating circuit keys

kongliangzhong:
A small tool for generating vk, pk files for circuits
Now works with blockType of deposit and onchainWithdrawal, 
still has some issues with offchainWithdrawal and trade keys generating.

### DIFF
--- a/packages/loopring_v3/generate_keys.py
+++ b/packages/loopring_v3/generate_keys.py
@@ -1,0 +1,33 @@
+from create_block import *
+import subprocess
+
+class Struct(object): pass
+
+def generate_keys(blockType, numElements, onchainDataAvailability):
+    block = Struct()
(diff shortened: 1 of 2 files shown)


kongliangzhong:
Based on protocols tags/protocol3.0.1-snapshot and lightcone2 tags/circuit-3.0-beta

Brechtpd:
@kongliangzhong There's no need to create complete valid blocks to be able to create the keys, just the meta data is enough. You can create a json file like this for any block permutation you want the keys for:

```
{
    "blockType": 0,
    "numElements": 8,
    "onchainDataAvailability": true
}
```

and then create the keys like this:

```
build/circuit/dex_circuit -createkeys [filename].json
```

kongliangzhong:
@Brechtpd , got it , thanks.

PR #221, opened 2019-06-19
[Protocol3] Calldata compression

Brechtpd:


dong77:
@Brechtpd I'll review this ASAP. In general, what's the performance gain after this PR?

packages/loopring_v3/contracts/iface/IBlockVerifier.sol
```diff
@@ -44,11 +46,13 @@ contract IBlockVerifier
...
+    /// @param blockVersion The block version (i.e. which circuit version needs to be used)
     /// @return True if the block can be verified, false otherwise
     function canVerify(
-        uint8 blockType,
-        bool onchainDataAvailability,
-        uint16 blockSize
+        uint8  blockType,
+        bool   onchainDataAvailability,
+        uint16 blockSize,
+        uint8  blockVersion
```
dong77: Can we use old number for data availability and even number for no-data availability?

dong77: Or use 0 for no-data availability and non zero for data availability with different compression algorithms

packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -367,16 +367,65 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
...
+        bytes calldata /*data*/,
         bytes calldata offchainData
         )
         external
         onlyOperator
         nonReentrant
     {
-        state.commitBlock(blockType, blockSize, data, offchainData);
+        // Decompress the data here so we can extract the data directly from calldata
+        bytes4 selector = IDecompressor(0x0).decompress.selector;
```
dong77: Why 0x0?

Brechtpd: The address doesn't matter, it just needs to be something valid so the function selector can be calculated using solidity (it could also be calculated manually, but this is more clear). The function selector value is independent of the address of the contract.

packages/loopring_v3/contracts/iface/IBlockVerifier.sol
Brechtpd: Not sure, this might make things more complicated than needed.

The versions can also be used to deprecate a block version. So for example a new block version can be added in IBlockVerifier and the old version can only be used for another ~1 week before no new blocks can be committed with that version (but they can still be verified after that deadline) . This allows us to update the circuits when needed without any interruption.

Brechtpd:
>  In general, what's the performance gain after this PR?

I can't give a good estimate, it depends too much on the data and the compressor. Though because of XOR on the ring header data (ring-matcher account, fee amount and fee token) this should turn 5 of the current total of 25 bytes for the complete DA for a trade to zero most of the time, which is a reduction of 20% just by doing that. Al other savings depends a lot on what rings are settled (like mostly the same trading pair, ...) and if there was any padding to fill the block.

PR #224, opened 2019-06-23
[Protocol3] reimplement user staking pool

dong77:


packages/loopring_v3/contracts/iface/IProtocolFeeManager.sol
```diff
@@ -0,0 +1,108 @@
...
+            uint remainingDev,
+            uint remainingReward
+        );
+
+    /// @dev Set Oedax address, only callable by the owner.
+    /// @param _oedaxAddress The address of Oedax contract.
+    function setOedax(address _oedaxAddress) external;
+
+    /// @dev Permanently disallow owner to withdraw non-LRC protocol fees, only callable by the owner.
+    function disableOwnerWithdrawaldisableOwnerWithdrawal() external;
```
Brechtpd: Should be just disableOwnerWithdrawal

packages/loopring_v3/contracts/impl/ProtocolFeeManager.sol
```diff
@@ -183,48 +196,53 @@ contract ProtocolFeeManager is IProtocolFeeManager, Claimable
...
-        //     tokenS,
-        //     auctionAddr
-        // );
+
+        IOedax oedax = IOedax(oedaxAddress);
+        uint ethStake = oedax.creatorEtherStake();
+
+        auctionAddr = oedax.createAuction.value(ethStake)(
+            token,  // askToken
+            tokenB, // bidToken
```
Brechtpd: What if token == tokenB?

packages/loopring_v3/contracts/impl/ProtocolFeeManager.sol
```diff
@@ -183,48 +196,53 @@ contract ProtocolFeeManager is IProtocolFeeManager, Claimable
...
+            minBidAmount,
+            P,
+            S,
+            M,
+            T,
+            T * 2
+        );
+
+        if (token == address(0)) {
+            auctionAddr.transfer(amount);
```
Brechtpd: `address.transfer` has a gas limit of 2300 gas which is probably too low to do the additional logic in the default function. Transferring the ETH like this doesn't have that gas limit:
```
(success, ) = auctionAddr.call.value(amount)("");
```

packages/loopring_v3/contracts/impl/ProtocolFeeManager.sol
dong77: when token == tokenB then the oedax auction will fail. This is checked inside the oedax contract.

dong77: I added another check.

packages/loopring_v3/contracts/impl/ProtocolFeeManager.sol
dong77: great.

Issue #231, opened 2019-06-26
[Protocol3] Mint reputation token for active addresses


dong77:
A certain percentage of protocol fee (DAO_PERCENTAGE) will be sent to our DAO smart contract. These LRC will fund our DAO and will be spent by approved proposals. Loopring stakeholders will need to vote for or against proposals. The overall weight for a vote is calculated by using `amount * reputation` where `amount` is the LRC he staked for or against the proposal, and the `reputation` is the user's non-transferrable reputation balance.

We will distribute reputations in two ways. Initially, all reputations will be distributed to the team and some token holders or stakers; then going forward, the reputations can be earned either by participating in our DAO or using our protocol.

**We need to figure out how to mint new reputations for addresses that use our protocol.** Lets discuss the topic here.

dong77:
Some user behaviors may earn reputations:

- user deposits
- user withdrawal with Merkle proofs
- operator generates block/proof
- owner buys downtime (for the first time?)
- user staking LRC in our staking pool
- DEX owner staking LRC
- market maker staking LRC

Some actions will trigger reputations to be destroyed:
- owner buys downtime (except the first time?)
- DEX goes to withdrawal mode
- DEX reverts blocks

dong77:
One desirable feature is the automatic decay of reputation balance — after each predefined *half-life* period, for example 4 years, each user balance will half. This will creates some dynamics in the distribution of reputation and make sure the old whales will eventually become smaller if they are not actively participating the DAO to earn more reputations.

dong77:
The most important roles in Loopring's ecosystem would be DEX owners and market makers. For roles, I fee that proposal-based reputation issuance should be good enough. Issuing tokens in the protocol's smart contract doesn't seem to be flexible and is costly.

PR #232, opened 2019-06-26
[Protocol3] Add MintableNonTransferableERC20Token and LOOP token

dong77:
Add MintableNonTransferableERC20Token for future Loopring DAO reputation token


dong77:
Related to #231

PR #233, opened 2019-06-26
[Protocol3] Add RoleBasedOperators.sol for DAO management

dong77:
For our DAO, the reputation token should be minted/destroyed over time by multiple contracts including DEX instances, the Loopring contract, UserStakingPools, as well as the DAO itself. This contract introduced a way to allow multiple authorized addresses, or operators, to mint/destroy LOOP tokens.

### DIFF
--- a/packages/loopring_v3/contracts/lib/RoleBasedOperators.sol
+++ b/packages/loopring_v3/contracts/lib/RoleBasedOperators.sol
@@ -1,0 +1,157 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


packages/loopring_v3/contracts/lib/RoleBasedOperators.sol
```diff
@@ -0,0 +1,159 @@
...
+                require(hasAnyRoles(msg.sender, ROOT), "UNAUTHORIZED");
+            }
+
+            operatorAddresses.push(who);
+            operator.id = operatorAddresses.length;
+            operator.roles = roles;
+        } else {
+            require(operatorMap[who].roles != roles, "ALREADY_EXIST");
+            require(hasAnyRoles(msg.sender, ROOT), "UNAUTHORIZED");
+            operatorMap[who].roles = roles;
```
Brechtpd: I think operatorAddresses needs to be updated here as well when needed

PR #241, opened 2019-06-28
[Protocol3] Misc small improvements + More tests

Brechtpd:


packages/loopring_v3/contracts/iface/ILoopringV3.sol
```diff
@@ -230,16 +230,30 @@ contract ILoopringV3
...
 
+    /// @dev Withdraw all protocol fees to the designated address.
+    ///      Fees first need to be withdrawn from an exchange using
+    ///      withdrawProtocolFeesFromExchange.
+    ///
+    ///      This function can only be called by the owner of this contract.
+    ///
+    /// @param  token The address of the token. Use 0x0 for Ether.
+    /// @param  recipient The address to receive the tokens.
+    function withdrawProtocolFees(
```
dong77: @Brechtpd , in the future, we'll have a [ProtocolFeeManager](https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/iface/IProtocolFeeManager.sol), can we actually have a method `setProtocolFeeManager(address pfm)` and then the owner can only withdraw to that address? If the address is 0x0 (default), `withdrawProtocolFees` will be disabled.

dong77: @Brechtpd , in the future, we'll have a [ProtocolFeeManager](https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/iface/IProtocolFeeManager.sol), can we actually have a method `setProtocolFeeManager(address pfm)` and then the owner can only withdraw to that address? If the address is 0x0 (default), `withdrawProtocolFees` will be disabled.

Once the ProtocolFeeManager address is not 0x0, the fees should be sent **directly** to that address instead of being held by the ILoopring contract.

The idea is simple: send all protocol fees to ProtocolFeeManager and let all fee token management (burn/staking reward/DAO) activities occor in ProtocolFeeManager, not here.

packages/loopring_v3/contracts/iface/ILoopringV3.sol
```diff
@@ -230,16 +230,30 @@ contract ILoopringV3
...
+        address token,
```
dong77: token => tokenAddress

packages/loopring_v3/contracts/lib/AddressUtil.sol
```diff
@@ -17,17 +17,62 @@
...
+        )
+        internal
+        returns (bool success)
+    {
+        if (amount == 0) {
+            return true;
+        }
+        address payable recipient = to.toPayable();
+        (success, ) = recipient.call.value(amount).gas(gasLimit)("");
+        return success;
```
dong77: this `return`-line can be deleted.

packages/loopring_v3/contracts/test/TransferContract.sol
```diff
@@ -0,0 +1,93 @@
...
+    function safeTransferFromWithGasLimit(
+        address token,
+        address from,
+        address to,
+        uint256 value,
+        uint    gasLimit
+        )
+        external
+    {
+        require(token.safeTransferFromWithGasLimit(from, to, value, gasLimit), "TRANSFER_FAILURE");
```
dong77: break in to two lines.

packages/loopring_v3/contracts/iface/ILoopringV3.sol
dong77: Update: maybe we should enforce the pfm address should be non-zero but upgradable. By doing this, we don't have to support `withdrawProtocolFees` method at all.

Brechtpd: @dong77 The exchange always withdraws to the owner of the account. The owner of the protocol fee account is currently the Loopring contract so all withdrawals go to the Loopring contract. Currently only the owner of an account can request an on-chain withdrawal.

I tried to keep it simple and without any special cases in Exchange. To be able to send the fees directly to the ProtocolFeeManager we can:
- Keep the Loopring contract as the owner, but add support for the account owner to set a recipient address that will be used for the withdrawals. We then allow the Loopring contract to set the recipient address to the ProtocolFeeManager and this address can be updated when needed if we want to allow that (updating it would need to be done for every exchange independently).
- Set the ProtocolFeeManager as the owner of the account. Ideally all withdrawal requests for the exchanges go through the ProtocolFeeManager but it's also possible to add a special case in Exchange that anyone can do an onchain withdrawal for the fees account. To be able to change the ProtocolFeeManager address we would need to make accounts transferable to a different owner address (updating it would need to be done for every exchange independently).
- Just add a special case when withdrawing from the protocol fees account to withdraw to Loopring.ProtocolFeeManager() address.

autumn84: @Brecht, what I meant is the recipient of the **protocol fees** which should have nothing to do with the owner of any account I guess.

Currently, all the protocol fees are sent to the ILoopring contract, which I think could be changed such that all protocol fees. can be sent to `ILoopring(loopring_address).pfm()` the pfm value should be different from the owner of ILoopring.

Brechtpd: Seems like special casing this for the protocol fee account in the Exchange contract is the only way this can be achieved so I'll do that.

Brechtpd: Done. Please check the updated code.

dong77:
LGTM

PR #256, opened 2019-07-06
[Protocol3] Switch hash function from MiMC/e7r91 to Poseidon

Brechtpd:
Currently reduces the number of constraints from 350,000 constraints/trade to 130,000 constraints/trade.

Uses a Quadtree instead of a Binary tree for the Merkle tree because it's more efficient with Poseidon.

130,000 constraints/trade:
- Merkle Tree updating: 60,000 constraints
- Ring signatures (3 signatures: ring-matcher + 2 dual-author): 27,000 constraints
- Order signatures (2 signatures): 20,000 constraints
- Data availability hashing: 12,500 constraints 
- Ring settlement logic: 14,000 constraints

Probably still possible:
- ~10,000 constraints can probably be be saved on signatures by switching from pedersen to poseidon
- ~10,000 constraints can be be saved on Merkle tree updates by using the minimal parameters that give 128bit security. Currently using more conservative parameters.

Does not contain an EVM version of Poseidon yet needed for withdrawal mode.


dong77:
Is it possible to configure the security level of an DEX between 128bits and 256bits? Not saying that we should offer it now.

dong77:
Do we need to try Octotree, not sure about the term, but I meant each parent has 8 children.

Brechtpd:
> Is it possible to configure the security level of an DEX between 128bits and 256bits? Not saying that we should offer it now.

I don't think we can ensure 256bit for everything. `sha256` for example only has 128bit collision resistance but has 256bit preimage resistance. For Poseidon both are the same. I think 128bit is more than good enough.

> Do we need to try Octotree, not sure about the term, but I meant each parent has 8 children.

This is very Poseidon specific. The optimal tree for Poseidon would be nodes with 5 children (t == 6 which gives us 5 inputs of 256bits). Because this is not a binary tree this complicates things a bit which may  make this kind of tree more expensive than a bit less optimal tree with 4 children, but I'm not sure about this at all.

dong77:
Can you elaborate on this “Does not contain an EVM version of Poseidon yet needed for withdrawal mode.”

Brechtpd:
> Can you elaborate on this “Does not contain an EVM version of Poseidon yet needed for withdrawal mode.”

In withdrawal mode we do the Merkle tree hashing on-chain to check agains the root so people can withdraw with just a Merkle proof. See [here](https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/impl/libexchange/ExchangeBalances.sol). This contains calls to MiMC which now needs to be Poseidon.

Brechtpd:
- Added EVM version of Poseidon to allow withdrawing from the Merkle tree onchain. This was already quite expensive with MiMC (~1.2M gas/withdrawal if I remember correctly). The cost is now ~1.4M gas/withdrawal.
- Added EdDSA version using Poseidon for hashing. Reduces the number of constraints per signature by about 4000/signature. Pedersen is not used anywhere any more. All hasing is done with Poseidon (expect the public data, we need to use sha256 because we have to hash a lot of data onchain and sha256 is cheap onchain).
- Uses minimal Poseidon parameters that give 128bit security.
- I tried 5 children/node for the Merkle tree. This could _very_ slightly reduce the number of constraints for the Merkle tree. However, because it's not a power-of-2 it doesn't optimally use all available bits we use for the data-availability, which is the main bottleneck. E.g. a tree like this with a depth of 4 needs `log2(5**4) = 9.28 bits`. A Quadtree is the best option.
- The Merkle tree uses a single Poseidon variant (t5/f6/p52) to make things easier (so a single EVM implementation is needed). This only increases the number of constraints very slightly.
- Already removed dual authoring because I needed to update the signatures everywhere. #262
- Removed the ring-matcher signature already that was needed for every ring. Because the operator wil replace the ring-matcher in the new fee model no signature/ring is needed. The operator will need to pay the protocol fee/rebates so the operator will need to authorize this somehow. A simple signature of e.g. [blockIdx, merkle root] for the complete block is sufficient for this. The ring-matcher is however currently still used until the new fee model is completely worked out.

Now: 85,000 constraints/trade:
- Merkle tree updating: 50,000 constraints
- Order signatures (2 signatures): 12,000 constraints
- Data availability hashing: 12,500 constraints
- Ring settlement logic: 12,500 constraints

dong77:
@brecht, please merge into a new beta3 branch.

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
```diff
@@ -58,7 +58,7 @@ library ExchangeGenesis
         S.lrcAddress = loopring.lrcAddress();
 
         ExchangeData.Block memory genesisBlock = ExchangeData.Block(
-            0x06ea7e01611a784ff676387ee0a6f58933eb184d8a2ff765608488e7e8da76d3,
+            0x2b4827daf74c0ab30deb68b1c337dec40579bb3ff45ce9478288e1a2b83a3a01,
```
dong77: should we make this a constant variable?

Brechtpd:
> Also, what about the gas cost per block submission and verification, especially for the settlement blocks.

These are unchanged in this PR.
- `commitBlock` (without DA calldata cost): 150,000 gas. The cost is mainly storing block data onchain like the merkle root.
- `verifyBlock`: 500,000 gas. This will be reduced to 200,000 gas/block with batched proof verification.

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
Brechtpd: Done

Brechtpd:
There was a simple merge conflict with `master` because another PR got merged (in which `tslint.json` got deleted). Unfortunately this currently 'dismisses' all approvals, so please approve again because nothing has changed. :)

PR #258, opened 2019-07-10
[Protocol3] add ProtocolFeesUpdated event

kongliangzhong:


packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -407,6 +414,12 @@ library ExchangeBlocks
                 S.onchainDataAvailability
             );
             data.timestamp = uint32(now);
+            emit ProtocolFeeChanged(
```
dong77: emit events should not happen in a validation method. 
Events should be emitted only when the actual change happens.

PR #259, opened 2019-07-10
[PROTOCOL3] bitarray encode update

kongliangzhong:
Update the BN encoding method, the problem of the current encoding is: 
    const bn0 = new BN("123456789");
    const numBitStr = bn0.toString(2);

    const bitarray1 = new bitarray.BitArray();
    bitarray1.addBN(bn0, bn0.toString(2).length);
    const bitarrayBitStr = bitarray1.getBits().join("");

numBitStr is not equal to numBitStr. This is not convenient for testing.

Furthermore, can we remove the `length` parameter of all functions?

### DIFF
--- a/packages/loopring_v3/test/bitarray.ts
+++ b/packages/loopring_v3/test/bitarray.ts
@@ -7,12 +7,19 @@ export class BitArray {
     this.data = initialData;
   }
 
+  public addBNWithoutLength(value: BN) {
+    const strBits = Array.from(value.toString(2));
+    const numBits = strBits.map(i => Number(i));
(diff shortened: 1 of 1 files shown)


Brechtpd:
We really need a fixed length because the circuits will always expect the data to take 'length' number of bits. This is important because the circuit will generate the hash always in the same way with the exact same message length, so the data needs to padded with 0s when needed. `BN.toString(2)` won't do this padding with 0s and will have a dynamic length.

kongliangzhong:
Ok, I understand. This PR is closed now.

PR #261, opened 2019-07-10
Update ParticipationHelper.sol

coreycaplan3:
Dolomite needs to give away the margin, always, in order to synchronize the way our matching engine matches trades and Loopring settles them.


kongliangzhong:
@coreycaplan3 , Looks good to me, but you need to test this yourself.

coreycaplan3:
yeah we tested this in our backend :)

PR #275, opened 2019-07-15
[Protocol2] Made the two updates necessary for Dolomite's launch & re-deployment

coreycaplan3:
This contains the two changes we need to launch Dolomite on v2 of Loopring:
1. disabling margin splitting
2. Emitting an event when there is a `waiveFeePercentage < 0`

After talking to @Brechtpd , we decided the easiest way to go about this, was to emit an event when there is a negative `waiveFeePercentage`. It was easiest to add the event to the `Library` and change the function declarations to no longer be `view`s.

This was tested via Ganache & Dolomite's backend in Scala. If the proposed changes are okay, can you redeploy this contract? Once it's deployed, Dolomite will be able to settle trades on the mainnet and we'll be ready to launch :)

### DIFF
--- a/packages/loopring_v2/deployment-source/v2.2-dolomite/RingSubmitter_flat.sol
+++ b/packages/loopring_v2/deployment-source/v2.2-dolomite/RingSubmitter_flat.sol
@@ -1064,6 +1064,18 @@ contract IRingSubmitter {
     event InvalidRing(
         bytes32 _ringHash
     );
+    
+    /// @dev   Event emitted when fee rebates are distributed (waiveFeePercentage < 0)
+    ///         _ringHash   The hash of the ring whose order(s) will receive the rebate
+    ///         _orderHash  The hash of the order that will receive the rebate
+    ///         _feeToken   The address of the token that will be paid to the _orderHash's owner
+    ///         _feeAmount  The amount to be paid to the owner
+    event DistributeFeeRebate(
+        bytes32 indexed _ringHash,
+        bytes32 indexed _orderHash,
+        address         _feeToken,
+        uint            _feeAmount
+    );
 
     /// @dev   Submit order-rings for validation and settlement.
     /// @param data Packed data of all rings.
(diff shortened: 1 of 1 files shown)


Brechtpd:
Looks good to me.

You may need to add the `DistributeFeeRebate` event definition to `IRingSubmitter` so it's defined in the contract as well as in the library, otherwise (if I remember correctly) you can't listen to the event on the RingSubmitter contract because it's not part of the contract ABI. Though it  seems you didn't have any problems with this in your testing?

coreycaplan3:
I had no issues with testing at all. It successfully parsed the events when they were emitted in the logs, without fail. I should probably, for vanity purposes, double-declare it in the interface so it's easier to read?

coreycaplan3:
Considering that the events are traditionally declared in the interface for stylistic purposes

coreycaplan3:
Never mind, I'm getting a compile error when I try building the contracts with the event double-declared.

If this can be deployed after being merged (and authorized by the trade delegate), that would be awesome!

Brechtpd:
Strange. Yes, best to add the definition to the interface as well in any case.

Seems like it could be client specific though we definitely had this problem when testing in protocol 3. 
Some more information can be found here if you're interested (under **Events and libraries**):
- https://blog.aragon.org/library-driven-development-in-solidity-2bebcaf88736/

Brechtpd:
Maybe this flattening changes the behaviour in some way. We double declare events all over the place in protocol 3, though in different files.

coreycaplan3:
"In the same way that libraries don't have storage, they don't have an event log. But they can dispatch events, let me explain that..."

Seems like it makes sense that it works then?

coreycaplan3:
I was wrong. It wasn't compiling because of a cache issue. When I deleted the `build` folder and recompiled, it worked like a charm!

coreycaplan3:
@kongliangzhong can you merge/deploy this when you get the chance? Thank you so much! 😄

PR #282, opened 2019-07-18
[Protocol2] Changes necessary for Dolomite's launch & re-deployment 

zrubenst:
Makes necessary changes to further fix order matching to conform to Dolomite's expectations

### DIFF
--- a/packages/loopring_v2/contracts/helper/ParticipationHelper.sol
+++ b/packages/loopring_v2/contracts/helper/ParticipationHelper.sol
@@ -83,9 +83,17 @@ library ParticipationHelper {
             p.feeAmountB = p.fillAmountB.mul(p.order.tokenBFeePercentage) / ctx.feePercentageBase;
         } else {
             // Calculate matching fees
-            p.feeAmount = p.order.feeAmount.mul(p.fillAmountS) / p.order.amountS;
             p.feeAmountS = 0;
             p.feeAmountB = 0;
+
+            // Use primary token fill ratio to calculate fee
+            // if it's a BUY order, use the amount B (tokenB is the primary)
+            // if it's a SELL order, use the amount S (tokenS is the primary)
+            if (p.order.isBuy()) {
+                p.feeAmount = p.order.feeAmount.mul(p.fillAmountB) / p.order.amountB;
+            } else {
+                p.feeAmount = p.order.feeAmount.mul(p.fillAmountS) / p.order.amountS;
+            }
 
             // If feeToken == tokenB AND owner == tokenRecipient, try to pay using fillAmountB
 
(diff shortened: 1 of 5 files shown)


packages/loopring_v2/contracts/helper/RingHelper.sol
```diff
@@ -171,11 +192,22 @@ library RingHelper {
...
+        // NOTICE: deprecated logic, rings must be of size 2 now
+        // ring.valid = ring.valid && (ring.size > 1 && ring.size <= 8); // invalid ring size
+        
+        ring.valid = ring.valid && ring.size == 2;
+
+        // Ring must consist of a buy and a sell
+        ring.valid = ring.valid && (
+            (ring.participations[0].order.isBuy() && !ring.participations[1].order.isBuy()) ||
+            (ring.participations[1].order.isBuy() && !ring.participations[0].order.isBuy())
+        );
```
Brechtpd: Is this needed? If the order is used as the maker it doesn't really matter if it's a buy or a sell order, so matching buy/buy and sell/sell should be fine. Unless you have other reasons to do this.

packages/loopring_v2/contracts/helper/ParticipationHelper.sol
```diff
@@ -109,13 +117,19 @@ library ParticipationHelper {
         }
 
         if ((p.fillAmountS - p.feeAmountS) >= prevP.fillAmountB) {
-            // The miner (or in a P2P case, the taker) gets the margin
-            p.splitS = (p.fillAmountS - p.feeAmountS) - prevP.fillAmountB;
+            // NOTICE: this line commented as order recipient should receive the margin
+            // p.splitS = (p.fillAmountS - p.feeAmountS) - prevP.fillAmountB;
+
+            // Order recipient receives margin
+            p.splitS = 0;
```
Brechtpd: No splitS trickery needed to store the correct trading history data for buy orders?

zrubenst: Pushing that soon.

zrubenst: Moved code from another codebase, forgot to pull most recent from that code base before moving it over

packages/loopring_v2/contracts/helper/RingHelper.sol
zrubenst: Its more so an easy way to check that the orders have the same primary token denoted in the `trancheB`. I'm not completely sure that orders need to have the same primary token set, however. @coreycaplan3 any thoughts?

zrubenst:
@Brechtpd Pushed a commit for handling splitS

Brechtpd:
> @Brechtpd Pushed a commit for handling splitS

Yes, that's what I expected. :)
It doesn't matter much, but you can remove `splitS` from the code below:
```
// Update spendables
uint totalAmountS = p.fillAmountS + p.splitS;
```

packages/loopring_v2/contracts/helper/RingHelper.sol
coreycaplan3: Yeah because we need to ensure that both orders have the same primary-ticker (base symbol). That would change how we match things on our end if the primary is not the same

coreycaplan3:
I adjusted the burn rate to `TIER_2` for everything except LRC. Ideally, Dolomite would like to allow users to get fee in `tokenB` in the future. This will make the UX more seamless, since fees can be taken out from the trade, instead of enforcing users having a certain amount of the token, in addition to the trade size.

This will make Loopring's Oedax more interesting, since there will be more tokens to auction off :)

I chose tier-2, since the 15% burn was acceptable for WETH and other popular tokens. I thought it was consistent with what we are trying to provide to the users, while not losing too much of our margin to the burn.

PR #290, opened 2019-07-22
[Protocol3] allow offchainData even when onchainDataAvailability is on

kongliangzhong:


dong77:
It makes sense to me to allow this parameter to have an arbitrary value regardless of the value of `onchainDataAvailability`.

dong77:
@kongliangzhong are you sure that all unit tests will pass?

kongliangzhong:
@dong77 , I am quite sure.

Issue #292, opened 2019-07-22
[lightcone_v2.js] Use eslint to auto lint/format js and ts files.


dong77:
Please refer to https://github.com/Loopring/protocols/pull/279 for how this can be done.

xiaowheat:
Completed.

PR #296, opened 2019-07-23
[lightcone_v2.js] enable eslint

dong77:
Addressing issue #292. Currently, this PR is not working as I encountered the following error. Note that even without my change, this error also occurs.

Screen Shot 2019-07-23 at 13 45 29


xiaowheat:
To avoid merge conflict, the PR will be merged after https://github.com/Loopring/protocols/pull/297.

PR #298, opened 2019-07-23
[Protocol3] pluggable blockType support

kongliangzhong:


kongliangzhong:
@letsgoustc , In order to update the State in Exchange contract, the only way is to use the library. Each
 blockType has its own processing library,  and these libraries are deployed separately. I think we can convert a library address into IBlockProcessor object and invoke the processBlock method. I am not sure this can work, still trying.

Brechtpd:
State can be updated in a different contract using `DELEGATECALL`.

If even deposits/on-chain withdrawals pass through this IBlockProcessor, and these update the state directly of the exchange, we're kind of creating a proxy contract for the block commitment, just split up in multiple contracts per block type. All off-chain request block types don't need to update any specific Exchange state so we only need to do some checks on the given data, so no Exchange state updates needed for this.

Another thing that is good to know is that sending all the data to the block processor will be pretty expensive (currently expensive memory-to-memory copies, may improve in the future). Either the amount of data can be limited/block type (e.g. call `processBlock` with only the starting 128 bytes of data for this block type, would have to be configurable) or we use something like #139 which splits the data up with what's needed on-chain and DA data (which is never needed onchain, but it is by far the most data when DA is used). The data needed on-chain is very limited and can be copied around cheaply.

kongliangzhong:
@Brechtpd ,  Thanks, Brecht, will try this way.

dong77:
@kongliangzhong please make sure you merge with the master first. Brecht has made quite some changes.

PR #308, opened 2019-07-26
[Protocol3] fix 'npm install' error

dong77:


packages/loopring_v3/package.json
```diff
@@ -98,9 +99,10 @@
     "sha3": "^2.0.0",
     "snarkjs": "0.1.11",
     "solidity-coverage": "^0.5.11",
-    "truffle": "^5.0.22",
+    "solium": "^1.2.4",
```
Brechtpd: `ethlint` (which is already under devDependencies) is the better package for solium which was renamed:
> If you're currently using the solium package for npm install, it is highly recommended that you move to ethlint. The solium package will not receive updates after December, 2019. There are no differences between the updates pushed to ethlint and solium packages.

https://github.com/duaraghav8/Ethlint#install

dong77: fixed

PR #310, opened 2019-07-26
[Protocol3] Verifier fixes + snark field overflow checks

### LINKED ISSUE #309: [Protocol3] Verifier v3 zkSNARK bug
Please ensure that all inputs you pass to the verifier are within the SNARK scalar field.

In the following locations:

 * https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/thirdparty/Verifier.sol#L121
 * https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/thirdparty/BatchVerifier.sol#L67
 * https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/thirdparty/Pairing.sol#L87
 * https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/thirdparty/BatchVerifier.sol#L86

For example:

```solidity
require( mul_input > 0 && mul_input < 21888242871839275222246405745257275088548364400416034343698204186575808495617 ); 
```

Where `q=21888242871839275222246405745257275088548364400416034343698204186575808495617` is the scalar field for zkSNARK inputs.

Not doing so could allow input aliasing, where the resulting point used as input is the same but the on-chain value is different.

For example, if you have a 'nullifier' which prevents double spend, the following piece of code can find alternative inputs which (when reduced modulo `q`) would result in the same value:

```python
p = 0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001
x = 0xb3e98c3f7346edc7a04b2abd18ca16c11c9419c3c50bd1a34cc485808e895d8e
y = x % p
import z3
s = z3.Solver()
k = z3.BitVec('k', 256)
s.add(k != x)
s.add(k % p == y)
while s.check():
    model = s.model()
    found = int(str(model[k]))
    print(found)
    s.add(k != found)
```

Brechtpd:
Should fix #309 

While we were currently not vulnerable to this exploit the extra checks were added for safety.


packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -122,8 +122,8 @@ library ExchangeBlocks
                 "PROOF_TOO_LATE"
             );
 
-            // Maybe we should strip the highest bits of the public input so we don't have any overflow (uint/prime field)
-            publicInputs[i] = uint(specifiedBlock.publicDataHash);
+            // Strip the highest bits of the public input so we don't have any overflow in the snark field
+            publicInputs[i] = uint(specifiedBlock.publicDataHash) >> 3;
```
dong77: The team "highest bits" is a bit confusing. If you meant the most significant bits, you should use:
`(uint(publicDatahash) << 3) >>3`; if you meant the least significant bits, then your implementation is correct.

packages/loopring_v3/contracts/impl/libexchange/ExchangeData.sol
```diff
@@ -164,6 +164,9 @@ library ExchangeData
     function GENESIS_MERKLE_ROOT() internal pure returns (bytes32) {
         return 0x2b4827daf74c0ab30deb68b1c337dec40579bb3ff45ce9478288e1a2b83a3a01;
     }
+    function SNARK_SCALAR_FIELD() internal pure returns (uint256) {
+        return 21888242871839275222246405745257275088548364400416034343698204186575808495617;
```
dong77: I wonder how this number is generated? its hex value is 0x30644e72e131a000000000000000000000000000000000000000000000000000

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -118,6 +118,10 @@ library ExchangeDeposits
             feeETH
         );
 
+        // Make sure the public key can be stored in the SNARK field
+        require(account.pubKeyX < ExchangeData.SNARK_SCALAR_FIELD(), "INVALID_PUBKEY");
```
dong77: since `ExchangeData.SNARK_SCALAR_FIELD()` is only used here, should we just declare a constant var locally?

packages/loopring_v3/test/testBlockVerifier.ts
```diff
@@ -560,11 +576,33 @@ contract("BlockVerifier", (accounts: string[]) => {
...
         commitBlocksSize1[0].blockVersion,
-        publicDataHashes,
+        publicInputs,
         exchangeTestUtil.flattenList(commitBlocksSize1.map(x => x.proof)),
         { from: anyone }
       );
       assert(!success, "verification should not be succesful");
     });
+
+    it("should not be able to verify multiple blocks of the same circuits with public input > scalar field", async () => {
```
dong77: In the current implementation, public input == scalar field will also fail.

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
dong77: Should we check pubKeyX/Y before account creation and password reset instead of at places where these values are used?

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
Brechtpd: Ah yes, the comment here is wrong, will fix.

packages/loopring_v3/contracts/impl/libexchange/ExchangeData.sol
Brechtpd: It's the prime number that is used for the field for the BN128 curve. See [EIP-196](http://eips.ethereum.org/EIPS/eip-196) for example.

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
Brechtpd: > since `ExchangeData.SNARK_SCALAR_FIELD()` is only used here, should we just declare a constant var locally?

I think it makes sense to do it this way because it may be used elsewhere in the future. Seems to be inline with how we use the other constant values there as well.

> Should we check pubKeyX/Y before account creation and password reset instead of at places where these values are used?

- Any account creation/account update will also do a deposit so the check will be done in these cases
- The deposit is the place where the data is sent to the circuit

So I think it makes sense to do it there to make sure the check happens in all cases (technically it's not even needed).

packages/loopring_v3/test/testBlockVerifier.ts
Brechtpd: Yes, it should say `>=`, will fix.

packages/loopring_v3/contracts/impl/libexchange/ExchangeData.sol
dong77: i see, can you add your response as a comment to the code.

dong77:
Please wait for @letsgoustc to take another look.

PR #313, opened 2019-07-29
[Protocol3] add support for future address KYC/whitelist

dong77:
The address white list will only be checked against when new accounts are created. We always allow existing accounts to update account settings. For accounts that have created, the operator can maintain an off-chain whitelist and use it to decide not to match some users' orders anymore.


dong77:
@Brechtpd believes putting all whitelisted addresses onchain will be overwhelmingly expensive for operators, which I totally missed and fully agree. I've created issue #317 to explore an off-chain-whitelist based solution which is more practical.

Issue #289, opened 2019-07-22
[Protocol3] Pluggable Block Types


dong77:
In our current design, there are some block-type specific logics to check the validity of blocks based on their types. I propose to migrate these logics to block's own smart contract and maintain a mapping between `blockTypeId` to the address of these block specific smart contracts. The mapping can be modified by the owner or later by the Loopring DAO. With this **pluggable block types**, we can enrich the protocol with new types of blocks such as privacy-preserved settlement and derivatives trading.

block type ids should replace the current block type enumeration and only be reused if the block's data scheme is fully compatible with the previous version.

Certain blocks such as on-chain deposit and withdrawal may be hardcoded into the main smart contract and become un-upgradable.

dong77:
@Brechtpd I talked about this with @kongliangzhong, he will follow up with an implementation.

dong77:
@kongliangzhong please also reference issue #294

dong77:
@Brechtpd After my attempt to implement this and thinking after your feedback, I feel like we should simply use OwnedUpgradeabilityProxy to make sure the entire `IExchange` is upgradable.

Brechtpd:
@dong77 I also think that's the best idea.

dong77:
ok. I’ll close this issue then. Going forward let’s use proxy to achieve upgradability.

PR #321, opened 2019-07-30
[Protocol3] (draft): implement pluggable block types

dong77:


dong77:
@Brechtpd @letsgoustc @kongliangzhong 
I've created this draft implementation for #https://github.com/Loopring/protocols/issues/289
Please review the **overall implementation** and let me know what you think.

packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -427,7 +427,41 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
...
+            let ptr := mload(0x40)
+            calldatacopy(ptr, 0, calldatasize)
+            let result := delegatecall(gas, _addr, ptr, calldatasize, 0, 0)
+            let size := returndatasize
+            returndatacopy(ptr, 0, size)
+
+            switch result
+            case 0 { revert(ptr, size) }
+            default { return(ptr, size) }
+        }
```
Brechtpd: Delegatecall can also be done in [solidity](https://solidity.readthedocs.io/en/v0.5.10/units-and-global-variables.html?highlight=delegatecall#members-of-address-types) like this:
`<address>.delegatecall(bytes memory) returns (bool, bytes memory)`

`encodeWithSelector` can be used to encode the parameters in bytes

Brechtpd:
I don't see any problems with the overall design.

But looking at the code I'm not sure what the pros/cons of only making the block commitment with proxys vs making the complete exchange with a proxy. A block type with an error can still risk all users funds. Hiding the comlete exchange after a proxy is more flexible, though you could say more risky (because more code can be changed) but can also be less risky because potential errors can be fixed.

packages/loopring_v3/contracts/impl/Exchange.sol
dong77: This is not important, you can optimize it later :)

dong77:
Putting  the DEX contract behind a proxy is an easier approach but it has a drawback: when we deploy a new DEX contract to replace the old one, we have to somehow replicate the entire history, eg the virtual blockchain. This can be very expensive. Otherwise users will have to withdraw and re-deposit. 

Putting BlockProcessors behind a proxy allows us to upgrade the DEX behaviors and in the meanwhile reuse the existing virtual blockchain data. 

DEX operators can always ignore new block types if they don’t want to opt-in — all onchain block types are now allowed to be disabled.

dong77:
A not for myself: disabling a blocktypr should still allow proof verification but not block commission.

Brechtpd:
> Putting the DEX contract behind a proxy is an easier approach but it has a drawback: when we deploy a new DEX contract to replace the old one, we have to somehow replicate the entire history, eg the virtual blockchain. This can be very expensive. Otherwise users will have to withdraw and re-deposit.

I don't think that's true. If you [change the implementation](https://github.com/Loopring/protocols/blob/f4b9da6963cbbd861a8ff3b57dcd105839efda57/packages/loopring_v3/contracts/thirdparty/Proxy.sol#L141) all state is immediately available in the new implementation (just like it's immediately available for the block processors in this design). Only the logic changes, the state remains unchanged (though additional state can be added if I remember correctly).

dong77:
BaseUpgradeabilityProxy is interesting. I'll try to figure out how to put it in front of a DEX instance.

Issue #317, opened 2019-07-29
[Protocol3] Allow DEX operators to reject account creation requests and send back tokens


dong77:
Our current design guarantees that all deposit requests will be handled and corresponding offchain accounts will be created. We should allow DEX operators to reject certain account creation requests, i.e, based on an off-chain address whitelist.

Balances associated with rejected deposits should be returned to users automatically in the same block, this requires some work. We should also consider how to prevent a Sybil account-creation attack as sending-back tokens will cost gas.

dong77:
@Brechtpd this may imply more work than expected.

Brechtpd:
Only ETH will need to be sent back immediately (which we already do when the user pays too much fee), tokens are transferred inside the contract in `deposit` so the function will not get there when doing the check earlier.

It's the depositor that pays all the gas for on-chain requests so I don't think any attack is possible.

dong77:
When a user is trying to create an account by depositing 10 LRC, those LRC will also be transferred to the DEX contract. Later when the operator decides to reject this request, those 10 LRC should be returned to the user in the corresponding deposit block.  Right?

Brechtpd:
Ah I think I understand the confusion. So my idea was that the signature check would still be checked on-chain, not off-chain at a later time. Something like this:
- A user wants to create an account and lets the DEX know
- The DEX signs something like (exchangeID, userAddress) and gives the user the signature
- The user can now create create an account with the given signature by passing it to e.g. `createOrUpdateAccount`.
- On-chain we check that the signature was created by the DEX, otherwise we reject the request immediately.

Now that I think about this, an even simpler way to do this is to only allow the DEX to create accounts. A user can then check if the owner of the account is his address (using `getAccount`) and can then start using it. We'd have to create a new function like `createAccountFor` or something like this

If we make the interface something like `isWhitelisted(address accountCreator, address accountOwner, bytes data)` the DEX can implement their own solution (only allow a certain accountCreator (== msg.sender) to create accounts, using whitelisted addresses stored onchain, by sending the signature in `data`, ...).

dong77:
I also removed this issue from beta3, FYI.

dong77:
If we can safely assume there are enough slots for accounts, maybe it's quite OK for anyone to be able to create any number of accounts with a fee -- when there are 50% slots used, the operator can increase the account creation fee by a factor of two or there.  The fee should be set so that no one will not likely to spend that much money to create too many accounts as a Sybil attack.

[...]

PR #329, opened 2019-08-01
[Protocol3] Allow permissioned account creation through an IAddressWhitelist contract

dong77:
This is to address #317.
 The signature verification part has not been implemented yet. But please review the overall solution.


dong77:
The first 64 bits can be a timestamp, the rest bits should be a multisig against `hash(user_address, timestamp)`

Brechtpd:
Looks good.

Any reason you went for the signature only approach (with some flexibility using multisig, but still a bit limited)? I still like the idea of doing stuff like this in a contract with an interface like `isWhitelisted(address accountCreator, address accountOwner, bytes permission)` instead of trying to come up with a data structure/implementation in the Exchange itself that is hopefully flexible enough.

The DEX would just set some contract, just like `_permissionedAccountCreation` is set now (if the contract is `0x0` all account creations are allowed). `bytes memory permission` would be some data completely opaque to the Exchange contract. It seems like this would move complexity outside of the Exchange contract and would be extremely flexible (the contract can be changed at any time, the DEX can call methods on this contract so that some signature can be enabled/disabled, ...).

dong77:
> Looks good.
> 
> Any reason you went for the signature only approach (with some flexibility using multisig, but still a bit limited)? I still like the idea of doing stuff like this in a contract with an interface like `isWhitelisted(address accountCreator, address accountOwner, bytes permission)` instead of trying to come up with a data structure/implementation in the Exchange itself that is hopefully flexible enough.
> 
> The DEX would just set some contract, just like `_permissionedAccountCreation` is set now (if the contract is `0x0` all account creations are allowed). `bytes memory permission` would be some data completely opaque to the Exchange contract. It seems like this would move complexity outside of the Exchange contract and would be extremely flexible (the contract can be changed at any time, the DEX can call methods on this contract so that some signature can be enabled/disabled, ...).

Will do that way.

dong77:
@Brechtpd please take another look. I suspect this will break some tests. Any suggestion for me where I should start fixing tests?

Brechtpd:
Looking good.

I would suggest fixing all updated functions in `testExchangeUtil.ts` first so pretty much all tests will work again.

dong77:
@Brechtpd  ready for review.

packages/loopring_v3/contracts/iface/IExchange.sol
```diff
@@ -200,12 +206,16 @@ contract IExchange
     /// @param  pubKeyX The first part of the account's trading EdDSA public key
     /// @param  pubKeyY The second part of the account's trading EdDSA public key.
     ///                 Note that pubkeyX and pubKeyY cannot be both `1`.
+    /// @param  permission The operator's signature to allow the creation of new accounts.
```
Brechtpd: 'The operator's signature' should be some data used for whitelisting.

packages/loopring_v3/contracts/iface/IExchange.sol
```diff
@@ -604,6 +614,9 @@ contract IExchange
     ///
     /// @param  pubKeyX The first part of the account's trading EdDSA public key
     /// @param  pubKeyY The second part of the account's trading EdDSA public key
+    /// @param  permission The operator's signature to allow the creation of new accounts,
```
Brechtpd: Same here

Issue #260, opened 2019-07-10
[Protocol3] Problem distributing to non-existing accounts in shutdown


Brechtpd:
Sometimes an account is used to send tokens to another account that does not need to sign something off-chain or isn't triggered by an on-chain request (the operator or the wallet). In cases like this it's possible `accountID >= accounts.length`. This is fine in almost all cases, the funds are lost or are locked up until the account is created.

In shutdown mode however, these funds need to withdraw to reset the Merkle tree. This causes 2 things:
- There is no account associated with the accountID so we would have to make sure we check that `accountID <  accounts.length` when distributing the tokens, and otherwise send the tokens to e.g. the protocol fee pool.
- We give extra time for the withdrawals the more accounts are created. We won't give extra time in this case because we don't know on-chain the account exists.

We can solve this by either
1. don't care about the slightly wrong estimate how much work is needed (which isn't a very good estimate anyway because the work required to reset an account can wildly differ) we just make sure that withdrawals don't fail when `accountID >= accounts.length`.
2. not allowing funds to be sent to accounts with `pubKey == 0` because that's the value for accounts that were not created yet. We simply disallow setting the pubKey to 0 when creating an account. We still allow setting an invalid pubKey to disallow signing, users just need to use a different value (e.g. `pubKey == 1`).
3. checking if `accountID < accounts.length` for all accounts that don't have to sign anything in the circuits.  We would have to pass  `accounts.length` to the circuits and do a slightly more expensive 'less than' check.

I think options 1 and 2 are the best because they are the easiest to implement. Option 1 is I think good enough to fix this issue, though it doesn't hurt to do option 2 because it's easy.

dong77:
I would vote for option#2, in the meanwhile, maybe we should make sure `accontID < accounts.length` always holds?

Brechtpd:
I implemented option 1 in #263 .

With the updated fee model only the operator will be able to receive fees on any account, and the operator will be controlled in some way by the DEX. So by using an invalid accountID in this case
- the operator does not receive the fees for the block
- the DEX makes the shutdown procedure more expensive for itself

In the current fee model the wallet account ID could be chosen freely by users which could be abused, but that will not be possible any more in the updated fee model.

Maybe still a good idea to do option 2, but we can do it when implementing the updated fee model.

dong77:
Ok. But let’s keep this issue open and revisit it after the fee model is upgraded

dong77:
Fixed with option 1&2 implemented.

Issue #197, opened 2019-05-17
[Protocol3] New fee model and examples


Brechtpd:
## New Fee Model Design
> @Brechtpd Staking has been moved to https://github.com/Loopring/protocols/issues/208.

### Protocol Fee

This fee is proportionally applied on every token transfer part of the trade. The protocol fee can be lowered by staking LRC. A different protocol fee can be used for maker orders and taker orders.
A protocol fee value is stored in a single byte in `bips/10`. This allows us to set the protocol fee up to `0.255%` in steps of `0.001%` (`0.1 bips`). 

#### Different treatment of maker and taker orders

The ring-matcher chooses which order is the maker and which order is the taker.
The order at position 0 in the ring is always the taker order, the order at position 1 is always the maker order. We always use the rate of the second order for the trade.

### Removal of the fee token

No more fee token/fee amount. The protocol fee is applied on the amount traded. A possible negative caused by this is that it's not possible to do fee payments in security tokens, NFTs, ... but we don't support these yet in protocol 3 anyway. Can be added back in a future version of the protocol when it makes sense.

### Fee payments

Paid in the tokenB of the order. These are not taxed or do not have any other disadvantage. The order owner pays the fee to the ring-matcher respecting `feeBips`. These values are calculated on the complete amount bought, the protocol fee does not change this in any way.

```
private calculateFees(
    fillB: BN,
    protocolFeeBips: number,
    feeBips: number,
    rebateBips: number
{
    const protocolFee = fillB.mul(new BN(protocolFeeBips)).div(new BN(100000));
    const fee = fillB.mul(new BN(feeBips)).div(new BN(10000));
    const rebate = fillB.mul(new BN(rebateBips)).div(new BN(10000));
    return [fee, protocolFee, rebate];
  }
```
feeBips's max value is 0.63% in steps of 0.01% (1 bips).
### Buy/Sell orders

Users can decide if they want to buy a fixed number of tokens (a buy order) of if they want to sell a fixed number of tokens (a sell order).  A buy order contains the maximum price the user wants to spend (by increasing amountS), a sell order contains the minimum price the users wants to get (by decreasing amountB). This allows us to do market orders. It only matters if an order is a buy or a sell order when the order is used as the taker order, the price for the maker order is always the price defined in the order.

So if there is a positive spread between the taker and the maker order we either use the spread to buy more tokens (sell order) for the taker, or let the taker keep the spread (buy order).

### Rebates

[...]

PR #331, opened 2019-08-01
[Protocol3] Updated fee model (removed ring-matcher, added labels)

Brechtpd:
See #197 and #260

- For ring settlement blocks the ring operator now signs (publicInput, nonce). The nonce of the operator is increased afterwards. This ensures the operator knows exactly what the block does. The nonce also prevents any replay attacks.
- Nothing in the DA is XORed anymore (the ring-matcher data was previously XORed). We will have to see what makes sense with realistic data.
- Accounts can now only be created with `pubKey.X > 0`. This allows the circuits to detect accounts that were not yet created by a simple `pubKey.X != 0` check.
- Labels are added to all off-chain requests. These labels are currently 32bit. It's actually quite cheap to make them bigger (making them 64bit instead of 32bit would only add 64 constraints/trade).


dong77:
what's the number of constraints for settlement circuit after this optimization?

Brechtpd:
> what's the number of constraints for settlement circuit after this optimization?

Updated the numbers [in the design doc here](https://github.com/Loopring/lightcone2/blob/edf1d39c818e6d80ad7c5371dfed5714e6547d85/PROTOCOL3_DESIGN.md#throughput-ring-settlements) with some Istanbul estimates.

Issue #213, opened 2019-06-10
[Protocol3] Fine-tune downtime price


Brechtpd:
Previous discussion: #212

Currently every exchange pays a fixed price for downtime:

```
cost = durationMinutes.mul(S.loopring.downtimePriceLRCPerMinute());
```

Maintenance mode locks out the ability of users to withdraw, so it' extremely important that the time in maintenance is as short as possible.

So in any case the price for downtime needs to be high in all cases. I think it should make sense to make it more cost effective for exchanges to have backup servers to handle a limited number of on-chain requests than to buy maintenance mode. Maintenance mode should mainly be used for unforeseen problems because it greatly impacts user experience.

We can change the downtime price by
- checking totalDowntime / totalExchangeLifetime and making it more expensive the more the exchange has been in maintenance
- checking how much funds are locked into the Exchange
- ...

If we implement this, it probably makes sense to implement any additional logic for this in the Loopring contract instead of the Exchange contract.

dong77:
@kongliangzhong, I'm assigning this issue to you so you can work on this one and https://github.com/Loopring/protocols/pull/349 and #350 all together **before** the release of beta3.

dong77:
Completed in #343

PR #343, opened 2019-08-03
[Protocol3] add IDowntimeCostCalculator contract

dong77:
This is to address #213.


dong77:
@Brechtpd  please take a look. I'll fix tests before committing.

packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
```diff
@@ -205,7 +205,22 @@ library ExchangeAdmins
         returns (uint)
     {
         require(!S.isInWithdrawalMode(), "INVALID_MODE");
-        return durationMinutes.mul(S.loopring.downtimePriceLRCPerMinute());
+
+        // Initially, the penalty is the percentage of DEX's downtime.
+        uint penalty = S.totalTimeInMaintenanceSeconds.mul(100) /
```
Brechtpd: I think this should be e.g. `.mul(10000)` because a minimum downtime of 1% is already very bad downtime.
I also think we should calculate with the extra downtime minutes already included (otherwise it's better to buy a lot of downtime when the penalty is still low), so something like this:
```
uint penalty = S.totalTimeInMaintenanceSeconds.add(durationMinutes.mul(60)).mul(10000) / 
     (now - S.exchangeCreationTimestamp);
```

Brechtpd:
> @Brechtpd please take a look. I'll fix tests before committing. BTW, in the test, where should I change to provide an initial value for the new parameter `downtimePriceMaxPenalty`.

Changes will have to be done in at least `4_deploy_protocol.js` (some dummy value), `testLoopring.ts` and `testExchangeUtil.ts` (the actual value that will be used in most tests). Maybe some other places as well.

packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
dong77: I knew this, and I want to give DEX owner the chance to buy more downtime than initially needed with a discount.  This will end up locking up extra LRC.

dong77: For business people, they just need to remember under 1%, the downtime price is the same. We don’t need to fine tune the price based in smaller granularity, it will probably confuse people. 1% means in one year, the DEX can be taken off chain for 3 days and a half.

Brechtpd: I'm okay with this is if the price for downtime is very high to begin with. Though I still think it's a good idea to increase the price the more downtime is purchased even in a single purchase, security of user funds should come first here.

This is a risky feature. User funds are safe, but the DEX owner can lock them up at any time for an unspecified time. If the downtime cost starts at $1000/hour, and the DEX buys the downtime in a single purchase when the price is low, than it costs _only_ $8.5M to lock up everybody's money for 1 year. Binance can lose $40M in a hack without any real problem... I can't really think of a good reason why a DEX would do this, but it can be done so it is a risk for users (and the exchange loses some of the non-custodialness because of this).

dong77: If the downtime calculation is more complicated than this one, I prefer moving it to another contract (`DowntimePriceCalculator` or a better name?) for upgradability.
It will have a method:

```

getDowntimePrice(
    uint totalTimeInMaintenanceSeconds,
    uint totalDEXLifetimeSeconds,
    uint availableDowntimeMinutes,
    uint amountOfLRCStakedbyOwner,
    uint durationToPurchaseMinutes
    )
    external
    view
    returns (uint priceAsLRCPerMinute);
```

packages/loopring_v3/contracts/impl/auxiliaries/DynamicDowntimePriceProvider.sol
```diff
@@ -0,0 +1,126 @@
...
+        uint  totalDEXLifeTimeSeconds,
+        uint  numDowntimeMinutes,
+        uint  /* exchangeStakedLRC */,
+        uint  durationToPurchaseMinutes
+        )
+        external
+        view
+        returns (uint)
+    {
+        uint total = numDowntimeMinutes.add(durationToPurchaseMinutes);
```
Brechtpd: Not sure if I understand this. `numDowntimeMinutes` is the number of downtime minutes the DEX has purchased that are still left (though this value is only updated after calling `stopMaintenanceMode`). If the DEX purchases 120 minutes and calls `stopMaintenanceMode` after 90 minutes then `numDowntimeMinutes == 30`. The DEX only needs to call `startOrContinueMaintenanceMode` immediately after stopping to get the `gracePeriodPrice` cost again which seems like it will be cheaper than the normal price.

dong77: good point, it is a bug.

dong77: Fixed, please take another look.

packages/loopring_v3/contracts/impl/auxiliaries/DowntimeCostCalculator.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint totalTimeInMaintanance,
+        uint totalLifetime,
+        uint downtime
+        )
+        private
+        view
+        returns (uint)
+    {
+        require(downtime <= maxAwailableDowntime, "PURCHASE_PROHIBITED");
+        uint total = totalTimeInMaintanance.add(downtime);
```
Brechtpd: `totalTimeInMaintanance` is in seconds, but `downtime` is in minutes here.

dong77: Good catch.

dong77: fixed.

dong77:
Guys this PR is ready for a final review

Issue #201, opened 2019-05-30
[Protocol3] Allow withdrawing exchange balance not owned by users


Brechtpd:
The exchange contract can be sent tokens not deposited by users e.g. air drops, token swaps, user error, ... These funds should be withdrawable, most likely by the exchange owner.

We can do this by keeping track of what balance is actually owned by the users. The part that can we withdrawn is `token.balanceOf(exchange) - (totalDeposited[token] - totalWithdrawn[token])`.

dong77:
We need to disable the default function.

PR #354, opened 2019-08-05
[Protocol3] allow withdrawal of balances not owned by users

dong77:
This is to address issue #201


packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
```diff
@@ -256,4 +256,36 @@ library ExchangeAdmins
...
+            ERC20(token).balanceOf(address(this));
+
+        uint userBalance = S.tokenBalances[token];
+        assert(totalBalance >= userBalance);
+
+        amount = totalBalance - userBalance;
+
+        if (amount > 0) {
+            bool success;
+            if (token == address(0)) {
```
Brechtpd: ETH case can be removed

dong77: Why?

Brechtpd: Oh sorry, I was confused, you're right.

dong77:
@Brechtpd since you are most familiar with withdrawal/deposit logics, please make sure I updated `tokenBalances` correctly in the right places.

Brechtpd:
> @Brechtpd since you are most familiar with withdrawal/deposit logics, please make sure I updated `tokenBalances` correctly in the right places.

I think deposits/withdrawals are correct, but there's a problem with ETH because it's used as the fee for on-chain requests (and can be withdrawn as the block fee for the operator), so ETH needs some extra logic.

Brechtpd:
> I think deposits/withdrawals are correct, but there's a problem with ETH because it's used as the fee for on-chain requests (and can be withdrawn as the block fee for the operator), so ETH needs some extra logic.

Now that ETH cannot be deposited to the contract any more using the default function, and all other functions  which use ETH can never pay too much ETH (the surplus is immediately sent back) there is no way ETH stored in the contract is not owned by anyone (either by a user or for the block fee). The ETH case in `withdrawBalanceNotOwnedByUsers` should be removed and it should not be possible to withdraw ETH this way.

For tokens everything should work correctly as it is now. But let me think about this some more. :)

dong77:
> > I think deposits/withdrawals are correct, but there's a problem with ETH because it's used as the fee for on-chain requests (and can be withdrawn as the block fee for the operator), so ETH needs some extra logic.
> 
> Now that ETH cannot be deposited to the contract any more using the default function, and all other functions which use ETH can never pay too much ETH (the surplus is immediately sent back) there is no way ETH stored in the contract is not owned by anyone (either by a user or for the block fee). The ETH case in `withdrawBalanceNotOwnedByUsers` should be removed and it should not be possible to withdraw ETH this way.
> 
> For tokens everything should work correctly as it is now. But let me think about this some more. :)

Done

Brechtpd:
It seems like we don't need `NoDefaultFunc` any more:
> Contracts that receive Ether directly (without a function call, i.e. using send or transfer) but do not define a fallback function throw an exception, sending back the Ether (this was different before Solidity v0.4.0). So if you want your contract to receive Ether, you have to implement a payable fallback function.

- [Solidity docs](https://solidity.readthedocs.io/en/v0.5.3/contracts.html#fallback-function)

dong77:
> It seems like we don't need `NoDefaultFunc` any more:
> 
> > Contracts that receive Ether directly (without a function call, i.e. using send or transfer) but do not define a fallback function throw an exception, sending back the Ether (this was different before Solidity v0.4.0). So if you want your contract to receive Ether, you have to implement a payable fallback function.
> 
> * [Solidity docs](https://solidity.readthedocs.io/en/v0.5.3/contracts.html#fallback-function)

@Brechtpd can you create a separate PR for this?

Brechtpd:
> @Brechtpd can you create a separate PR for this?

Sure. But it's actually not used anywhere else (except for some old imports that should have been removed) so if it's removed here there's not much to do any more. But if you merge this one as is I'll make a PR that removes it completely.

dong77:
> > @Brechtpd can you create a separate PR for this?
> 
> Sure. But it's actually not used anywhere else (except for some old imports that should have been removed) so if it's removed here there's not much to do any more. But if you merge this one as is I'll make a PR that removes it completely.

I'll remove it.

Issue #353, opened 2019-08-05
[Protocol3] Support DEX upgradability be default


dong77:
[OwnedUpgradeabilityProxy](https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/thirdparty/OwnedUpgradeabilityProxy.sol) is very convenient to make sure a DEX is upgradable. The issue is that user assets are at risk -- the owner of the proxy can always upgrade the proxy to point to another contract that may move user fund.

My idea is to disable instant proxy upgrade to arbitrary contract and only allow the proxy owner to schedule an upgrade to a declared address in advance, for example, 30 days. When a scheduled upgrade is still pending, the owner can cancel it or schedule a new one to replace it, and the waiting period shall restart.

@letsgoustc @Brechtpd @hosschao please let me know what you think.

dong77:
BTW, we are of course assuming DAOs are not ready. The objective is not to prevent DEX owners from upgrading yo a malicious contract, but allows users of a DEX to have enough time to respond during the waiting period.

dong77:
After talking to @Brechtpd and @letsgoustc, a better way to offer transparent upgradability is to create a VersionManager that always maintain the latest version and force each DEX's proxy to point to the lastest version.

dong77:
Here is the basic idea.

Right now we have ILoopring3, but in the future, we may have ILoopring4, ILoopring5. And we have to assume they are not compatible. For Loopring3, the current design allows the creation of multiple exchange instances (each with its own id), but if we use a proxy instead for each DEX, we only need to have one single exchange instance behind all the proxies.

First, we can ask `ILoopring3/4/5` to implement a new interface called `ILoopring`, which has the following methods:

```javascript
   function getProtocolInfo() returns (uint16 version, address dexInstanceAddress);
   function registerProxy(address proxy) returns (uint dexId); // this will also initialize the proxy's exchange ID
```
When ILoopring3 is deployed, its constructor will automatically calls the `createInstance` function -- currently public but should be private -- to initialize the `address _dexInstanceAddress` field that will be retuned by the `getProtocolInfo` function. But the dexInstance's id will be 0.

We then create an IVersionManager instance which implements the following:

```javascript
  struct Protocol {
     address loopring;
     address instance;
  }
  mapping (uint16 => Protocol) protocols;
  function registerProtocol(address loopring) {
    ILoopring loopring = ILoopring(loopring);
    (ver, instance) = loopring.getProtocolInfo();
[...]

PR #355, opened 2019-08-05
[Loopring3](WIP) refactor deployment process to support upgradability

dong77:
This PR is to address #353.


dong77:
@Brechtpd  This is still a work-in-progress, but please take a look at the overall design.

Brechtpd:
Looks good to me.

packages/loopring_v3/contracts/impl/ExchangeDeployer.sol
```diff
@@ -24,26 +24,17 @@ import "./Exchange.sol";
...
     {
-        Exchange exchange = new Exchange(
-            exchangeId,
-            loopringAddress,
-            owner,
-            operator,
-            onchainDataAvailability
-        );
-        return address(exchange);
+        return address(new Exchange());
```
dong77: I’m thinking to initialize this instance in a way that nobody can use it as a standalone instance.

packages/loopring_v3/contracts/impl/LoopringV3.sol
```diff
@@ -139,16 +139,17 @@ contract LoopringV3 is ILoopringV3, Claimable
         emit ProtocolFeeVaultUpdated(protocolFeeVault);
     }
 
-    function createExchange(
+    function registerExchange(
+        address exchangeAddress,
+        address _owner,
         address payable _operator,
-        bool onchainDataAvailability
+        bool    onchainDataAvailability
```
dong77: I should probably add noReentrence to this method.

packages/loopring_v3/contracts/impl/ExchangeDeployer.sol
Brechtpd: Do we even need the ExchangeDeployer this any more? Because this deployment is only needed just once for every Exchange version it doesn't seem to make any sense to do it via ExchangeDeployer which also needs to be updated and redeployed for each Exchange version it deploys.

But it's possible someone could use the Exchange directly, yes. But if we deploy it we can disable it immediately by e.g. having an extra flag in the constructor of the Exchange which disables everything. There's probably some mechanism already out there that can be used for this.

packages/loopring_v3/contracts/impl/LoopringV3.sol
Brechtpd: I don't think it's needed now, but because none of the functions are called a lot, so the gas cost doesn't matter, it's probably a good idea to just mark all external functions as `nonReentrant` for safety.

dong77: Ok, will do

dong77: I will do it in a separate PR.

PR #365, opened 2019-08-07
[Protocol3] Fix a bug?

dong77:
@Brechtpd was it a bug?


Brechtpd:
@dong77 Yes, there's only a test that withdraws the complete amount. Seems like this needs an extra test.

PR #367, opened 2019-08-08
[Protocol3] refactor deployment to support upgradability (tests fail)

dong77:
1. ABI files will be generated automatically before tests run, so they are deleted from git (but we we add new solidity files, we need to update genAbi.sh)
2. We need further ensure ReentrancyGuard is used correctly and default function are used correctly in `Exchange.sol` and `LoopringRegistry` @Brechtpd please help me review this.
3. LoopringRegistry now maintains the ids of all exchanges (cross-version), LoopringV3.sol only maintains exchanges based on LoopringV3.
4. Many test files have been modified, but they should work just fine without (much) modification. Please only review **testExchangeUtil.ts**.


packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -230,8 +272,8 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
         address tokenAddress
         )
         external
-        nonReentrant
```
Brechtpd: I'd like to keep the order like this everywhere:
```
nonReentrant
onlyOwner
```
This makes it much easier to test non-reentrancy. The tests will use the ABI file to find all `external` functions and will test them automatically for reentrancy. If `onlyOwner` (or some other modifier) is the first modifier the test will fail because `msg.sender != owner` (the reentrancy is tested from a test contract).

packages/loopring_v3/contracts/impl/ProtocolRegistry.sol
```diff
@@ -0,0 +1,176 @@
...
+        (instance, version) = getProtocol(protocol);
+    }
+
+    function forgeExchange(
+        address protocol,
+        bool    supportUpgradability,
+        bool    onchainDataAvailability
+        )
+        public
+        nonReentrant
```
Brechtpd: It's best to only use `nonReentrant` on external functions. `public` functions can also be called in the contract itself, so if a `public` function that is `nonReentrant` is called from an `external` functions that is also `nonReentrant` it will fail, which is not something you'd expect.

packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -49,18 +50,43 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
...
+
+    /// @dev The constructor must do NOTHING to support proxy.
+    constructor() public {}
+
+    /// @dev The default function will do Ether depoeits.
+    function() external payable
+    {
+        uint96 amount = uint96(msg.value);
+        require(uint(amount) == msg.value, "TOO_LARGE");
+        state.depositTo(msg.sender, address(0), amount, 0);
```
Brechtpd: Currently the `amount` in the deposit is the actual amount the user wants to deposit, without the fee (to make the behavior the same with ERC20 tokens + to guard users against sudden increased deposit fees). So this will only work if the deposit fee of the exchange is 0. Subtracting the deposit fee here is possible (so that `amount - depositFee` is deposited) but is a bit dangerous for users for the reason given above.

packages/loopring_v3/contracts/impl/Exchange.sol
dong77: great, I'll change it.

packages/loopring_v3/contracts/impl/ProtocolRegistry.sol
dong77: ok.

packages/loopring_v3/contracts/impl/Exchange.sol
dong77: do you suggest we should disable this default function?

dong77: I should have fixed the issue, but maybe we should simply disable the default function.

Brechtpd: I also think it's best to disable the default function here (by not defining any).

packages/loopring_v3/contracts/impl/Exchange.sol
```diff
@@ -706,9 +737,9 @@ contract Exchange is IExchange, Claimable, ReentrancyGuard
         external
         nonReentrant
         onlyOwner
-        returns (address payable)
+        returns (address payable oldOperator)
     {
-        return state.setOperator(_operator);
+        // oldOperator = state.setOperator(_operator);
```
Brechtpd: Still needs to be commented out?

packages/loopring_v3/contracts/impl/Exchange.sol
dong77: OK, will do.

PR #371, opened 2019-08-08
[Protocol3] Update README.md and package.json

dong77:


Brechtpd:
It's a bit strange that the "run tests" steps contains the same instructions as the "build" steps directly above, though I guess it can't hurt to repeat it there.

dong77:
> It's a bit strange that the "run tests" steps contains the same instructions as the "build" steps directly above, though I guess it can't hurt to repeat it there.

Some people will just look at the test section.

PR #373, opened 2019-08-08
[Protocol3] fixes test failures

dong77:
After this PR only `testExchangeNonReentrant` has one test failure.

### DIFF
--- a/packages/loopring_v3/test/testExchangeMaintenanceMode.ts
+++ b/packages/loopring_v3/test/testExchangeMaintenanceMode.ts
@@ -31,7 +31,7 @@ contract("Exchange", (accounts: string[]) => {
     const numMinutesLeft = (await exchange.getRemainingDowntime()).toNumber();
     const numMinutesToBuy =
       duration >= numMinutesLeft ? duration - numMinutesLeft : 0;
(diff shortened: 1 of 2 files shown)


packages/loopring_v3/test/testExchangeMaintenanceMode.ts
```diff
@@ -446,12 +442,12 @@ contract("Exchange", (accounts: string[]) => {
 
       let duration = 123;
       let expectedCost = await getDowntimeCost(duration);
-      let cost = await exchange.getDowntimeCostLRC(duration);
+      let cost = await getDowntimeCost(duration);
```
Brechtpd: The idea here was to compare the solidity implementation of the downtime cost calculation with a JS implementation. Now both just use the solidity code. But okay, we can fine-tune the tests later.

Brechtpd:
I had to update the ABI files, then `testExchangeNonReentrant` worked okay. So I think all tests pass again now.

packages/loopring_v3/test/testExchangeMaintenanceMode.ts
dong77: will fix it later :)

PR #384, opened 2019-08-08
[Loopring3] refactor deployment process to support upgradability

dong77:
1. ABI files will be generated automatically before tests run, so they are deleted from git.
3. LoopringRegistry now maintains the ids of all exchanges (cross-version) and will burn LRC when an exchange is created, LoopringV3.sol only maintains exchanges based on LoopringV3.


dong77:
@Brechtpd  All tests pass.

packages/loopring_v3/contracts/impl/LoopringV3.sol
```diff
@@ -16,68 +16,103 @@
...
+        returns (address exchange)
+    {
+        exchange = ExchangeV3Deployer.deploy();
+        emit ExchangeCreated(exchange);
+    }
+
+
+    function initializeExchange(
+        address exchangeAddress,
+        uint    exchangeId,
```
letsgoustc: Instead of asking caller to pass into an exchangeId, why not we maintain it in LoopringV3 contract?

packages/loopring_v3/contracts/impl/LoopringV3.sol
```diff
@@ -251,20 +227,22 @@ contract LoopringV3 is ILoopringV3, Claimable
...
-                amountLRC
-            ),
+            lrcAddress.safeTransferFrom(msg.sender, address(this), amountLRC),
             "TRANSFER_FAILURE"
         );
-        stakedLRC = exchanges[exchangeId - 1].exchangeStake.add(amountLRC);
-        exchanges[exchangeId - 1].exchangeStake = stakedLRC;
+
+        Exchange storage exchange = exchanges[exchangeId];
+        require(exchange.exchangeAddress != address(0), "INVALID_EXCHANGE_ID");
```
letsgoustc: It's better to add exchangeID check before transferring real token.

packages/loopring_v3/genAbi.sh
```diff
@@ -0,0 +1,46 @@
...
+mv ABI/version30/contracts_lib_Killable_sol_Killable.abi ABI/version30/Killable.abi
+mv ABI/version30/contracts_lib_MathUint_sol_MathUint.abi ABI/version30/MathUint.abi
+mv ABI/version30/contracts_lib_Ownable_sol_Ownable.abi ABI/version30/Ownable.abi
+mv ABI/version30/contracts_lib_Poseidon_sol_Poseidon.abi ABI/version30/Poseidon.abi
+mv ABI/version30/contracts_lib_ReentrancyGuard_sol_ReentrancyGuard.abi ABI/version30/ReentrancyGuard.abi
+mv ABI/version30/contracts_thirdparty_BatchVerifier_sol_BatchVerifier.abi ABI/version30/BatchVerifier.abi
+mv ABI/version30/contracts_thirdparty_OwnedUpgradabilityProxy_sol_OwnedUpgradabilityProxy.abi ABI/version30/OwnedUpgradabilityProxy.abi
+mv ABI/version30/contracts_thirdparty_Proxy_sol_Proxy.abi ABI/version30/Proxy.abi
+mv ABI/version30/contracts_thirdparty_UpgradeabilityProxy_sol_UpgradeabilityProxy.abi ABI/version30/UpgradeabilityProxy.abi
+mv ABI/version30/contracts_thirdparty_Verifier_sol_Verifier.abi ABI/version30/Verifier.abi
```
letsgoustc: Instead of using hard coded file names. It's better to use code like this:

#!/bin/sh
  
ABI_PATH="ABI/version30"
for file in $ABI_PATH/*
do
    rename_file=`echo "$file" | awk -F'[_]' '{print $5}'`
    mv $file $ABI_PATH/$rename_file
done

packages/loopring_v3/contracts/impl/libexchange/ExchangeData.sol
```diff
@@ -188,6 +189,38 @@ library ExchangeData
...
 
+    function getConstants()
+        public
+        pure
+        returns(
+            bytes32 merkleRoot,
+            uint[19] memory constants
+        )
+    {
+        merkleRoot = GENESIS_MERKLE_ROOT();
```
Brechtpd: I don't think there's any need to special case `merkleRoot` here, putting it in the constants array should be fine.

packages/loopring_v3/migrations/2_deploy_libraries.js
```diff
@@ -0,0 +1,29 @@
+// Deploy all libraries that are shared by/linked from other contracts.
+// The purpose of deploying these libraries stand-alone is to reduce gas
+// usage in deploying other contracts.
+
+var AddressUtil = artifacts.require("./lib/LRC.sol");
```
Brechtpd: Seems like this has the wrong path

packages/loopring_v3/genAbi.sh
dong77: done

packages/loopring_v3/migrations/2_deploy_libraries.js
```diff
@@ -0,0 +1,29 @@
+// Deploy all libraries that are shared by/linked from other contracts.
+// The purpose of deploying these libraries stand-alone is to reduce gas
+// usage in deploying other contracts.
```
Brechtpd: Because these libraries only contain `internal` functions, I would think the contracts using them will still include the code inside the contract itself instead of the code in library so I'm not sure this works, but I never tried it.

packages/loopring_v3/migrations/2_deploy_libraries.js
dong77: Thank you for detecting this bug.

packages/loopring_v3/contracts/impl/ExchangeProxy.sol
```diff
@@ -0,0 +1,76 @@
...
+
+
+/// @title ExchangeProxy
+/// @dev This proxy is designed to support transparent upgradeability offered by a
+///      IProtocolRegistry contract.
+/// @author Daniel Wang  - <daniel@loopring.org>
+contract ExchangeProxy is Proxy
+{
+    bytes32 private constant registryPosition = keccak256("org.loopring.protocol.exchange.proxy.registry");
+    bytes32 private constant loopringPosition = keccak256("org.loopring.protocol.exchange.proxy.loopring");
```
Brechtpd: I think this variable can be removed

packages/loopring_v3/contracts/impl/LoopringV3.sol
dong77: Currently, we have only one major version (LoopringV3), but in the future, we may have multiple ones. All of them should be registered in the same ProtocolRegistry. Therefore the ID is global to ProtocolRegistry, not LoopringV3.

packages/loopring_v3/contracts/impl/LoopringV3.sol
dong77: this contract allows `exchangeId` to be 0, `ProrotocolRegister` may have stricter rules. Therefore we only care if the id's corresponding address is address(0) or not.

packages/loopring_v3/contracts/impl/LoopringV3.sol
```diff
@@ -273,88 +251,75 @@ contract LoopringV3 is ILoopringV3, Claimable
...
-                amountLRC
-            ),
+            lrcAddress.safeTransferFrom(msg.sender, address(this), amountLRC),
             "TRANSFER_FAILURE"
         );
-        stakedLRC = exchanges[exchangeId - 1].protocolFeeStake.add(amountLRC);
-        exchanges[exchangeId - 1].protocolFeeStake = stakedLRC;
+
+        Exchange storage exchange = exchanges[exchangeId];
+        require(exchange.exchangeAddress != address(0), "INVALID_EXCHANGE_ID");
```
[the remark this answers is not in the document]
dong77: same reason

packages/loopring_v3/contracts/impl/LoopringV3.sol
```diff
@@ -390,13 +356,21 @@ contract LoopringV3 is ILoopringV3, Claimable
...
     }
 
-    function()
-        external
-        payable
-    {}
+    function getProtocolFeeStake(
+        uint exchangeId
+        )
+        public
```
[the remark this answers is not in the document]
dong77: done

packages/loopring_v3/contracts/impl/ProtocolRegistry.sol
```diff
@@ -0,0 +1,222 @@
...
+        require(_lrcAddress != address(0), "ZERO_ADDRESS");
+        lrcAddress = _lrcAddress;
+    }
+
+    function registerProtocol(
+        address protocol,
+        string  calldata version
+        )
+        external
+        nonReentrant
```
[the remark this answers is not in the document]
dong77: you are right, fixed.

packages/loopring_v3/contracts/impl/libexchange/ExchangeData.sol
dong77: done

packages/loopring_v3/contracts/impl/ExchangeProxy.sol
dong77: done

packages/loopring_v3/contracts/impl/ProtocolFeeVault.sol
```diff
@@ -61,15 +62,16 @@ contract IOedax {
...
 /// @dev See https://github.com/Loopring/protocols/blob/master/packages/oedax_v1/contracts/iface/IAuction.so
-contract IAuction {
+contract IAuction
+{
     function settle() public;
     function ask(uint amount) external returns (uint accepted);
 }
 
 
 /// @title An Implementation of IUserStakingPool.
```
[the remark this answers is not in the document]
dong77: done

packages/loopring_v3/migrations/2_deploy_libraries.js
dong77: I think you are right. I found this:

> Embedded Library: If a smart contract is consuming a library which have only internal functions than EVM simply embeds library into the contract. Instead of using delegate call to call a function, it simply uses JUMP statement(normal method call). There is no need to separately deploy library in this scenario.

dong77:
@letsgoustc in the following code, we don't have to check `exchangeId !=0` because `exchanges[0]` is always `address(0)`.

```solidity
 function burnExchangeStake(
        uint exchangeId,
        uint amount
        )
        external
        nonReentrant
        returns (uint burnedLRC)
    {
        Exchange storage exchange = exchanges[exchangeId];
        address exchangeAddress = exchange.exchangeAddress;

        require(exchangeAddress != address(0), "INVALID_EXCHANGE_ID");
        require(exchangeAddress == msg.sender, "UNAUTHORIZED");

        burnedLRC = exchange.exchangeStake;

        if (amount < burnedLRC) {
            burnedLRC = amount;
        }
        if (burnedLRC > 0) {
            require(
                BurnableERC20(lrcAddress).burn(burnedLRC),
                "BURN_FAILURE"
            );

            exchange.exchangeStake = exchange.exchangeStake.sub(burnedLRC);
            totalStake = totalStake.sub(burnedLRC);
        }
        emit ExchangeStakeBurned(exchangeId, burnedLRC);
    }
```

dong77:
It took me 7 hours to run all the tests on my latest MacBook Pro. Is this normal, @Brechtpd ?


```shell
Using network 'development'.
Compiling your contracts...
===========================
> Compiling ./contracts/iface/IDowntimeCostCalculator.sol
> Compiling ./contracts/iface/IExchangeV3.sol
> Compiling ./contracts/impl/BlockVerifier.sol
> Compiling ./contracts/impl/DowntimeCostCalculator.sol
> Compiling ./contracts/impl/ExchangeProxy.sol
> Compiling ./contracts/impl/ExchangeV3.sol
> Compiling ./contracts/impl/ExchangeV3Deployer.sol
> Compiling ./contracts/impl/LoopringV3.sol
> Compiling ./contracts/impl/ProtocolFeeVault.sol
> Compiling ./contracts/impl/ProtocolRegistry.sol
> Compiling ./contracts/impl/libexchange/ExchangeAccounts.sol
> Compiling ./contracts/impl/libexchange/ExchangeAdmins.sol
> Compiling ./contracts/impl/libexchange/ExchangeBlocks.sol
> Compiling ./contracts/impl/libexchange/ExchangeData.sol
> Compiling ./contracts/impl/libexchange/ExchangeDeposits.sol
> Compiling ./contracts/impl/libexchange/ExchangeGenesis.sol
> Compiling ./contracts/impl/libexchange/ExchangeMode.sol
> Compiling ./contracts/impl/libexchange/ExchangeTokens.sol
> Compiling ./contracts/impl/libexchange/ExchangeWithdrawals.sol
> Compiling ./contracts/test/AccountContract.sol
> Compiling ./contracts/test/FixPriceDowntimeCostCalculator.sol
> Compiling ./contracts/test/Operator.sol
> Compiling ./contracts/test/tokens/TEST.sol
> Artifacts written to /var/folders/jr/yf45x41525qddnbr0nfyy1400000gn/T/test-11979-34607-vqdk42.rwrf
> Compiled successfully using:
   - solc: 0.5.10+commit.5a6ea5b1.Emscripten.clang

>>>>>>>> Deployed contracts addresses (deploy_registry):
ProtocolRegistry: 0xBA31F57570F3a3106EcA46dCBc379D5dcc7AA496

>>>>>>>> Deployed contracts addresses (deploy_aux):
BlockVerifier: 0x3c6d8c44D5B1C71e82Ed229c37F98C424E4eC83D
FixPriceDowntimeCostCalculator: 0xb458Dc448DdD95F373Fb5f3b38119A91f7a38e55
UserStakingPool: 0x22DdC9c131E714eB84becDE7F6bDb1b8Daf47F94
ProtocolFeeVault: 0x9fbfc833e510C32d5C046375e2b5d682d889487c

>>>>>>>> Deployed contracts addresses (deploy_exchange):
ExchangeData:  0x430aE9c3d093d03E4583D205eb4ccCD648F1d78d
ExchangeBalances:  0x10DdF58989d23c10d18Cfe9e3EDe5d1C2A20F82a
ExchangeMode:  0xF1Eba8137fa9D614E31feb502904fcd03A972814
ExchangeAccounts:  0xef894f6DD3AB55f7905F53238904585F7dd7f8A5
ExchangeAdmins:  0x33ca3B7D7d2DF260aF3CCDA85e56cb7BF56181d7
ExchangeBlocks:  0x1918f7EE746d056C4426e7eF55024BD96928141c
ExchangeTokens:  0x94133D47eDD6cd6319C7aa04e0607A572880043B
ExchangeGenesis:  0x08B16f74b2f7ca626CC43318df5902D7547b754E
ExchangeDeposits:  0x9033617C3419dfCC45bc3F3F4042bAB656B2eD09
ExchangeWithdrawals:  0xc816e4d7BC330ba6e5587819aaf3049FEB4ec113
WETHToken: 0x5FEb64E2500dB1eF12E9Be96f86ecAF48E7a8c6e
LRCToken: 0x67acdFC08B1022F9f909822fBc11B6373932784C
ExchangeV3Deployer: 0xb4a04Fd82305C843753Cd569c929620a808E07c3

>>>>>>>> Deployed contracts addresses (deploy_protocol):
LoopringV3: 0x2DFa93fD3E177fbF4594E656BB0E94D38A6662de



  Contract: BlockVerifier
    owner
      ✓ should be able to register a new circuit (162ms)
      ✓ should not be able to register a circuit twice (200ms)
      ✓ should be able to disable a circuit (274ms)
      ✓ should not be able to disable a circuit that wasn't registered (51ms)
    anyone
      ✓ should not be able to register a new circuit (45ms)
      ✓ should not be able to disable a circuit (196ms)
    Block verification
       ✓ should be able to verify a single block with a valid proof (660ms)
      ✓ should not be able to verify a single block with an invalid proof (671ms)
      ✓ should not be able to verify a single block with invalid publicData (654ms)
      ✓ should not be able to verify a single block with publicData >= scalar field (48ms)
      ✓ should not be able to verify a block with an unknown circuit (88ms)
      ✓ should be able to verify multiple blocks of the same circuit with valid proofs (3276ms)
      ✓ should not be able to verify multiple blocks of different circuits with valid proofs (2615ms)
      ✓ should not be able to verify multiple blocks of the same circuits with invalid proofs (804ms)
      ✓ should not be able to verify multiple blocks of the same circuits with invalid public inputs (2438ms)
      ✓ should not be able to verify multiple blocks of the same circuits with public input >= scalar field (91ms)

  Contract: Claimable
    owner
      ✓ should be able to transfer ownership (89ms)
      ✓ should not be able to transfer ownership to an invalid address (47ms)
      ✓ should not be able to transfer ownership to the current owner (43ms)
    pending owner
      ✓ should be able to claim ownership (53ms)
    anyone
      ✓ should not be able to transfer ownership (41ms)
      ✓ should not be able to claim ownership (81ms)

  Contract: Compression
    LZ compression
      ✓ Test data (100ms)
      ✓ Random data (1779ms)
      ✓ n-byte sequences (3634ms)
      ✓ Zeros (68ms)
      ✓ Single byte data (47ms)
      ✓ Large data size (1068ms)

  Contract: Exchange
    Admin
      Exchange owner
        ✓ should be able to set the operator (4221ms)
        ✓ should be able to set the fees (4235ms)
        ✓ should not be able to set the withdrawal fee higher than allowed (3787ms)
      anyone
        ✓ should not be able to set the operator (3585ms)
        ✓ should not be able to set the fees (4579ms)

  Contract: Exchange
    Permutations
      ✓ Ring Settlement (529378ms)
      ✓ Deposit (203267ms)
      ✓ Onchain withdrawal (271831ms)
      ✓ Offchain withdrawal (619461ms)
      ✓ Order Cancellation (676551ms)

  Contract: Exchange
    Blocks
      Operator
        commitBlock
          ✓ should not be able to commit unsupported blocks (896ms)
          ✓ should not be able to commit block from different exchanges (881ms)
          ✓ should not be able to commit blocks starting from a wrong merkle root state (947ms)
          ✓ should not be able to commit settlement blocks with an invalid timestamp (1063ms)
          ✓ should not be able to commit settlement blocks with invalid protocol fees (1171ms)
          ✓ should not be able to commit deposit/on-chain withdrawal blocks with invalid data (4605ms)
          ✓ On-chain requests should be forced after MAX_AGE_REQUEST_UNTIL_FORCED (169224ms)
        verifyBlocks
          ✓ should be able to verify blocks in any order (122029ms)
          ✓ should not be able to verify a block more than once (192314ms)
          ✓ should be able to verify multiple blocks using the same circuit (1344403ms)
          ✓ should not be able to verify multiple blocks when one of the proofs is incorrect (882544ms)
          ✓ should not be able to verify multiple blocks using different circuits (173338ms)
          ✓ should not be able to verify the same block twice in a single call (172602ms)
          ✓ should not be able to verify a block too late (119995ms)
          ✓ should not be able to verify a block with an invalid proof (51131ms)
          ✓ should not be able to verify a block with incorrect public data (122703ms)
          ✓ should not be able to verify blocks with wrong call input format (4176ms)
        revertBlock
          ✓ Revert (211111ms)
        Block fee
          ✓ Withdraw block fee (deposit block - in time) (113407ms)
          ✓ Withdraw block fee (deposit block - fined) (113811ms)
          ✓ Withdraw block fee (deposit block - no reward) (112962ms)
          ✓ Withdraw block fee (withdrawal block - in time) (156542ms)
          ✓ Withdraw block fee (withdrawal block - fined) (154315ms)
          ✓ Withdraw block fee (withdrawal block - no reward) (155639ms)
          ✓ should not be able to withdraw a block fee of a block type without a block fee (178250ms)
      anyone
        ✓ shouldn't be able to commit blocks (3867ms)
        ✓ shouldn't be able to verify blocks (3754ms)
        ✓ shouldn't be able to revert blocks (3818ms)
        ✓ shouldn't be able to withdraw the block fee (3752ms)

  Contract: Exchange
    Cancel
      ✓ Cancel (orderToken != feeToken) (234871ms)
      ✓ Cancel (orderToken == feeToken) (192823ms)
      ✓ Cancel (owner == operator) (192715ms)
      ✓ Reuse trade history slot that was cancelled (192825ms)
      ✓ Cancel all orders from an account by changing the account's public key (35700ms)

  Contract: Exchange
    DepositWithdraw
      ✓ Create account (5186ms)
      ✓ Update account (4666ms)
      ✓ ERC20: Deposit (5927ms)
      ✓ ETH: Deposit (1504ms)

      ✓ Number of open deposits needs to be limited (4593225ms)
      ✓ Onchain withdrawal request (104289ms)
      ✓ Number of open onchain withdrawal requests needs to be limited (3348138ms)
      ✓ Offchain withdrawal request (token == feeToken) (124088ms)
      ✓ Offchain withdrawal request (token != feeToken) (123633ms)
      ✓ Offchain withdrawal request (owner == operator) (123701ms)
      ✓ Withdraw (normal account) (140778ms)
      ✓ Withdraw (protocol fee pool account) (166580ms)
      ✓ Deposits totalling the balance more than MAX_AMOUNT should be capped to MAX_AMOUNT (99571ms)
      ✓ Distribute withdrawals (by operator) (167339ms)
      ✓ Distribute withdrawals (not by operator) (165504ms)
      ✓ Distribute withdrawals in multiple parts (251777ms)
      exchange owner
        ✓ should be able to disable the depositing of a token (5116ms)
        ✓ should not be able to disable deposits for LRC/ETH/WETH (4484ms)
      anyone
        ✓ should not be able to disable/enable the depositing of a token (3765ms)

  Contract: Exchange
    Maintenance mode
      ✓ should not be able to commit settlement blocks while in maintenance mode (56942ms)
      exchange owner
        ✓ should be able to start and continue maintenance mode (4621ms)
        ✓ should be able to exit maintenance mode (3712ms)
      anyone
        ✓ should not be able to purchase downtime a token (826ms)

  Contract: Exchange
    Reentrancy
      ✓ registerToken (5515ms)
      ✓ setAddressWhitelist (5092ms)
      ✓ verifyBlocks (5259ms)
      ✓ disableTokenDeposit (5495ms)
      ✓ revertBlock (5500ms)
      ✓ withdrawProtocolFees (4531ms)
      ✓ withdrawFromMerkleTree (4589ms)
      ✓ commitBlock (4663ms)
      ✓ deposit (5067ms)
      ✓ withdrawFromMerkleTreeFor (5126ms)
      ✓ updateAccountAndDeposit (5970ms)
      ✓ setFees (7079ms)
      ✓ enableTokenDeposit (7286ms)
      ✓ withdraw (5831ms)
      ✓ startOrContinueMaintenanceMode (4764ms)
      ✓ burnExchangeStake (5904ms)
      ✓ withdrawProtocolFeeStake (6829ms)
      ✓ withdrawTokenNotOwnedByUsers (7584ms)
      ✓ depositTo (5392ms)
      ✓ withdrawFromDepositRequest (6460ms)
      ✓ setOperator (5429ms)
      ✓ withdrawBlockFee (5198ms)
      ✓ initialize (4719ms)
      ✓ withdrawExchangeStake (4708ms)
      ✓ stopMaintenanceMode (6164ms)
      ✓ distributeWithdrawals (4901ms)
      ✓ createOrUpdateAccount (4861ms)
      ✓ withdrawFromApprovedWithdrawal (5031ms)
      ✓ shutdown (4496ms)

  Contract: Exchange
    Trade
      ✓ Perfect match (151412ms)
      ✓ Matchable (orderA < orderB) (135114ms)
      ✓ Matchable (orderA > orderB) (138846ms)
      ✓ Market buy (single ring) (189222ms)
      ✓ Market buy (multiple rings) (220695ms)
      ✓ Market sell (single ring) (181678ms)
      ✓ Market sell (multiple rings) (196172ms)
      ✓ Buy order used as taker and maker (200261ms)
      ✓ Sell order used as taker and maker (201245ms)
      ✓ feeBips < maxFeeBips (122200ms)
      ✓ Rebate (single order) (161746ms)
      ✓ Rebate (both orders, rebate token == operator fee token) (166346ms)
      ✓ Unmatchable (119599ms)
      ✓ No funds available (118736ms)
      ✓ allOrNone (Buy, successful) (119786ms)
      ✓ allOrNone (Sell, successful) (121196ms)
      ✓ allOrNone (Buy, unsuccessful) (119726ms)
      ✓ allOrNone (Sell, unsuccessful) (119396ms)
      ✓ Self-trading (160094ms)
      ✓ selling token with decimals == 0 (118390ms)
      ✓ fillAmountB rounding error > 1% (119197ms)
      ✓ fillAmountB is 0 because of rounding error (118257ms)
      ✓ operator == order owner (118750ms)
      ✓ orderA.owner == orderB.owner == operator (159938ms)
      ✓ tokenS/tokenB mismatch (33264ms)
      ✓ tokenS == tokenB (32432ms)
      ✓ Wrong order signature (38039ms)
      ✓ should not be able to increase a balance to > MAX_AMOUNT (48579ms)
      ✓ validUntil < now (142462ms)
      ✓ validSince > now (128107ms)
      ✓ Multiple rings (182675ms)
      ✓ Order filled in multiple rings (order.orderID == tradeHistory.orderID) (181551ms)
      ✓ Reused OrderID (tradeHistory.filled > order.amount) (183829ms)
      ✓ Trimmed OrderID (order.orderID > tradeHistory.orderID) (180470ms)
      ✓ Cancelled OrderID (order.orderID < tradeHistory.orderID) (179482ms)

  Contract: Exchange
    Shutdown
      ✓ Withdraw exchange stake (288868ms)
      ✓ Incomplete shutdown (57093ms)
      ✓ Should not be able to shutdown when already shutdown (4298ms)

  Contract: Exchange
    Tokens
      ✓ token registration cost should be as expected (1371ms)
      ✓ LRC, ETH and WETH should be preregistered (1235ms)
      exchange owner
        ✓ should be able to register a token (1539ms)
        ✓ should not be able to register a token multiple times (1628ms)
      anyone
        ✓ should not be able to register a token (899ms)
        ✓ should not be able to get the token ID for an unregistered token (918ms)
        ✓ should not be able to get the token address for an invalid token ID (896ms)

  Contract: Exchange
    Withdrawal Mode
      ✓ should go into withdrawal mode when a deposit request isn't processed (1405ms)
      ✓ should go into withdrawal mode when a withdrawal request isn't processed (110994ms)
      ✓ should go into withdrawal mode when a block stays unverified (and is not reverted) (17795ms)
      ✓ should go into withdrawal mode when shutdown without reverting to initial state (54117ms)
      ✓ ERC20: withdraw from merkle tree (79346ms)
      ✓ ETH: withdraw from merkle tree (65879ms)
      ✓ Withdraw from merkle tree (protocol fee account) (136717ms)
      ✓ Withdraw from deposit block (92019ms)
      ✓ Should not be able to do any more block state changes (57543ms)

  Contract: Loopring
    Staking
      Exchange owner
        ✓ should be able to withdraw the protocol fee stake (930ms)
        ✓ should be able to lower the protocol fees (1837ms)
    Owner
      ✓ should be able to update settings (82ms)
    anyone
      ✓ should not be able to burn the stake (43ms)
      ✓ should not be able to withdraw the stake (42ms)
      ✓ should not be able to set the update the settings (84ms)
      ✓ should not be able to set the update the protocol fee settings (40ms)

  Contract: Operator
    Operator
      ✓ Commit and Verify (420253ms)
      ✓ Distribute withdrawals (242694ms)
      ✓ Withdraw block fee (58282ms)
      ✓ Revert (80796ms)

  Contract: Ownable
    owner
      ✓ should be able to transfer ownership (85ms)
      ✓ should not be able to transfer ownership to an invalid address (52ms)
    anyone
      ✓ should be able to get the owner
      ✓ should not be able to transfer ownership (66ms)

  Contract: Poseidon
    ✓ Poseidon t5/f6/p52 (60268ms)

  Contract: Transfers
    ERC20.transferFrom
      ✓ should succeed when a token transfer behaves normally (104ms)
      ✓ should succeed when a token transfer does not throw and returns nothing (101ms)
      ✓ should succeed when a token transfer returns nothing and the call succeeds (46ms)
      ✓ should fail when a token transfer 'require' fails (171ms)
      ✓ should fail when a token transfer returns false (170ms)
      ✓ should fail when a token transfer returns more than 32 bytes (138ms)
      ✓ should fail when a token transfer consumes more gas than allowed (116ms)
    ERC20.transfer
      ✓ should succeed when a token transfer behaves normally (173ms)
      ✓ should succeed when a token transfer does not throw and returns nothing (122ms)
      ✓ should succeed when a token transfer returns nothing and the call succeeds (46ms)
      ✓ should fail when a token transfer 'require' fails (330ms)
      ✓ should fail when a token transfer returns false (116ms)
      ✓ should fail when a token transfer returns more than 32 bytes (164ms)
      ✓ should fail when a token transfer consumes more gas than allowed (109ms)
    ETH
      ✓ should succeed when the recipient behaves normally (99ms)
      ✓ should fail when a transfer is done to a contract without fallback function (93ms)
      ✓ should fail when a transfer 'require' fails (92ms)
      ✓ should fail when a transfer consumes more gas than allowed (106ms)

  Contract: UserStakingPool
    stakeA
      ✓ should not stake if dont have any LRC (107ms)
      ✓ should not withdraw if haven't staked (70ms)
      ✓ set User A balance as 1000 (95ms)
      ✓ should get user staking equal to staked (93ms)
      ✓ should get total staking equal to staked
      ✓ should NEED_TO_WAIT when withdraw (43ms)
      ✓ advance block timestampe after MIN_WITHDRAW_DELAY
      ✓ set protocolfee as 100 (60ms)
      ✓ should get claimed equal as expected (97ms)
      ✓ should get withdrawed equal as expected (63ms)
    stakeA&B&C
      ✓ set User A balance as 1000 (92ms)
      ✓ set User B balance as 1000 (172ms)
      ✓ set User C balance as 1000 (179ms)
      ✓ should get user A staking equal to staked (97ms)
      ✓ advance block timestampe after 30 days
      ✓ should get user B staking equal to staked (96ms)
      ✓ advance block timestampe after 30 days
      ✓ should get user C staking equal to staked (91ms)
      ✓ advance block timestampe after MIN_WITHDRAW_DELAY: 90 days
      ✓ set protocolfee as 120 (60ms)
      ✓ should get user A claimed equal as expected (92ms)
      ✓ should get user B claimed equal as expected (90ms)
      ✓ should get user C claimed equal as expected (96ms)
      ✓ User A withdraw all (60ms)
      ✓ User B withdraw all (60ms)
      ✓ User C withdraw all (60ms)
      ✓ ProtocolFeeValut status should as expected
      ✓ advance block timestampe after 30 days
    stakeA&B&C
      ✓ set User A balance as 1000 (89ms)
      ✓ set User B balance as 1000 (177ms)
      ✓ set User C balance as 1000 (181ms)
      ✓ should get user A staking equal to staked (93ms)
      ✓ advance block timestampe after 30 days
      ✓ should get user B staking equal to staked (96ms)
      ✓ advance block timestampe after 30 days
      ✓ should get user C staking equal to staked (183ms)
      ✓ advance block timestampe after MIN_WITHDRAW_DELAY: 90 days
      ✓ set protocolfee as 120 (74ms)
      ✓ should get user A claimed equal as expected (102ms)
      ✓ should get user B claimed equal as expected (121ms)
      ✓ should get user C claimed equal as expected (98ms)
      ✓ User A withdraw all (68ms)
      ✓ User B withdraw all (62ms)
      ✓ User C withdraw all (81ms)
      ✓ ProtocolFeeValut status should as expected (43ms)
      ✓ advance block timestampe after 30 days


  258 passing (7h)
```

Brechtpd:
> It took me 7 hours to run all the tests on my latest MacBook Pro. Is this normal, @Brechtpd ?

I think so. There were/are some problems on Mac with release builds/multi-threading [which may be fixed recently](https://github.com/HarryR/ethsnarks/pull/138).

If these problems are fixed (I have no way to test this) it's better to build using `cmake-openmp-release` so multi-threading is enabled. You need to change this [here in the makefile](https://github.com/Loopring/protocols/blob/eb6b011f70539fdfc53a4a93b7b265e8e7202625/packages/loopring_v3/Makefile#L4).

On my Linux PC with 4 cores it takes about 2 hours to run all tests with multi-threading enabled, so your 7 hours single threaded makes sense.

Brechtpd:
Oh, [the makefile](https://github.com/Loopring/protocols/blob/eb6b011f70539fdfc53a4a93b7b265e8e7202625/packages/loopring_v3/Makefile#L16) even forces it to `cmake-debug` on Mac. That's even worse.

dong77:
> Oh, [the makefile](https://github.com/Loopring/protocols/blob/eb6b011f70539fdfc53a4a93b7b265e8e7202625/packages/loopring_v3/Makefile#L16) even forces it to `cmake-debug` on Mac. That's even worse.

I will change the line you mentioned to `CMAKE_TYPE := cmake-openmp-release` in this PR, but I wonder if the way we wrote the Makefile needs to be updated.

dong77:
Screen Shot 2019-08-09 at 22 45 55
Mac is not supported for multithreading I guess.

Brechtpd:
> Screen Shot 2019-08-09 at 22 45 55
> Mac is not supported for multithreading I guess.

I think that's just a cmake error that it can't find the OpenMP lib. You probably need to install it manually.

dong77:
> > Screen Shot 2019-08-09 at 22 45 55
> > Mac is not supported for multithreading I guess.
> 
> I think that's just a cmake error that it can't find the OpenMP lib. You probably need to install it manually.

I'll try to do it later.

dong77:
Damn, when I run `truffle migrate --network development` I see this error again:

```
>>>>>>> contracts deployed by deploy_exchange:
ExchangeData:  0x117dAca59429F40D7F67379E15515da71915cea9
ExchangeBalances:  0x9d5707f2F37F485EfC8F9FbfF54a142De75f620d
ExchangeMode:  0x6aD9Dbe01A680730db3B4228EEA7bEa79238809D
ExchangeAccounts:  0x0DAFac9cDAB5D638912187d68D00cE253639c6eD
ExchangeAdmins:  0x80c15093DC7059F54457104354CC3233598415D9
ExchangeBlocks:  0xecBE03E4c424c5d8FBe98345FF77c33Fa54768E5
ExchangeTokens:  0x565724Ea349bD72D45BB50D7B0E04dD9AeA2Ecc6
ExchangeGenesis:  0xe60Bc7DeeBD8f0cC70f60fC7d060101992B97dD8
ExchangeDeposits:  0x1B3AF808C5107c56CDd43B9E3b2D08C815DEb362
ExchangeWithdrawals:  0x4d6dbbCf4396094408E3253803177B5C6Ee77007
   ⠇ Blocks: 0            Seconds: 0
   Deploying 'ExchangeV3Deployer'
   ------------------------------
   ⠦ Blocks: 0            Seconds: 0WETHToken: 0x318eaB4Ed889eE827a39459d3Dedbbbf04266159
LRCToken: 0xC3959cF81314E17B8f138360982fDEd5016167F3
Error:  *** Deployment Failed ***

"ExchangeV3Deployer" ran out of gas (using a value you set in your network config or deployment parameters.)
   * Block limit:  0x201c061
   * Gas sent:     6700000

    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/truffle-deployer/src/deployment.js:364:1
    at processTicksAndRejections (internal/process/task_queues.js:89:5)
Truffle v5.0.15 (core: 5.0.15)
Node v12.1.0

```

dong77:
@Brechtpd @kongliangzhong  it turns out I have to change some `internal` functions inside `Exchange**.sol` libraries to `public` in order not to **inine** those functions into `ExchangeV3Deployer` to reduce the size of `ExchangeV3Deployer`.

dong77:
@Brechtpd @letsgoustc @kongliangzhong please take a final look :)

Brechtpd:
> @Brechtpd @kongliangzhong it turns out I have to change some `internal` functions inside `Exchange**.sol` libraries to `public` in order not to **inline** those functions into `ExchangeV3Deployer` to reduce the size of `ExchangeV3Deployer`.

Another way to lower the deployment cost is to lower the number of 'runs' in the optimizer [here](https://github.com/Loopring/protocols/blob/eb6b011f70539fdfc53a4a93b7b265e8e7202625/packages/loopring_v3/truffle.js#L14). This normally also increases the gas cost of actually using the functions, but  not making functions `internal` also increases the gas cost, so we'll have to experiment a bit to find the best parameters. But we can do that later.

dong77:
> > @Brechtpd @kongliangzhong it turns out I have to change some `internal` functions inside `Exchange**.sol` libraries to `public` in order not to **inline** those functions into `ExchangeV3Deployer` to reduce the size of `ExchangeV3Deployer`.
> 
> Another way to lower the deployment cost is to lower the number of 'runs' in the optimizer [here](https://github.com/Loopring/protocols/blob/eb6b011f70539fdfc53a4a93b7b265e8e7202625/packages/loopring_v3/truffle.js#L14) (delete all build files of contracts because truffle isn't always smart enough to recompile the contracts). This normally also increases the gas cost of actually using the functions, but not making functions `internal` also increases the gas cost, so we'll have to experiment a bit to find the best parameters. But we can do that later.

Good to know. But now we are good to go. Not sure what will happen after we introduce more features.

packages/loopring_v3/contracts/impl/LoopringV3.sol
letsgoustc: What I mean is not exchangeID check 0 or not. I mean we should first require exchangeAddress cannot be zero address ahead of calling safeTransferFrom.

We need switch the two require statement's sequence just like in withdrawExchangeStake.

packages/loopring_v3/contracts/impl/LoopringV3.sol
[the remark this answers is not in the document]
letsgoustc: just like comment for depositExchangeStake.

dong77:
@kongliangzhong can you review the IProtocolRegistry and ProtocolRegistry for me?

packages/loopring_v3/contracts/impl/LoopringV3.sol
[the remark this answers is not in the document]
dong77: also done

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -67,7 +67,7 @@ library ExchangeBlocks
         bytes  memory data,
         bytes  memory /*offchainData*/
         )
-        internal  // inline call
+        public
```
Brechtpd: It's important that `commitBlock` remains `internal` for efficiency, otherwise instead of passing all the DA data by reference we need to copy the data again. This costs quite a bit of gas even when using assembly, and this costs a lot of gas with the copy code generated by solidity.

Better to make another function `public` or decrease the `run` parameter of optimization if the deployment cost is still too high (if this is still the case with `cloning`).

dong77: unfortunately, making the method `internal` and reducing the run to `1` still end up more-than-block-gas-limit error.

dong77: let me try to fix this later in a separate PR.

dong77: Actually, I figured out a way to fix it.

PR #400, opened 2019-08-11
[Protocol3] Decouple exchange contract and loopring contract

dong77:
ILoopringV3 no longer creates IExchangeV3 instances.


dong77:
@kongliangzhong can you review the IProtocolRegistry and ProtocolRegistry for me?

PR #406, opened 2019-08-12
[Protocol3] Update Makefile to use release on Mac

dong77:


dong77:
@Brechtpd  let me run `npm run test` tonight to make sure this change also works for other tests.

dong77:
it works

PR #409, opened 2019-08-12
Add internal transfer related test script

yueawang:
Add functions and scripts to test internal transfer circuit logic. WIP...


dong77:
@yueawang is this PR ready to be reviewed? If so, please remove "WIP"

yueawang:
@dong77 not yet. This is only a half of the whole test process, and we need to add corresponding code in smart contract later.  Actually the other circuit PR is much closer to review. Thanks!

packages/loopring_v3/test/testExchangeUtil.ts
```diff
@@ -1948,6 +2037,169 @@ export class ExchangeTestUtil {
...
+        exchangeID,
+        currentBlockIdx + 1
+      );
+
+      const block = JSON.parse(fs.readFileSync(blockFilename, "ascii"));
+      const bs = new Bitstream();
+      bs.addNumber(block.exchangeID, 4);
+      bs.addBN(new BN(block.merkleRootBefore, 10), 32);
+      bs.addBN(new BN(block.merkleRootAfter, 10), 32);
+      for (const trans of block.internalTransferres) {
```
Brechtpd: Not sure why this is here as well as below. The only non-DA data is `exchangeID`, `merkleRootBefore`, `merkleRootAfter` and `labelhash` (though this value is also ignored on-chain, but we publish it always anyway because the cost is insignificant and constant for a block). Nothing needs to be stored on-chain for an offchain transfer (so this is different compared to a withdrawal because a user needs to be able to withdraw from the contract using on-chain data).

Brechtpd:
>This is only a half of the whole test process, and we need to add corresponding code in smart contract later. 

I don't think there's anything you need to do in the smart contract. As long as you use the same block type id value (`ExchangeData.BlockType.TRANSFER == 5` is already defined) that's it. `exchangeID`, `merkleRootBefore` and `merkleRootAfter` are handled the same for all blocks.

yueawang:
@Brechtpd, Do you mean the transfer block commit and verify is already supported? 
To test basic logic of circuit, yes, I don't need to touch the contract, but eventually we need to enable this block type (if necessary) and verify the proof onchain, i.e. make the test a close loop just like the other types. I assume it's missed in contract, that's why I said I did half of the whole job. Anyway it's good to me If the transfer block type is ready in contract :)
Thanks!

packages/loopring_v3/test/testExchangeUtil.ts
yueawang: Hi @Brechtpd, Thanks for the review! In fact, I have a question here: how to choose the DA, any criteria? I selected these members just like off-chain withdraw to quick enable/test my transfer circuit implementation. The value I picked somehow like a combination of ApprovedWithdrawalData and DataAvailabilityData in OffchainWithdrawGadget.
And here I have another question, what's the purpose of ApprovedWithdrawalData?
Thanks!

Brechtpd: > I have a question here: how to choose the DA, any criteria? I selected these members just like off-chain withdraw to quick enable/test my transfer circuit implementation. The value I picked somehow like a combination of ApprovedWithdrawalData and DataAvailabilityData in OffchainWithdrawGadget.

It's the minimal amount of data so that you can create new Merkle tree from the older Merkle tree + the DA data. So I think the data you've selected is correct.

> And here I have another question, what's the purpose of ApprovedWithdrawalData?

This data will be stored on-chain. It's important that this is stored on-chain because:
- A user can request a withdrawal on the contract, so this data needs to be available in the contract.
- This is needed for the 2 step `commitBlock`/`verifyBlocks`. This data will be stored on-chain in `commitBlock`, but the data can only be used by a user for a withdrawal once the block is finalized (otherwise the state is not guaranteed correct. An operator could commit an invalid block with too large a withdrawal amount, but the operator will not be able to create a proof for a block like this).

Brechtpd:
> Do you mean the transfer block commit and verify is already supported?

Yes, I don't think you will need to modify `commitBlock` and `verifyBlocks`.  `commitBlock` doesn't need to do anything other blocks don't need to do. `verifyBlocks` will also work as expected because the verification keys for the transfer circuit will be automatically registered in `BlockVerifier` if you use the existing functions for the testing.

packages/loopring_v3/operator/state.py
```diff
@@ -844,6 +875,58 @@ def cancelOrder(self,
...
+
+        # Update account From
+        rootBefore = self._accountsTree._root
+        accountBefore = copyAccountInfo(self.getAccount(accountFromID))
+        nonce = accountBefore.nonce
+        proof = self._accountsTree.createProof(accountFromID)
+
+        balanceUpdateF_From = self.getAccount(accountFromID).updateBalance(feeTokenID, -feeValue)
+        balanceT = int(self.getAccount(accountFromID).getBalance(transTokenID))
+        uAmountTrans = int(amountRequested) if (int(amountRequested) < balanceT) else balanceT
```
Brechtpd: I think this `min` logic can be removed because the circuit doesn't do this any more. The circuit will fail either way if this happens so in the end it doesn't really matter.

packages/loopring_v3/test/simulator.ts
```diff
@@ -266,6 +267,144 @@ export class Simulator {
...
+    operatorAccountID: number
+  ) {
+    const fee = roundToFloatValue(transfer.fee, constants.Float16Encoding);
+
+    const accountFrom = exchangeState.accounts[transfer.accountFromID];
+    let balanceFrom = accountFrom.balances[transfer.transTokenID].balance;
+    if (transfer.transTokenID === transfer.feeTokenID) {
+      balanceFrom = balanceFrom.sub(fee);
+    }
+    const amountToTrans = balanceFrom.lt(transfer.amount)
```
Brechtpd: Same thing here. This logic doesn't seem needed any more.

Brechtpd:
@dong77 I would merge this PR as soon as possible, even before #473 is done because that will need to modify the internal transfer code as well. This PR touches quite a lot of files in general so best not to keep this open to long otherwise we'll have new conflicts.

dong77:
I got it merged.

PR #410, opened 2019-08-12
[Lightcone_V2.js] Catch error when sending transaction

xiaowheat:


packages/lightcone_v2.js/src/lib/wallet/ethereum/walletAccount.ts
```diff
@@ -193,8 +193,10 @@ export class WalletAccount {
   //     throw Error('unimplemented');
   // }
 
-  sendTransaction(ethNode, signedTx) {
-    return ethNode.sendRawTransaction(signedTx);
```
xiaowheat: If we don't catch response of sendRawTransation, downstream request won't be able to catch the error. Then it will fail silently.

We don't have to use await and async here. Promise also works. If we use Promise, we will need resolve or reject to catch the response.

packages/lightcone_v2.js/src/wallet/privateKey.ts
```diff
@@ -142,12 +142,26 @@ export class PrivateKey {
    * @param gasPrice in gwei
    */
   public async depositTo(symbol: string, amount: number, gasPrice: number) {
```
xiaowheat: depositTo and withdrawFrom have a chain of API requests: exchange.deposit and this.account.sendTransaction
Using await and async is easier.

We can also use Promise here. However, we still need to catch errors.

packages/lightcone_v2.js/src/wallet/privateKey.ts
```diff
@@ -142,12 +142,26 @@ export class PrivateKey {
...
-    exchange
-      .deposit(this.account, symbol, amount, gasPrice)
-      .then((rawTx: Transaction) => {
-        const signedTx = this.account.signEthereumTx(rawTx.raw);
-        return this.account.sendTransaction(this.ethNode, signedTx);
```
xiaowheat: If we use Promise, we need to add resolve and reject inside then()

packages/lightcone_v2.js/src/sign/exchange.ts
```diff
@@ -223,15 +232,19 @@ export class Exchange {
...
+
+        const nonce = await ethereum.wallet.getNonce(this.getAddress());
+
+        const transaction = new Transaction({
           to: to,
           value: this.dexConfigurations.deposit_fee_eth,
           data: data,
           chainId: config.getChainId(),
-          nonce: fm.toHex(await ethereum.wallet.getNonce(this.getAddress())),
+          nonce: fm.toHex(nonce),
```
xiaowheat: This line of code doesn't change any functionalities.

PR #411, opened 2019-08-12
[Lightcone_V2.js] Update the value of gasLimit in exchange.ts in lightcone_v2.js

xiaowheat:
If we use ```config.getGasLimitByType("eth_transfer").gasLimit```, it will return an error: 
```
"Error: base fee exceeds gas limit
    at /usr/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:61:1851624
    at /usr/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:2:21647
    at process._tickCallback (internal/process/next_tick.js:61:11)"
```

### DIFF
--- a/packages/lightcone_v2.js/src/sign/exchange.ts
+++ b/packages/lightcone_v2.js/src/sign/exchange.ts
@@ -25,6 +25,8 @@ import {
   SimpleOrderCancellationReq
 } from "../grpc/proto_gen/service_dex_pb";
 
+const MOCK_EXCHANGE_GAS_LIMIT = "0xf0f09f0";
+
 // TODO: Rename Exchange
 export class Exchange {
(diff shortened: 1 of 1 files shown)


packages/lightcone_v2.js/src/sign/exchange.ts
```diff
@@ -25,6 +25,8 @@ import {
   SimpleOrderCancellationReq
 } from "../grpc/proto_gen/service_dex_pb";
 
+const MOCK_EXCHANGE_GAS_LIMIT = "0xf0f09f0";
```
xiaowheat: We are using a relative large value. Unless, there is a better value. @iwbrhwfy @kongliangzhong

PR #430, opened 2019-08-13
[Protocol3] Use external function in libexchange/*.sol to reduce gas cost

dong77:


dong77:
Here are some comparisons when I change the `commitBlock` in *ExchangeBlocks.sol*  to `external`, `public`, and `internal`.

Screen Shot 2019-08-13 at 16 15 56

Brechtpd:
Better than I expected!

dong77:
> Better than I expected!
Not too bad :)

PR #438, opened 2019-08-16
[Protocol3] use a Proxy to reduce the gas of creating a DEX without upgradability

dong77:
Creating a plain exchange instance is very expensive.


dong77:
> This will decrease the cost of creating an exchange, but it will increase the cost of using the exchange. A proxy adds at least ~700 gas to every function call because it needs to forward every call to the implementation contract, which without upgradeability isn't needed. So the previous method will certainly end up using less gas than with a proxy contract, but with larger initial cost.

I will re-consider this PR.

PR #445, opened 2019-08-17
[Protocol3] Reimplement protocol and exchange registry

dong77:


packages/loopring_v3/contracts/lib/Claimable.sol
```diff
@@ -27,6 +27,8 @@ contract Claimable is Ownable
 {
     address public pendingOwner;
 
+    constructor() public Ownable() {}
```
Brechtpd: Is this so `owner` isn't set in the constructor in the implementation contract when used with a proxy? If so, why is this done here instead of `Ownable` itself? Maybe it even makes sense to make it a parameter like `setOwner` for the constructor to set the owner or not so we can reuse `Ownable` for use with a proxy and not

packages/loopring_v3/contracts/impl/UserStakingPool.sol
```diff
@@ -195,10 +195,7 @@ contract UserStakingPool is Claimable, ReentrancyGuard, IUserStakingPool
 
         staking.claimedAt = (staking.claimedAt == 0) ?
             staking.depositedAt :
-            staking.balance
-                .mul(staking.claimedAt)
-                .add(additionalBalance.mul(now)) / balance;
-
+            staking.balance .mul(staking.claimedAt) .add(additionalBalance.mul(now)) / balance;
```
Brechtpd: Strange formatting with empty space before the dot.

packages/loopring_v3/contracts/lib/Claimable.sol
dong77: My understanding is this: currently (before this change), Claimable is inheried by some contracts, and in their constructors, we call `Claimable()`. But Claimable.sol's own constructor does not call Ownable's construct, I assume this will end up the `owner` value is not initialized for those contracts. There are two ways to fix it: 1) force Caimable contract's constructor to call `Ownable()`, or change all other contracts that inherit Claimable to call `Ownerable()` in their constructors.

packages/loopring_v3/contracts/impl/UserStakingPool.sol
dong77: restored

packages/loopring_v3/contracts/lib/Claimable.sol
Brechtpd: Oh I read this wrong, I though this redefined the `Ownable` constructor. 

I did some additional testing to be sure and it's not necessary to specify the constructor call to the base class if the base class contructor doesn't have any parameters. `Claimable` doesn't define any constructor so the default one will be generated, but even this default constructor will correctly initialize `Ownable`.

The examples with `Owned` do the same thing in the solidity docs:
- https://solidity.readthedocs.io/en/v0.5.11/contracts.html#inheritance

So with or without this addition `owner` is initialized correctly. Probably a good idea to add this anyway so we don't depend on this behaviour and make it easier to understand.

dong77: It seems my modification is not necessary. Reverted.

PR #454, opened 2019-08-24
[Protocol 2] Re-Implement Brokers for Dolomite

zrubenst:
An order with the `broker` field set to a contract that inherits `IBrokerDelegate` will have the protocol call the function `brokerRequestAllowance` _(Supplying order/transfer information)_ to request approval from a broker to transfer the needed tokens to complete an order.

Use cases include using `CREATE-2` to enable smart wallet functionality _(eliminating the end-user's need for Ether, as wrapping and approvals are no longer needed)_ and sourcing liquidity from on-chain interfaces without having to front any tokens to complete an order fill _(This is made possible by having the broker transfers occur after the other batched transfers)_.

On top of these changes, the version of Solidity was increased to 0.5.7 as the rest of our codebase uses this version for other smart contracts.

### DIFF
--- a/packages/loopring_v2/contracts/helper/OrderHelper.sol
+++ b/packages/loopring_v2/contracts/helper/OrderHelper.sol
@@ -14,13 +14,13 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
-pragma solidity 0.5.2;
+pragma solidity 0.5.7;
 
 import "../impl/Data.sol";
 import "../lib/ERC20.sol";
 import "../lib/MathUint.sol";
 import "../lib/MultihashUtil.sol";
-
+import "../iface/IBrokerDelegate.sol";
 
 /// @title OrderHelper
 /// @author Daniel Wang - <daniel@loopring.org>.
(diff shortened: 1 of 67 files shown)


packages/loopring_v2/contracts/helper/RingHelper.sol
```diff
@@ -340,51 +340,93 @@ library RingHelper {
...
+                ctx.transferPtr,
+                p.order.tokenS,
+                p.order.owner,
+                prevP.order.tokenRecipient,
+                amountSToBuyer
+            );
+        } else {
+            
+            // Calculates amount received from other participant
+            uint amountBFromSeller = prevP.fillAmountS
```
zrubenst: @Brechtpd Is this the correct way to calculate how much this participation is receiving from the other? We brought the max participants in a ring down to 2, so I figured taking the logic to calculate how much this participant and sending the other and flipping it would do the trick

Brechtpd: I think so yes. Using the other order's `amountSToBuyer` is the amount the order actually receives, which seems to be correctly calculated here.

packages/loopring_v2/contracts/test/DummyBrokerDelegate.sol
```diff
@@ -0,0 +1,44 @@
...
+    address receivedToken,
+    uint receivedAmount,
+    address orderTokenRecipient,
+    address requestedToken, 
+    uint requestedAmount, 
+    address requestedRecipient,
+    bytes memory extraOrderData,
+    bool isFee
+  ) public {
+    address payable payableTokenAddress = address(uint160(requestedToken));
```
Brechtpd: Not that it matters, but this seems unnecessary.

packages/loopring_v2/contracts/helper/OrderHelper.sol
```diff
@@ -423,6 +413,7 @@ library OrderHelper {
     {
         if (!tokenSpendable.initialized) {
```
Brechtpd: Make sure the spendables are correctly set. There could be some strange combinations (order with/without a broker but with the same owner) but probably not really a problem in practice because that won't happen.

PR #464, opened 2019-08-27
Fast withdrawal impl

kongliangzhong:


Brechtpd:
It seems like this will need a lot of onchain transactions:
- The fast withdrawal request
- The fast withdrawal transfer
- The withdrawal that transfers to an operator address

I would keep the fast withdrawal request offchain, this will be cheaper and faster (a fast withdrawal can be done this way in the max time to include the transfer in an Ethereum block, so max 15 seconds if so desired):
1. The user signs a fast withdrawal request and sends it to the DEX
2. The DEX, if it wants to (ignoring the request is allowed), transfers the tokens to the user using the cookie jar contract. A smart contract function does the transfer and stores at least the user's account ID and the amount that was transferred (similar to deposits).
3. The DEX can now use the offchain fast withdrawal request in a block. The input for this block is the data that was stored onchain in step2 (could be done similarly to deposits/onchain withdrawals), this makes sure all of this is trustless and the amounts transferred match. This does an offchain transfer from the user's balance to the cookie jar account (or any other account held by the DEX could also work I guess).

This will drain the onchain balances of the cookie jar contract. To replenish the cookie jar contract the DEX simply does a withdrawal on the offchain account that receives the funds when wanted/necessary. In the best case the owner of this account is also the cookie jar contract so only a single onchain transfer is needed for this, but this is not required.

kongliangzhong:
> It seems like this will need a lot of onchain transactions/transfers:
> 
> * The fast withdrawal request
> * The fast withdrawal transfer
> * The withdrawal that transfers to an operator address
> 
> I would keep the fast withdrawal request offchain, this will be cheaper and faster (a fast withdrawal can be done this way in the max time to include the transfer in an Ethereum block, so max 15 seconds if so desired):
> 
> 1. The user signs a fast withdrawal request and sends it to the DEX
> 2. The DEX, if it wants to (ignoring the request is allowed), transfers the tokens to the user using the cookie jar contract. A smart contract function does the transfer and stores at least the user's account ID and the amount that was transferred (similar to deposits).
> 3. The DEX can now use the offchain fast withdrawal request in a block. The input for this block is the data that was stored onchain in step2 (could be done similarly to deposits/onchain withdrawals), this makes sure all of this is trustless and the amounts transferred match. This does an offchain transfer from the user's balance to the cookie jar account (or any other account held by the DEX could also work I guess).
> 
> This will drain the onchain balances of the cookie jar contract. To replenish the cookie jar contract the DEX simply does a withdrawal on the offchain account that receives the funds when wanted/necessary. In the best case the owner of this account is also the cookie jar contract so only a single onchain transfer is needed for this, but this is not required.

@Brechtpd , By this way, we have to add a new BlockType and  circuit , right?

Brechtpd:
We could probably make the offchain transfer circuit general enough to support both _normal_ offchain tranfers and these _conditional_ transfers. This will make normal offchain transfers slightly more expensive because the circuit will have to do a bit more work than needed. I estimate a couple of 1000s of constraints more because of extra hashing of data (the on-chain data we need to check if the operator transferred the correct amount). Not 100% sure of this yet.

But I find the current solution also a bit complex. Quite a lot of data needs to be stored on-chain and some extra checks are needed on-chain for every on-chain withdrawal. The operator may not want to do the fast withdrawal, but the user already paid the higher fee when doing the on-chain fast withdrawal request, etc...

Brechtpd:
I think we should be able to do something similar for off-chain transfers as we do with on-chain withdrawals. We reuse the on-chain withdrawal circuit when in shutdown to do all withdrawals, but we're in a special mode that doesn't check if the ending hash matches the one on-chain. This allows the operator to do any withdrawal he wants without having on-chain requests by the user. 

Let's say the off-chain request data for off-chain transfers and fast withdrawals is the same (with an extra member that says if this request is conditional or not). The request always needs to be signed by the user and the same data is hashed, so no difference between offchain transfers and fast withdrawals for this.
The only difference is that for fast withdrawals we need to check if the amount withdrawn matches with the amount that was transferred on-chain by the DEX. For normal offchain transfers we can ignore this.

So to keep things simple let's say we can only do normal transfers or fast withdrawals in a block, not both (though there are ways to allow this as well I think). The only thing we have to do different is comparable with withdrawals in shutdown mode:
- For fast withdrawals the ending hash of the onchain data needs to match with the value calculated on-chain.
- For offchain transfers we don't check the ending hash, the operator calculates the ending hash as needed to make the circuit valid with the given offchain transfers.

There's probably ways to improve on this further, but this should be a good start.

Brechtpd:
A easier way to reuse the withdrawal circuit may be something like this:
- The user signs the fast withdrawal request using his Ethereum keypair
- The user sends the request/signature to the operator
- The operator decided to do the fast withdrawal or not (the user actually has the balance he wants to withdraw, the cookie jar contract has enough funds left to do it, ...)
- If he wants to do it:
- The operator does an on-chain withdrawal request for the user. The signature of the fast withdrawal request of the user is also sent on-chain and this signature is verified. If the signature is valid the on-chain withdrawal request is done AND the amount the user wants to withdraw is immediately sent to the user from the cookie jar contract.
- The on-chain withdrawal request is processed like normal, but the destination address for the approved withdrawal data is changed to an address provided by the DEX when doing the transfer.

I think this solves some of the issues with the approach in this PR. For efficiency this will probably still limit the user to pay a fee in ETH (doing yet another ERC20 transfer would be expensive), but this fee can be user-specific.

This approach still does 2 onchain transfers instead of 1. There are also some extra costs associated with the on-chain signature validation and the on-chain withdrawal request. With the off-chain transfer approach we're very close to the minimum work we can do on-chain, so it is quite a bit more scalable than this (and thus also cheaper). The user would also be able to pay a fee in any token he wants. It would also allow for other use cases of conditional payments using on-chain data. So I think I would still prefer the off-chain transfer one.

kongliangzhong:
I would rather prefer the last solution instead of the off-chain transfer one because the user's assets are transferred to the operator's account in the Merkle tree. So the operator has to send on-chain withdrawals regularly. With the on-chain withdraw solution, the tokens are sent directly to the operator or some cookie jar contract.
The user send ethereum tx off-chain to our DEX is brilliant so the DEX can reject the fast-withdrawal request and if so the user will not pay fast-withdrawal fees.
Pay the fast withdrawal fee in ETH is not a problem for me, for we only support ETH fee for deposit/onchain-withdrawal in the same way.

kongliangzhong:
To make things clear, In the onchain-withdraw solution, you merge the following two onchain request:
1. user request a fast withdrawal, and pay fast-withdraw fee by ETH 
2. operator proceed the fast-withdraw request, calculate the actual amount it can withdraw and pay user immediately.
into the following one:
*  The operator pay user, use the data signed by user's ethereum private key to deduct the fast-withdraw fee, and add this request to withdrawal chain data.

I think this would be work, I need to do some experiments.

Brechtpd:
> I would rather prefer the last solution instead of the off-chain transfer one because the user's assets are transferred to the operator's account in the Merkle tree. So the operator has to send on-chain withdraw requests regularly. With the on-chain withdraw solution, the tokens are sent directly to the operator.

That's true, but the on-chain withdrawal request method will be ~3x more expensive (maybe even more).  Especially for small withdrawal amounts (which should be the large majority of the withdrawals) where the cookie jar contract can be used many times even with limited funds without recharging in it there will be significant cost savings:
- An extra token transfer: ~20,000 gas
- More data needs to be stored on-chain for the request (fee, timestamp): ~20,000-60,000 gas (depends on how this will be implemented)
- More data needs to be stored on-chain for the approved withdrawal data: ~5000 gas
- The withdrawal needs to be distributed by the operator: Storage needs to be updated so the withdrawal can't be done twice (gas cost harder to estimate, updating the storage has a cost, but there will also be a refund for resetting the data to 0). The operator/DEX also pays for token transfer here, so the fee charged to the user will also include the cost to do this already.

With the off-chain transfer method it could make sense for a DEX to do ALL withdrawals using the fast withdrawal mechanism if possible because it may actually be cheaper than doing a normal withdrawal (though probably only a bit less expensive for off-chain withdrawal requests, should be comparable in any case with much improved user experience)! The proving cost will be slightly different but should be similar enough.

Also, there seems to be some hype about conditional payments, though I don't really know what the practical use cases are so I can't say if the off-chain transfer method is powerful enough to support those.

kongliangzhong:
@Brechtpd , Ok, You convinced me. I will follow the off-chain transfer way.

kongliangzhong:
This PR depends on the inner transfer PR, So I will close it and open a new one.

PR #467, opened 2019-08-30
[Protocol-2] Changes tokenB fee behavior, fixes issue with Ring Mined events, makes broker transfers more efficient

zrubenst:
* Previously if the `owner` != `tokenRecipient` and `feeToken` == `tokenB`, the fee would not be taken out of the received tokens and instead the owner would need to hold the tokens in their balance. This is being changed to follow the normal behavior even of the tokenRecipient is not the order owner

* In the last set of changes the Participation struct was modified and was being encoded improperly  when inline assembly constructed the Ring Mined event. The Participation struct is now being left untouched and a different method of getting information to the RingSubmitter for broker transfers has been implemented

### DIFF
--- a/packages/loopring_v2/contracts/helper/OrderHelper.sol
+++ b/packages/loopring_v2/contracts/helper/OrderHelper.sol
@@ -278,6 +278,7 @@ library OrderHelper {
                 order.hash,
                 order.sig
             );
+            require(order.valid, 'INVALID_SIGNATURE');
         }
     }
 
@@ -294,6 +295,7 @@ library OrderHelper {
                 miningHash,
                 order.dualAuthSig
             );
+            require(order.valid, 'INVALID_DUAL_AUTH_SIGNATURE');
         }
     }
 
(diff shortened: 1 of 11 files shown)


dong77:
@zrubenst do you still want to merge this PR? It's been approved for quite a while.

zrubenst:
> @zrubenst do you still want to merge this PR? It's been approved for quite a while.

Will be adding in some more minor changes in the next few days (Adding better revert messages, as we are submitting rings on a one-taker-basis, we need to know exactly why a revert happens rather than simply emitting a ring failed event).

zrubenst:
Added in new Commit:

- More verbose revert messaging so we know what is going wrong when an order fails to be matched properly on-chain

- Overhaul of the IBrokerDelegate and the way transfers are performed via brokers. Now transfers are batched together & combined if possible, the number of calls to the IBrokerDelegate implementation has been dramatically reduced and more information about the orders has been given to the IBrokerDelegate implementation so it can make better decisions on doing it's own form of batching transfers and other calls that. Ultimately, these changes have reduced gas consumption from broker trades **considerably**. Before these changes each broker trade (trade between broker and broker) costed, on average, 400,000 gas. Now this has been reduced to 180,000. Even less if all (or most of) the makers come from the same (or two) broker(s).

zrubenst:
@dong77 @Brechtpd

PR #470, opened 2019-09-03
[Protocol3] Browser lib

Brechtpd:
Still needs some clean up. Reverts (and probably some other minor things) are also not yet supported.


packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,188 @@
...
+    tokenID: number;
+    amountRequested: BN;
+    amountWithdrawn?: BN;
+
+    transactionHash: string;
+  }
+
+  export interface Trade {
+    blockIdx: number;
+    requestIdx: number;
```
dong77: what is this Idx? is this the per-account nonce?

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    accountIdA: number;
```
dong77: should we use M(maker) to replace A and T(taker) to replace B, not only here but also in the protocol?

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  export interface OffchainWithdrawal {
+    blockIdx: number;
+    requestIdx: number;
+
+    accountID: number;
+    tokenID: number;
+    amount: BN;
+    feeTokenID: number;
+    fee: BN;
```
dong77: Do we need the following fields:
```
  amountRequested: String,
        amountWithdrawn: String,
        transactionHash: String,
```

packages/loopring_v3.js/src/SparseMerkleTree.ts
```diff
@@ -0,0 +1,125 @@
+import poseidon = require("./poseidon");
```
dong77: We probably don't need to construct any tree when we process the onchain data. So the tree will be flattened to its' leaves, so we'll have a table for balance, a table for trade history. Only when a user requests to construct the Merkle root, then we calculate on demand.

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  export enum ForgeMode {
+    AUTO_UPGRADABLE = 0,
+    MANUAL_UPGRADABLE,
+    PROXIED,
+    NATIVE,
+  }
+
+  export interface Block {
+    blockIdx: number;
```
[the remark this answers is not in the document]
Brechtpd: That's a good idea.

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,188 @@
...
+    tokenID: number;
+    address: string;
+    enabled: boolean;
+  }
+
+  export interface Deposit {
+    blockIdx?: number;
+    requestIdx?: number;
+
+    depositIdx: number;
```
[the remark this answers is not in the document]
Brechtpd: `requestIdx` is the index of the request in the list of ALL requests that were processed in all blocks, `depositIdx` is the on-chain index in the deposit queue. So if on an exchange there were 4 deposits and 4 trades than the last trade would have `requestIdx == 7` because it's the 8th request that was processed. Currently I store all requests in a single list (so also perhaps a single database table) which contains all requests of all types (deposits, withdrawals, trades,...) in the order they were processed by blocks. You can get them with `getProcessedRequests(startIdx: number, count: number)` (or `getRequestsInBlock(blockIdx: number)` per block).

packages/loopring_v3.js/src/types.ts
Brechtpd: See above.

packages/loopring_v3.js/src/types.ts
Brechtpd: The naming with `A` and `B` predates the taker/maker decision and it would take a lot of refactoring work to change it. Naming them `takerOrder` and `makerOrder` would also work I think. It would certainly be an improvement to rename them, but it will take some time to do all the refactoring.

packages/loopring_v3.js/src/types.ts
Brechtpd: `amountRequested` is the amount that was requested by the user in the on-chain withdrawal request, `amountWithdrawn` is what was actually withdrawn. So I think `amountWithdrawn` is definitely needed, and `amountRequested` I think could be useful to help a user find the withdrawal he requested on-chain (it could be confusing to see a different value than requested if it doesn't really match with `amountWithdrawn` because his balance was lower).

`transactionHash` is the Ethereum transaction hash the request/block was done. So this allows us to link to an Ethereum block explorer to see the actual Ethereum transaction, which I think would be useful.

packages/loopring_v3.js/src/SparseMerkleTree.ts
Brechtpd: The Merkle tree is only built when requested using the API (`buildMerkleTree` or `buildMerkleTreeForWithdrawalMode`). The Merkle tree does NOT get build incremental while the blocks are being processed and the leaf data is being updated.

packages/loopring_v3.js/src/types.ts
dong77: ok, I'm fine without changing the names. We just need to memorize A is maker and B is taker.

packages/loopring_v3.js/src/types.ts
dong77: The comment is regarding the OffchainWithdrawal definition, I think it misses the fields you described.

packages/loopring_v3.js/src/SparseMerkleTree.ts
dong77: i see. this is also my understanding

packages/loopring_v3.js/src/types.ts
Brechtpd: For off-chain withdrawals we don't have the `amountRequested` data on-chain because it's not needed (what was requested is not needed to update the Merkle tree, only what was actually withdrawn). The request is also completely off-chain so there is no `transactionHash` which could point to the Ethereum transaction in which the request was made.

packages/loopring_v3.js/src/constants.ts
```diff
@@ -0,0 +1,50 @@
...
+    numBitsExponent: 5,
+    numBitsMantissa: 11,
+    exponentBase: 10
+  };
+  static readonly Float12Encoding: FloatEncoding = {
+    numBitsExponent: 5,
+    numBitsMantissa: 7,
+    exponentBase: 10
+  };
+  static readonly Float8Encoding: FloatEncoding = {
```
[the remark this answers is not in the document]
Brechtpd: Float12 and Float8 are not used anywhere yet, I made these while I was experimenting with the precession. I will remove them.

I think float is the correct term for them, see 
- https://en.wikipedia.org/wiki/Decimal_floating_point

We don't have a sign bit, so the exponent is always positive (so no fractions which would be possible with e.g. `10**-2`) but they are pretty standard decimal floating point numbers.

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,412 @@
...
+/**
+ * Trade history data.
+ */
+export interface TradeHistory {
+  /** How much the order is filled. */
+  filled: BN;
+  /** Wheter the order is cancelled. */
+  cancelled: boolean;
+  /** The orderID of the order the trade history is currently stored for. */
+  orderID: number;
```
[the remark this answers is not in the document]
Brechtpd: I guess it would possible to just dump all balances/trade history for an account as json in the database entry of the account because the amount of data for balances/trade history for a single account will in almost all cases be very, very small. 

I have no experience with databases so if those values make sense for how you want to store this data in the database I'll add them.

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/**
+ * Balance data.
+ */
+export interface Balance {
+  /** How amount of tokens the account owner has for a token. */
+  balance: BN;
+  /** The trade history data. */
+  tradeHistory: { [key: number]: TradeHistory };
```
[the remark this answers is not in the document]
Brechtpd: I kind of expected this, but I was unsure how it would actually work best with the database so I left it like this.

I will change it to use separate lists if that's how you want to do it in the database.

packages/loopring_v3.js/src/types.ts
```diff
@@ -0,0 +1,412 @@
...
+  /** The fee charged (in ETH) for depositing. */
+  depositFeeETH: BN;
+  /** The fee charged (in ETH) for an on-chain withdrawal request. */
+  withdrawalFeeETH: BN;
+}
+
+/**
+ * The protocol fees that need to be paid for trades.
+ */
+export interface ProtocolFees {
```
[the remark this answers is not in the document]
Brechtpd: They are the latest state. The events aren't exposed/stored because that data cannot be reverted (excluding Ethereum reorgs of course).

packages/loopring_v3.js/src/eddsa.ts
```diff
@@ -0,0 +1,76 @@
+// Taken and modified from
+// https://github.com/iden3/circomlib
+
+const createBlakeHash = require("blake-hash");
+import crypto = require("crypto");
```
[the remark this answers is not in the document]
Brechtpd: Yes, will change

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+   */
+  public async sync(ethereumBlockTo: number) {
+    if (ethereumBlockTo <= this.syncedToEthereumBlockIdx) {
+      return;
+    }
+
+    // Process the events
+    const events = await this.loopring.getPastEvents("allEvents", {
+      fromBlock: this.syncedToEthereumBlockIdx + 1,
+      toBlock: ethereumBlockTo
```
dong77: if the number of blocks is too many, will this cause an out-of-memory error? Are we supposed to sync to only a couple of blocks per call? if so, we probably want to check `toBlock - fromBlock <= MAX_BLOCK_PER_CALL`

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+    });
+    for (const event of events) {
```
dong77: maybe we can use `.then(...)` instead of await?

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+   * Returns the amount of LRC staked as protocol fee stake for an exchange
+   * @param   exchangeId   The exchange ID of the exchange to get the protocol fee stake for
+   * @return  The amount staked in LRC
+   */
+  public getProtocolFeeStake(exchangeId: number) {
+    this.stakeAmounts[exchangeId] = this.stakeAmounts[exchangeId] || {
+      exchangeStake: new BN(0),
+      protocolFeeStake: new BN(0)
+    };
+    return this.stakeAmounts[exchangeId].protocolFeeStake;
```
dong77: maybe do this 
```
return this.stakeAmounts[exchangeId] || DEFAULT // DEFAULT should be a predefined, resuable const.
```

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+  private async processExchangeStakeDeposited(event: any) {
+    const exchangeId = parseInt(event.returnValues.exchangeId);
+    const amount = new BN(event.returnValues.amount);
+    this.stakeAmounts[exchangeId] = this.stakeAmounts[exchangeId] || {
+      exchangeStake: new BN(0),
+      protocolFeeStake: new BN(0)
+    };
+    this.stakeAmounts[exchangeId].exchangeStake = this.stakeAmounts[
+      exchangeId
+    ].exchangeStake.add(amount);
```
dong77:
```
var staking = this.stakeAmounts[exchangeId] || DEFAULT;
staking.exchangeStake = staking.exchangeStake.add(amount);
this.stakeAmounts[exchangeId] = staking;
```

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+
+interface StakeAmounts {
+  exchangeStake: BN;
+  protocolFeeStake: BN;
+}
+
+/**
+ * Processes all data of the Loopring v3 protocol contract.
+ */
+export class ProtocolV3 {
```
dong77: I guess I still need to figure out how this class shall be used. We should probably populate data into a database. I'll keep reading the other code and get back to this comment.

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+    this.version = version;
+    this.enabled = true;
+
+    this.syncedToEthereumBlockIdx = 0;
+
+    const ABIPath = "ABI/version30/";
+    this.loopringAbi = fs.readFileSync(ABIPath + "ILoopringV3.abi", "ascii");
+
+    this.loopring = new web3.eth.Contract(JSON.parse(this.loopringAbi));
+    this.loopring.options.address = this.loopringAddress.toLowerCase();
```
dong77: seems this "options" is not used anywhere else.

packages/loopring_v3.js/src/protocol_v3.ts
```diff
@@ -0,0 +1,232 @@
...
+        await this.processExchangeStakeBurned(event);
+      } else if (event.event === "ProtocolFeeStakeDeposited") {
+        await this.processProtocolFeeStakeDeposited(event);
+      } else if (event.event === "ProtocolFeeStakeWithdrawn") {
+        await this.processProtocolFeeStakeWithdrawn(event);
+      } else if (event.event === "OwnershipTransferred") {
+        await this.processOwnershipTransferred(event);
+      }
+    }
+    this.syncedToEthereumBlockIdx = ethereumBlockTo;
```
dong77: can this method return a list of events extracted?

packages/loopring_v3.js/src/request_processors/deposit_processor.ts
```diff
@@ -0,0 +1,71 @@
...
+    const length = data.extractUint32(offset + 4);
+
+    const deposits: Deposit[] = [];
+    for (let i = startIdx; i < startIdx + length; i++) {
+      const deposit = state.deposits[i];
+
+      this.processDeposit(state, deposit);
+
+      deposit.blockIdx = block.blockIdx;
+      deposit.requestIdx = state.processedRequests.length + i - startIdx;
```
dong77: should we push this deposit into `state.processedRequests`, if so, maybe we can do:

```
state.processedRequests.push(deposit);
deposits.push(deposit);
deposit.requestIdx = state.processedRequest.length;
```

packages/loopring_v3.js/src/request_processors/deposit_processor.ts
```diff
@@ -0,0 +1,71 @@
...
+      deposits.push(deposit);
+    }
+    return deposits;
```
dong77: I also think this class shall emit events so this class's caller can persist the events and maybe states into a DB o choice. 

In general, these processors CAN maintain in-memory state after loading a state-snapshot from DB upon start/restart, then they should omit new events and new state-values for callers to persist.

dong77: The database may have two types of tables, one for events:
```
event_type, block_idx, request_idx, ... other data
```
and for each event type a state table 
```
state_type, block_idx ... other accumulative data
```

Event `Deposit`/`Withdrawl`'s corresponding state is `Balance`
Event `Trade`/`Cancel`'s corresponding state is `Order`

---

When we revert block N+1, we delete all events data with block_idx >= N+1 and all states wich block_idx >= N+1,
then we can reload the state with block_idx === N, ....

---

The caller of these processors can keep up to most recent M state records but all event records.

---

The state shall be calculated by one single thread in memory based on the last state and events in the current block. But we probably want to query last state from DB on demand instead of loading all of them in the memory -- it may become too large.

dong77: I think we can merge the PR, then let me define the DB access API, then you can use the API to load and write data.

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: I think it can go out of memory, but a maximum can be hard to decide on the library level because the amount of memory that can be used is by the library is hard to predict. . I think the code using the library is better at deciding this and can do small ranges to play safe and increase it if memory use is sufficiently low.

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: I think it's probably better to do that at the `sync` level  because all events need to be processed sequentially? Will have to look up the differences to make sure.

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: Yes, that's better!

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: Maybe some of the code can be done elsewhere, but what the code does needs to be done somewhere. Doing the processing of the data here as well made sense to me, but we'll see.

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: It's the address that is used for the web3 contract used to do the calls on the smart contract and get the events. It's used internally in the `web3.eth.Contract`.

packages/loopring_v3.js/src/protocol_v3.ts
Brechtpd: Sure. But do we need those events in the database? The events make sense to keep up to date with the smart contract, but otherwise you can just get the latest data from the smart contract. Unless we want them for reorgs in some way?

packages/loopring_v3.js/src/request_processors/deposit_processor.ts
Brechtpd: I don't push them in `state.processedRequests` here because it's not needed when the block is replayed to restore the state in the case of reverts. So this is done in a level above in `ExchangeV3.processBlock` which knows if this is the initial processing of the block or a replay.

packages/loopring_v3.js/src/request_processors/deposit_processor.ts
Brechtpd: > I also think this class shall emit events so this class's caller can persist the events and maybe states into a DB o choice.

The code can use an interface + implementation provided by the app for storing/retrieving data. Just an idea instead of having to return the event data all over the place, because the library ideally also has easy access to all state. The implementation can be left to the app.

> In general, these processors CAN maintain in-memory state after loading a state-snapshot from DB upon start/restart, then they should omit new events and new state-values for callers to persist.

Keeping all the accounts in memory + balances + trading history for the current state will take A LOT of memory when there are a lot of accounts. Which could be fine for the server that will use this library for the website. Ideally this lib + database could also be used by normal users so they don't have to depend on a web server to create the Merkle proofs for the withdrawal in withdrawal mode. These users will have a normal PC with a couple of GBs of RAM.

> The database may have two types of tables, one for events:
event_type, block_idx, request_idx, ... other data
and for each event type a state table
state_type, block_idx ... other accumulative data
Event Deposit/Withdrawl's corresponding state is Balance
Event Trade/Cancel's corresponding state is Order

Not really sure what's stored in these tables. Is this comparable to what I do now with the `processedRequests` but specifically for every state value that was updated? The on-chain Deposit/Withdrawal events don't immediately update the state and never have to be reverted, only when they are processed does the state they modified need be reverted, but the event is still valid, it just needs to be processed in a new block. But these events in the table would be 'events' that actually modify the state, not the actual on-chain events?

> When we revert block N+1, we delete all events data with block_idx >= N+1 and all states wich block_idx >= N+1,
then we can reload the state with block_idx === N, ....

How can we reload the state with block_idx? Do we revert the state updates stored in the state events on the current block until we get to the block we reverted to?

> The state shall be calculated by one single thread in memory based on the last state and events in the current block. But we probably want to query last state from DB on demand instead of loading all of them in the memory -- it may become too large.

Yes, that's how I would do it as well. Just load the account from DB when needed.

> I think we can merge the PR, then let me define the DB access API, then you can use the API to load and write data.

I think that's a good idea. It should be straightforward for me to update the code where necessary.

dong77:
@Brechtpd I approved this PR as the most fundamental logics here work. As to the application-layer refactor, I believe we can do it later once I have the database schema or API defined. Please feel free to merge.

PR #491, opened 2019-09-18
[REMOVED][PROTOCOL-2] Adds functionality & efficiency improvements to brokers

zrubenst:
- Requests for token approvals from brokers are now batched, similar to how transfers are batched for non-brokered orders
- An optional function `onOrderFillReport` has been added to the `IBrokerDelegate` interface which is called after all fills have been performed if an only if the `brokerRequestAllowance` call returns true. This allows for actions to be executed after all rings are mined/settled.

**Note:** There was a branch `dolomite-deploy` that this could be merged into without conflicts & only showing the changes that have been made between that version and this one, but it is no longer available

### DIFF
--- a/packages/loopring_v2/contracts/helper/MiningHelper.sol
+++ b/packages/loopring_v2/contracts/helper/MiningHelper.sol
@@ -14,7 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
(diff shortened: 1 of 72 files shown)


dong77:
@zrubenst can you resolve all conflicts first?

PR #496, opened 2019-09-20
[protocol3] Reduce gas cost by 20.000 gas

MariusVanDerWijden:
Unix timestamps can be no bigger than ~uint40, using uint64 for the timestamps reduces gas cost for every store of the Staking struct by 20.000 gas.


Brechtpd:
I think even `uint32` should be enough for the timestamps, though in this case it doesn't matter. :)

Looks good to me. Thanks!

Brechtpd:
The only thing we have to watch out for is that we don't have any overflows when doing calculations on the `uint64` values. I haven't checked if that could be the case here.

PR #497, opened 2019-09-20
Save gas cost in universal registry

MariusVanDerWijden:
Address are 20 bytes, we can save the two bools within the protocol address and we can save 20.000 gas if the version string is less than 12 bytes long. I would also consider making the version a number (uint96)

### DIFF
--- a/packages/loopring_v3/contracts/impl/UniversalRegistry.sol
+++ b/packages/loopring_v3/contracts/impl/UniversalRegistry.sol
@@ -34,10 +34,10 @@ contract UniversalRegistry is IUniversalRegistry {
     struct Protocol
     {
         address protocol;
-        address manager;
-        string  version;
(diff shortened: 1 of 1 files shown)


dong77:
@Brechtpd this is an interesting one :) I wonder if we should aim for such extreme optimization...

Brechtpd:
This is fine for me! Solidity handles the complexity of the storage packing/unpacking, we only have to order the member variables correctly in the struct declaration so why not. :)

Because the cost of these struct isn't really important (written once, in very select cases maybe a very slight reduction in reading cost) I think we shouldn't limit the functionality to these costs. So storing the version as a general string seems to be preferable. Just changing the order the members are declared doesn't change anything, so easy gas savings (though in the big picture insignificant because this struct is almost never used when running an exchange)!

PR #508, opened 2019-09-24
[protocol3] Type convert uint to uint64 to reduce gas cost

adamchen0510:
Recommit https://github.com/Loopring/protocols/pull/496
to fix build issue

### DIFF
--- a/packages/loopring_v3/contracts/impl/UserStakingPool.sol
+++ b/packages/loopring_v3/contracts/impl/UserStakingPool.sol
@@ -169,13 +169,13 @@ contract UserStakingPool is Claimable, ReentrancyGuard, IUserStakingPool
 
             total.balance = total.balance.add(claimedAmount);
             total.claimedReward = total.claimedReward.add(claimedAmount);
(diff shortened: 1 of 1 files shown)


adamchen0510:
stake:  198330 -> 169020
withdraw: 42658 -> 42578

packages/loopring_v3/contracts/impl/UserStakingPool.sol
```diff
@@ -169,13 +169,13 @@ contract UserStakingPool is Claimable, ReentrancyGuard, IUserStakingPool
 
             total.balance = total.balance.add(claimedAmount);
             total.claimedReward = total.claimedReward.add(claimedAmount);
-            total.claimedAt = (totalPoints >= userPoints) ?
-            now.sub(totalPoints.sub(userPoints) / total.balance) : now;
+            total.claimedAt = (uint64)((totalPoints >= userPoints) ?
```
dong77: can we use `uint64(...)` instead of `(uint64)(...)`

adamchen0510: Done

PR #509, opened 2019-09-25
[Lightcone_v2.js]modify approve && add approve method unit test

iwbrhwfy:


packages/lightcone_v2.js/src/lib/wallet/config/index.ts
```diff
@@ -169,6 +169,10 @@ function getWallets() {
   return data.wallets;
 }
 
+function getMaxAmountInWEI() {
```
xiaowheat: We should add tests for getMaxAmountInWEI()

PR #512, opened 2019-09-26
[Protocol3] Refactor ProtocolFeeVault and UserStakingPool

dong77:
Some tests are failing and I'm working on fixing them.


packages/loopring_v3/contracts/impl/ProtocolFeeVault.sol
```diff
@@ -100,169 +75,85 @@ contract ProtocolFeeVault is Claimable, ReentrancyGuard, IProtocolFeeVault
         external
         nonReentrant
     {
-        require(msg.sender == userStakingPoolAddress, "UNAUTHORIZED");
-
         require(
-            lrcAddress.safeTransfer(msg.sender, amount),
-            "TRANSFER_FAILURE"
+            userStakingPoolAddress != address(0) &&
+            msg.sender == userStakingPoolAddress,
```
Brechtpd: `msg.sender` can't be `0x0` so the extra `userStakingPoolAddress != address(0)` check seems unnecessary.

dong77: fixed

dong77:
@Brechtpd @adamchen0510 This PR is ready to review and merge. 
I disabled unit tests for UserStakingPool as I will completely re-write it.

packages/loopring_v3/contracts/impl/DowntimeCostCalculator.sol
```diff
@@ -81,8 +81,10 @@ contract DowntimeCostCalculator is Claimable, IDowntimeCostCalculator
         onlyOwner
     {
         require(
-            _basePricePerMinute > 0 && _maxPenalty > 0 &&
-            _gracePeriodPricePerMinute > 0 && _maxAwailableDowntimeMinutes > 0,
+            _basePricePerMinute > 0 &&
+            _maxPenalty > 0 &&
+            _gracePeriodPricePerMinute > 0 &&
+            _maxAwailableDowntimeMinutes > 0,
```
adamchen0510: _maxAwailableDowntimeMinutes -> _maxAvailableDowntimeMinutes

packages/loopring_v3/contracts/iface/IProtocolFeeVault.sol
```diff
@@ -33,38 +33,59 @@ contract IProtocolFeeVault
...
+        uint    amount
     );
+    event SettingsUpdated(
+        address _userStakingPoolAddress,
+        address _tokenSellerAddress,
+        address _daoAddress
+    );
+
+    /// @dev Sets depdending contract address. All these addresses can be zero.
+    /// @param _userStakingPoolAddress The address of the user stking pool.
```
adamchen0510: stking -> staking

packages/loopring_v3/contracts/impl/ProtocolFeeVault.sol
```diff
@@ -25,73 +25,48 @@ import "../lib/MathUint.sol";
...
-        require(_userStakingPoolAddress != address(0), "ZERO_ADDRESS");
-
-        allowOwnerWithdrawal = true;
         lrcAddress = _lrcAddress;
+    }
+
+    function updateSettings(
+        address _userStakingPoolAddress,
+        address _tokenSellerAddress,
+        address _daoAddress
```
adamchen0510: need separate to 3 single update when only update one which can save gas?

dong77: this method is rarely used, so saving gas  is not a concern.

packages/loopring_v3/package.json
```diff
@@ -77,6 +77,8 @@
     "pretty-quick": "^1.11.1",
     "solc": "^0.5.11",
     "solidity-coverage": "^0.5.11",
+    "solium": "^1.2.5",
```
Brechtpd: `ethlint` is the correct package for solium:
- https://github.com/duaraghav8/Ethlint#install

packages/loopring_v3/package.json
```diff
@@ -77,6 +77,8 @@
...
+    "solium-plugin-security": "^0.1.1",
```
Brechtpd: I don't think this is necessary any more:
- https://github.com/duaraghav8/solium-plugin-security#installation

packages/loopring_v3/test/testUserStakingPool.ts
```diff
@@ -1,266 +1,307 @@
...
 
   before(async () => {
-    exchangeTestUtil = new ExchangeTestUtil();
-    await exchangeTestUtil.initialize(accounts);
-    userstaking = exchangeTestUtil.userStakingPool;
-    protocolfee = exchangeTestUtil.protocolFeeVaultContract;
-    userstaking.setProtocolFeeVault(protocolfee.address);
+    mockLRC = await MockContract.new();
+    mockProtocolFeeVault = await MockContract.new();
+    userStakingPool = await UserStakingPool.new(mockLRC.address);
```
Brechtpd: I guess the downside of using a mock address here is that token transfers aren't being tested. You could comment out all token transfers in UserStakingPool and the tests would still pass. So if you make a mistake when calling the token transfer functions these tests will not catch that.

Arguably this could make them more _integration_ tests than _unit_ tests, but I feel like it needs to be tested somewhere, so it may as well be in these tests. I think it makes sense that if you for example test the 'stake' function that you also test that LRC if effectively transferred correctly and staked.

packages/loopring_v3/test/testUserStakingPool.ts
```diff
@@ -1,266 +1,307 @@
+import { Artifacts } from "../util/Artifacts";
+import { advanceTimeAndBlockAsync } from "../util/TimeTravel";
 import { expectThrow } from "./expectThrow";
 import BN = require("bn.js");
-import { ExchangeTestUtil } from "./testExchangeUtil";
+const truffleAssert = require("truffle-assertions");
```
Brechtpd: Nice library!

packages/loopring_v3/test/testUserStakingPool.ts
```diff
@@ -1,266 +1,307 @@
...
+        3: claimableReward
+      } = await userStakingPool.getUserStaking(bob);
+
+      assert(withdrawalWaitTime.eq(new BN(1)), "withdrawalWaitTime");
+      assert(rewardWaitTime.eq(new BN(1)), "rewardWaitTime");
+      assert(balance.eq(amount), "balance");
+      assert(claimableReward.eq(ZERO), "claimableReward");
+
+      // - Check: withdrawal will fail
+      await truffleAssert.fails(
```
Brechtpd: This can check against the actual revert message using `expectThrow` or `truffleAssert.reverts`. This way it can be sure the function reverts for the correct reason.

packages/loopring_v3/test/testUserStakingPool.ts
```diff
@@ -1,266 +1,307 @@
...
-      it("User A withdraw all", async () => {
-        await userstaking.withdraw(0, { from: owner1 });
-      });
-      it("User B withdraw all", async () => {
-        await userstaking.withdraw(0, { from: owner2 });
-      });
-      it("User C withdraw all", async () => {
-        await userstaking.withdraw(0, { from: owner3 });
+  describe("when no protocol fee vault is set, user Charles", () => {
+    const amount = new BN("1000" + "000000000000000000", 10);
```
Brechtpd: Not a fan of `new BN(web3.utils.toWei("1000", "ether"))`?

dong77:
All tests pass.

packages/loopring_v3/test/testUserStakingPool.ts
dong77: Nice to know this. I'll refactor it in my next PR which will introduce more test cases.

packages/loopring_v3/test/testUserStakingPool.ts
dong77: Will also do it in my next PR which focus on this test file.

packages/loopring_v3/test/testUserStakingPool.ts
dong77: The current tests haven't actually mocked the non-default behavior of the depending contracts yet, such as LRC transfer failure. These cases need to be tested here in this file. I'll create another PR for more tests.

PR #533, opened 2019-09-27
[Lightcone_v2.js] Fix eddsa verification function.

yueawang:
1. Fix eddsa verify function and add a simple unit test.
2. Align data with python eddsa implemention used for market maker.

The verify did not work before, so market maker's orders which were badly signed ruin the on-chain status and lead to roll back. This PR tries to enable verify on js and make data align with python eddsa which is used by Hummingbot.


Brechtpd:
Looks good. Keep in mind that beta-3 switches from Pedersen to Poseidon for hashing the requests and the message.

Issue #318, opened 2019-07-29
[Protocol3] Generic Extension for Request Validation (GERV)


Brechtpd:
This is interesting:
- https://medium.com/starkware/starkexchange-fast-withdrawals-using-cookie-jars-88eefea6a11a

This allows withdrawals before a block is proven.

I think we can do something similar like this with minimal changes:
- The user does a withdrawal (on-chain/off-chain) like usual
- The withdrawal is included in a block and the approved withdrawal data is stored on-chain (but the block is not yet proven)
- The user wants to do a fast withdrawal so requests the withdrawn tokens through the cookie jar contract, the contract checks the approved withdrawal data for the user (accountID, tokenID, amountWithdrawn), if the cookie jar has enough funds to do this transfer:
  - The cookie jar immediately sends the tokens to the user using the funds stored in the cookie jar contract.
  - The accountID in the approved withdrawal data is changed to the account of the cookie jar contract so once the block is proven the cookie jar contract receives the tokens it 'borrowed to the user

The cookie jar contract is operated by the exchange so of course the exchange trusts itself to submit a valid proof for the block. If not, the exchange loses the funds (and the user is happy because he can withdraw the funds again! User funds remain safe.).

The user probably needs to approve this somehow because this could be a security risk for users if the cookie jar contract is malicious. The approved withdrawal data could maybe include a single bit if the user allows these fast withdrawals. Or the cookie jar contract cannot be changed easily for an exchange/needs to be approved by the Loopring contract/...
EDIT: the token transfer from the cookie jar contract to the user can be done in the exchange contract which makes it secure for users even without having to trust the cookie jar contract, no approval from users needed. The only change for users is that they get their tokens earlier than they would otherwise, which is only a benefit.

Could be useful, though the number of withdrawals we can do in a single block is very limited by the approved withdrawal data because we need to store that onchain (which is expensive), currently max ~1800 withdrawals/block so the time to prove a block like this will be quite low to begin with.

dong77:
This is a great idea. Let me write down my understanding so @letsgoustc  and @kongliangzhong  can understand it well.

[...]

PR #540, opened 2019-09-29
[Protocol3] WIP:  Generic Extension for Request Validation (GERV)  to enable features such as Fast withdrawals

Brechtpd:
Continuation of #318 with quite some changes.

Untested with many mistakes no doubt. Only here to show the high-level design.

The current design:
- Uses the internal token transfer circuit with no significant changes and practically no additional constraint cost. There is an additional cost gas cost for every conditional transfer included in a block, but no additional overhead for normal internal transfers.
- Complete freedom for when/in what order the conditional transfers are included in a block. There is no queue.
- Allows tokens tokens to be transferred to an account, while a fee can be paid using on-chain balances (could be any Ethereum address) or off-chain balances (always the operator).
- All data to include a conditional transfer in a block is on-chain, it does not depend on any off-chain signed data. It does not depend in any way on the nonce of the account stored in the Merkle tree.

To keep the circuits costs the same we use the data already available on-chain for data availability, with 1 additional byte/transfer that stores the transfer `type`. We have 2 types of transfers:
- Type 0: The transfer has a valid EdDSA signature. If this is the case this is a normal internal transfer.
- Type 1:  The transfer has no valid EdDSA signature. This is a conditional transfer and a block with a transfer like this can only be included when the transfer was approved on-chain. This is checked using the data availability data of the transfer (i.e. the result of the transfer, which is all that matters) and the salt value (only necessary so that multiple conditional transfers can be approved on-chain with the same data-availability data. For example a user may transfer 1ETH 10 times to the same address. Without the salt all these conditional transfers would be exactly the same which the smart contracts disallow (to prevent replay attacks). The salt is used to make the transfer unique, a timestamp can be used for example).

The internal transfer circuit only needs some very small changes:
- Add a `type` byte in the data-availability of the transfer.
- Instead of requiring the signature to be valid, the signature can be invalid. In that case the `type` byte is set to 1, if the signature is valid `type` it is set to 0.
- If `type == 1` we don't increment the nonce (this change is not necessary for the design, but makes the most sense)

That's it. For the no data-availability version we still need to send the data-availabiltiy data on-chain, but the data of the normal transfers can be zeroed out so save gas (and all conditional transfers could be done at the start of the block, so the data will end with only zeros which can easily be compressed away).

A mechanism with an EdDSA request/signature for fast withdrawals is also possible, but this method is more limited (nonces, no immediate authorization on-chain so we are limited in what we can do, no on-chain fee payments for example, ...) and is more expensive in the circuits (additional data for the condition hashes would need to be hashed with sha256, and because hashes are big this would be quite expensive). It could save some gas on-chain, but I think it's not worth it because of the reduced flexibility. Supporting both methods would increase the complexity and increase costs for both.

I think this design has a nice balance between complexity, flexibility, and cost. But I change my mind almost everyday on how to do this so any feedback welcome!

### DIFF
--- a/packages/loopring_v3/contracts/impl/ExchangeV3.sol
+++ b/packages/loopring_v3/contracts/impl/ExchangeV3.sol
@@ -29,6 +29,9 @@ import "./libexchange/ExchangeWithdrawals.sol";
 
 import "../iface/IExchangeV3.sol";
 
+import "../lib/ERC20SafeTransfer.sol";
+import "../lib/MathUint.sol";
+
 
 /// @title An Implementation of IExchangeV3.
 /// @dev This contract supports upgradability proxy, therefore its constructor
@@ -49,12 +52,20 @@ contract ExchangeV3 is IExchangeV3
     using ExchangeMode          for ExchangeData.State;
     using ExchangeTokens        for ExchangeData.State;
     using ExchangeWithdrawals   for ExchangeData.State;
+    using ERC20SafeTransfer     for address;
+    using MathUint              for uint;
 
     ExchangeData.State private state;
 
     modifier onlyOperator()
     {
         require(msg.sender == state.operator, "UNAUTHORIZED");
+        _;
+    }
+
+    modifier onlyRelayer()
+    {
+        require(msg.sender == state.relayer, "UNAUTHORIZED");
         _;
     }
 
@@ -154,6 +165,32 @@ contract ExchangeV3 is IExchangeV3
         )
     {
         return updateAccountAndDepositInternal(
+            msg.sender,
+            pubKeyX,
+            pubKeyY,
+            address(0),
+            0,
+            permission
+        );
+    }
+
+    function createOrUpdateAccount(
+        address owner,
+        uint    pubKeyX,
+        uint    pubKeyY,
+        bytes   calldata permission
+        )
+        external
+        payable
+        nonReentrant
+        returns (
+            uint24 accountID,
+            bool   isAccountNew,
+            bool   isAccountUpdated
+        )
+    {
+        return updateAccountAndDepositInternal(
+            owner,
             pubKeyX,
             pubKeyY,
             address(0),
@@ -363,6 +400,7 @@ contract ExchangeV3 is IExchangeV3
         uint16 blockSize,
         uint8  blockVersion,
         bytes  calldata /*data*/,
+        bytes  calldata auxiliaryData,
         bytes  calldata offchainData
         )
         external
@@ -419,7 +457,7 @@ contract ExchangeV3 is IExchangeV3
               revert(0, 0)
           }
         }
-        state.commitBlock(blockType, blockSize, blockVersion, decompressed, offchainData);
+        state.commitBlock(blockType, blockSize, blockVersion, decompressed, auxiliaryData, offchainData);
     }
 
     function verifyBlocks(
@@ -491,6 +529,7 @@ contract ExchangeV3 is IExchangeV3
         )
     {
         return updateAccountAndDepositInternal(
+            msg.sender,
             pubKeyX,
             pubKeyY,
             token,
@@ -507,7 +546,7 @@ contract ExchangeV3 is IExchangeV3
         payable
         nonReentrant
     {
-        state.depositTo(msg.sender, token, amount, 0);
+        state.deposit(msg.sender, msg.sender, token, amount, 0);
     }
 
     function depositTo(
@@ -519,7 +558,85 @@ contract ExchangeV3 is IExchangeV3
         payable
         nonReentrant
     {
-        state.depositTo(recipient, tokenAddress, amount, 0);
+        state.deposit(msg.sender, recipient, tokenAddress, amount, 0);
+    }
+
+    function deposit(
+        address from,
+        address to,
+        address tokenAddress,
+        uint96  amount
+        )
+        external
+        payable
+        nonReentrant
+    {
+        state.deposit(from, to, tokenAddress, amount, 0);
+    }
+
+    function approveConditionalTransfer(
+        address from,
+        address to,
+        address token,
+        uint24  fAmount,
+        address feeToken,
+        uint24  fFee,
+        uint32  salt
+        )
+        external
+        nonReentrant
+        onlyRelayer
+    {
+        uint24 fromAccountID = state.getAccountID(from);
+        uint24 toAccountID = state.getAccountID(from);
+        uint16 tokenID = state.getTokenID(token);
+        uint16 feeTokenID = state.getTokenID(feeToken);
+
+        // Same packing as onchain DA + salt
+        uint key = fromAccountID;
+        key <<= 20;
+        key |= toAccountID;
+        key <<= 8;
+        key |= tokenID;
+        key <<= 24;
+        key |= fAmount;
+        key <<= 8;
+        key |= feeTokenID;
+        key <<= 16;
+        key |= fFee;
+        key <<= 32;
+        key |= salt;
+
+        // Make sure we didn't already approve this conditional transfer
+        require(!state.conditionalTransfers[key].approved, "TRANSFER_ALREADY_APPROVED");
+        state.conditionalTransfers[key] = ExchangeData.ConditionalTransferState(
+            true,
+            0,
+            0
+        );
+
+        emit ConditionalTransferApproved(
+            from,
+            to,
+            token,
+            uint96(uint(fAmount).decodeFloat(24)),
+            feeToken,
+            uint96(uint(fFee).decodeFloat(16)),
+            salt
+        );
+    }
+
+    function onchainTransferFrom(
+        address from,
+        address to,
+        address token,
+        uint    amount
+        )
+        external
+        nonReentrant
+        onlyRelayer
+    {
+        require(token.safeTransferFrom(from, to, amount), "TRANSFER_FAILED");
     }
 
     // -- Withdrawals --
@@ -836,6 +953,7 @@ contract ExchangeV3 is IExchangeV3
 
     // == Internal Functions ==
     function updateAccountAndDepositInternal(
+        address owner,
         uint    pubKeyX,
         uint    pubKeyY,
         address token,
@@ -860,6 +978,6 @@ contract ExchangeV3 is IExchangeV3
         } else if (isAccountUpdated) {
             additionalFeeETH = state.accountUpdateFeeETH;
         }
-        state.depositTo(msg.sender, token, amount, additionalFeeETH);
+        state.deposit(owner, owner, token, amount, additionalFeeETH);
     }
 }
--- a/packages/loopring_v3/contracts/impl/ExchangeV3.sol
+++ b/packages/loopring_v3/contracts/impl/ExchangeV3.sol
@@ -604,6 +604,8 @@ contract ExchangeV3 is IExchangeV3
         key |= feeTokenID;
         key <<= 16;
         key |= fFee;
+        key <<= 8;
+        key |= 1;
         key <<= 32;
         key |= salt;
 
--- a/packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
+++ b/packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
@@ -445,9 +445,9 @@ library ExchangeBlocks
         // Run over all conditional transfers
         uint numBytesPerTransfer = 13;
         for (uint i = 0; i < auxiliaryData.length; i += 6) {
-            // auxiliaryData contains:
+            // auxiliaryData contains for each conditional transfer:
             // - Transfer index: 2 bytes (the position of the conditional transfer inside data)
-            // - Salt: The salt of the transfer
+            // - Salt: 4 bytes (the salt of the transfer)
             uint index;
             uint salt;
             assembly {
@@ -459,6 +459,9 @@ library ExchangeBlocks
             assembly {
                 transferData := and(mload(add(data, offset)), 0xFFFFFFFFFFFFFFFFFFFFFFFFFF)
             }
+
+            // Check that this is a conditional transfer
+            require((transferData & 0xFF) == 1, "INVALID_AUXILIARYDATA");
 
             // The key of the transfer is the combination of the transfer DA data and the salt
             uint key = (transferData << 32) | salt;
--- a/packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
+++ b/packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
@@ -65,6 +65,7 @@ library ExchangeBlocks
         uint16 blockSize,
         uint8  blockVersion,
         bytes  calldata data,
+        bytes  calldata auxiliaryData,
         bytes  calldata /*offchainData*/
         )
         external
@@ -74,7 +75,8 @@ library ExchangeBlocks
             ExchangeData.BlockType(blockType),
             blockSize,
             blockVersion,
-            data
+            data,
+            auxiliaryData
         );
     }
 
@@ -208,10 +210,11 @@ library ExchangeBlocks
         ExchangeData.BlockType blockType,
         uint16 blockSize,
         uint8  blockVersion,
-        bytes  memory data  // This field already has all the dummy (0-valued) requests padded,
-                            // therefore the size of this field totally depends on
-                            // `blockSize` instead of the actual user requests processed
-                            // in this block. This is fine because 0-bytes consume fewer gas.
+        bytes  memory data,  // This field already has all the dummy (0-valued) requests padded,
+                             // therefore the size of this field totally depends on
+                             // `blockSize` instead of the actual user requests processed
+                             // in this block. This is fine because 0-bytes consume fewer gas.
+        bytes  memory auxiliaryData
         )
         private
     {
@@ -236,11 +239,13 @@ library ExchangeBlocks
         );
 
         // Extract the exchange ID from the data
-        uint32 exchangeIdInData = 0;
-        assembly {
-            exchangeIdInData := and(mload(add(data, 4)), 0xFFFFFFFF)
+        {
+            uint32 exchangeIdInData = 0;
+            assembly {
+                exchangeIdInData := and(mload(add(data, 4)), 0xFFFFFFFF)
+            }
+            require(exchangeIdInData == S.id, "INVALID_EXCHANGE_ID");
         }
-        require(exchangeIdInData == S.id, "INVALID_EXCHANGE_ID");
 
         // Get the current block
         ExchangeData.Block storage prevBlock = S.blocks[S.blocks.length - 1];
@@ -374,10 +379,11 @@ library ExchangeBlocks
                 require(inputEndingHash == endingHash, "INVALID_ENDING_HASH");
                 numWithdrawalRequestsCommitted += uint32(count);
             }
+        } else if (blockType == ExchangeData.BlockType.TRANSFER) {
+            validateConditionalTransfers(S, uint32(S.blocks.length), data, auxiliaryData);
         } else if (
             blockType != ExchangeData.BlockType.OFFCHAIN_WITHDRAWAL &&
-            blockType != ExchangeData.BlockType.ORDER_CANCELLATION &&
-            blockType != ExchangeData.BlockType.TRANSFER) {
+            blockType != ExchangeData.BlockType.ORDER_CANCELLATION) {
             revert("UNSUPPORTED_BLOCK_TYPE");
         }
 
@@ -408,6 +414,7 @@ library ExchangeBlocks
             blockSize,
             blockVersion,
             uint32(now),
+            uint32(block.number),
             numDepositRequestsCommitted,
             numWithdrawalRequestsCommitted,
             false,
@@ -418,6 +425,61 @@ library ExchangeBlocks
         S.blocks.push(newBlock);
 
         emit BlockCommitted(S.blocks.length - 1, publicDataHash);
+    }
+
+    function validateConditionalTransfers(
+        ExchangeData.State storage S,
+        uint32 newBlockIdx,
+        bytes  memory data,
+        bytes  memory auxiliaryData
+        )
+        private
+    {
+        // Read how many conditional transfers are included in the block
+        uint numConditionalTransfers;
+        assembly {
+            numConditionalTransfers := and(mload(add(data, 104)), 0xFFFFFFFF)
+        }
+        require(auxiliaryData.length == numConditionalTransfers * 6, "INVALID_AUXILIARYDATA");
+
+        // Run over all conditional transfers
+        uint numBytesPerTransfer = 13;
+        for (uint i = 0; i < auxiliaryData.length; i += 6) {
+            // auxiliaryData contains:
+            // - Transfer index: 2 bytes (the position of the conditional transfer inside data)
+            // - Salt: The salt of the transfer
+            uint index;
+            uint salt;
+            assembly {
+                index := and(mload(add(auxiliaryData, add(i, 2))), 0xFFFF)
+                salt := and(mload(add(auxiliaryData, add(i, 6))), 0xFFFFFFFF)
+            }
+            uint offset = 104 + index * numBytesPerTransfer;
+            uint transferData;
+            assembly {
+                transferData := and(mload(add(data, offset)), 0xFFFFFFFFFFFFFFFFFFFFFFFFFF)
+            }
+
+            // The key of the transfer is the combination of the transfer DA data and the salt
+            uint key = (transferData << 32) | salt;
+            ExchangeData.ConditionalTransferState memory transferState = S.conditionalTransfers[key];
+            require(transferState.approved, "TRANSFER_UNAUTHORIZED");
+
+            // Make sure the request was either not used yet, or used in a block that was reverted
+            if (transferState.blockIdx != 0 && transferState.blockIdx < S.blocks.length) {
+                // A block exists, but the block could have been changed.
+                // Check if the block has a newer Ethereum block number
+                require(
+                    S.blocks[transferState.blockIdx].ethereumBlockNumber > transferState.ethereumBlockNumber,
+                    "TRANSFER_ALREADY_CONSUMED"
+                );
+            }
+
+            // Update the state
+            transferState.blockIdx = newBlockIdx;
+            transferState.ethereumBlockNumber = uint32(block.number);
+            S.conditionalTransfers[key] = transferState;
+        }
     }
 
     function validateAndUpdateProtocolFeeValues(
--- a/packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
+++ b/packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
@@ -66,23 +66,26 @@ library ExchangeDeposits
         timestamp = request.timestamp;
     }
 
-    function depositTo(
+    function deposit(
         ExchangeData.State storage S,
-        address recipient,
+        address from,
+        address to,
         address tokenAddress,
         uint96  amount,  // can be zero
         uint    additionalFeeETH
         )
         external
     {
-        require(recipient != address(0), "ZERO_ADDRESS");
+        require(from != address(0), "ZERO_ADDRESS");
+        require(to != address(0), "ZERO_ADDRESS");
+        require(S.isAuthorized(from), "UNAUTHORIZED");
         require(S.areUserRequestsEnabled(), "USER_REQUEST_SUSPENDED");
         require(getNumAvailableDepositSlots(S) > 0, "TOO_MANY_REQUESTS_OPEN");
 
         uint16 tokenID = S.getTokenID(tokenAddress);
         require(!S.tokens[tokenID].depositDisabled, "TOKEN_DEPOSIT_DISABLED");
 
-        uint24 accountID = S.getAccountID(recipient);
+        uint24 accountID = S.getAccountID(to);
         ExchangeData.Account storage account = S.accounts[accountID];
 
         // Total fee to be paid by the user
@@ -90,7 +93,7 @@ library ExchangeDeposits
 
         // Transfer the tokens to this contract
         transferDeposit(
-            account.owner,
+            from,
             tokenAddress,
             amount,
             feeETH
@@ -134,7 +137,11 @@ library ExchangeDeposits
         );
         S.deposits.push(_deposit);
 
+        // Keep track how many tokens are deposited in the exchange
         S.tokenBalances[tokenAddress] = S.tokenBalances[tokenAddress].add(amount);
+        // Limit the total token balance to MAX_TOTAL_TOKEN_BALANCE.
+        // This way we can never have overflows when depositing, so users can never lose any funds this way.
+        require(S.tokenBalances[tokenAddress] <= ExchangeData.MAX_TOTAL_TOKEN_BALANCE(), "TOKEN_BALANCE_TOO_HIGH");
 
         emit DepositRequested(
             uint32(S.depositChain.length - 1),
--- a/packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
+++ b/packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
@@ -25,6 +25,8 @@ import "../../lib/MathUint.sol";
 import "../../thirdparty/ECDSA.sol";
 
 /// @title An Implementation of IRelayer.
+/// @dev FastWithdrawal needs to extend RelayerData to make sure the data layout matches
+///      with that layout in the Relayer contract.
 /// @author Brecht Devos - <brecht@loopring.org>
 contract FastWithdrawal is RelayerData
 {
--- a/packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
+++ b/packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
@@ -1,0 +1,92 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+import "./RelayerData.sol";
+
+import "../../iface/IExchangeV3.sol";
+import "../../iface/ICookieJarContract.sol";
+
+import "../../lib/MathUint.sol";
+import "../../thirdparty/ECDSA.sol";
+
+/// @title An Implementation of IRelayer.
+/// @author Brecht Devos - <brecht@loopring.org>
+contract FastWithdrawal is RelayerData
+{
+    using ECDSA             for bytes32;
+    using MathUint          for uint;
+
+    function executeFastWithdrawal(
+        address cookieJarContractAddress,
+        address exchangeAddress,
+        address destinationAddress,
+        address from,
+        address to,
+        address token,
+        uint24  fAmount,
+        address feeToken,
+        uint16  fFee,
+        bool    onchainFeePayment,
+        uint32  salt,
+        bytes   calldata signature
+        )
+        external
+    {
+        // Check the signature
+        bytes32 hash = keccak256(
+            abi.encodePacked(
+                cookieJarContractAddress,
+                exchangeAddress,
+                destinationAddress,
+                from,
+                to,
+                token,
+                uint(fAmount).decodeFloat(24),
+                feeToken,
+                uint(fFee).decodeFloat(16),
+                onchainFeePayment,
+                salt
+            )
+        );
+        require(hash.toEthSignedMessageHash().recover(signature) == from, "UNAUTHORIZED");
+
+        // Transfer the tokens directly to the destination address
+        ICookieJarContract(cookieJarContractAddress).transfer(destinationAddress, token, uint(fAmount).decodeFloat(24));
+
+        // Do fee payment directly from the user's wallet if requested
+        if (onchainFeePayment) {
+            IExchangeV3(exchangeAddress).onchainTransferFrom(
+                from,
+                to,
+                feeToken,
+                uint(fFee).decodeFloat(16)
+            );
+        }
+
+        // Approve the conditional transfer
+        IExchangeV3(exchangeAddress).approveConditionalTransfer(
+            from,
+            to,
+            token,
+            fAmount,
+            feeToken,
+            onchainFeePayment ? 0 : fFee,
+            salt
+        );
+    }
+}
--- a/packages/loopring_v3/contracts/impl/relayer/Relayer.sol
+++ b/packages/loopring_v3/contracts/impl/relayer/Relayer.sol
@@ -1,0 +1,50 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+import "./RelayerData.sol";
+
+import "../../lib/Claimable.sol";
+import "../../lib/ReentrancyGuard.sol";
+
+
+/// @title An Implementation of IRelayer.
+/// @author Brecht Devos - <brecht@loopring.org>
+contract Relayer is Claimable, ReentrancyGuard, RelayerData
+{
+    function executeTransaction(
+        address module,
+        bytes memory transaction
+        )
+        public
+        nonReentrant
+    {
+        require(modules[module], "UNAUTHORIZED_MODULE");
+        (bool success, ) = module.delegatecall(transaction);
+        require(success, "MODULE_DELEGATECALL_FAILED");
+    }
+
+    function setAuthorized(
+        address module,
+        bool authorized
+        )
+        external
+        onlyOwner
+    {
+        modules[module] = authorized;
+    }
+}
--- a/packages/loopring_v3/contracts/iface/ICookieJarContract.sol
+++ b/packages/loopring_v3/contracts/iface/ICookieJarContract.sol
@@ -1,0 +1,34 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+
+/// @title ICookieJarContract
+/// @author Brecht Devos - <brecht@loopring.org>
+contract ICookieJarContract
+{
+    /// @dev Transfers funds out of the contract to a specified address
+    /// @param to The recipient of the funds
+    /// @param token The token address (0x0 for ETH)
+    /// @param amount The number of tokens transferred to the recipient
+    function transfer(
+        address to,
+        address token,
+        uint    amount
+        )
+        external;
+}
--- a/packages/loopring_v3/contracts/iface/IExchangeV3.sol
+++ b/packages/loopring_v3/contracts/iface/IExchangeV3.sol
@@ -136,6 +136,16 @@ contract IExchangeV3 is IExchange
         uint8 previousMakerFeeBips
     );
 
+    event ConditionalTransferApproved(
+        address indexed fromAccountID,
+        address         toAccountID,
+        address         token,
+        uint96          amount,
+        address         feeToken,
+        uint96          fee,
+        uint32          salt
+    );
+
     // -- Initialization --
     /// @dev Initializes this exchange. This method can only be called once.
     /// @param  owner The owner of this exchange.
@@ -468,16 +478,27 @@ contract IExchangeV3 is IExchange
     ///            - Number of withdrawals processed: 4 bytes
     ///            - For every withdrawal:
     ///                - Token ID: 1 bytes
-    ///                - Account ID: 2,5 bytes
-    ///                - Amount: 3,5 bytes
+    ///                - Account ID: 2.5 bytes
+    ///                - Amount: 3.5 bytes
     ///        For OFFCHAIN_WITHDRAWAL blocks add the following data:
     ///            - For every withdrawal:
     ///                - Token ID: 1 bytes
-    ///                - Account ID: 2,5 bytes
-    ///                - Amount: 3,5 bytes
+    ///                - Account ID: 2.5 bytes
+    ///                - Amount: 3.5 bytes
     ///            - Label hash: 32 bytes
     ///        For ORDER_CANCELLATION blocks add the following data:
     ///            - Label hash: 32 bytes
+    ///        For TRANSFER blocks add the following data:
+    ///            - Label hash: 32 bytes
+    ///            - Number of conditional transfers (4 bytes)
+    ///            - For every transfer:
+    ///                - From account ID: 2.5 bytes
+    ///                - To account ID: 2.5 bytes
+    ///                - Token ID: 1 bytes
+    ///                - Amount: 3 bytes
+    ///                - Fee token ID: 1 bytes
+    ///                - Fee amount: 2 bytes
+    ///                - Type: (1 bytes) (isConditional ? 1 : 0)
     ///
     ///        The 'onchain data availability' data (if enabled) is added
     ///        at the end. This allows anyone to recreate the Merkle tree
@@ -486,10 +507,10 @@ contract IExchangeV3 is IExchange
     ///        For RING_SETTLEMENT blocks add the following data:
     ///            - Operator account ID: 3 bytes
     ///            - For every ring
-    ///                - OrderA.orderID: 2,5 bytes
-    ///                - OrderB.orderID: 2,5 bytes
-    ///                - OrderA.accountID: 2,5 bytes
-    ///                - OrderB.accountID: 2,5 bytes
+    ///                - OrderA.orderID: 2.5 bytes
+    ///                - OrderB.orderID: 2.5 bytes
+    ///                - OrderA.accountID: 2.5 bytes
+    ///                - OrderB.accountID: 2.5 bytes
     ///                - For both Orders:
     ///                    - TokenS: 1 bytes
     ///                    - FillS: 3 bytes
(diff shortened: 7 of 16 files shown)


dong77:
@Brechtpd Your approach is interesting, I cannot say I fully digest your ideas yet -- I have to run to a 4-hour training in 30 minutes, so please allow me to speak my mind, then I'll read your PR again.

Your approach makes the Exchange contract more complicated which is not the desired effect. Here is my idea -- maybe it can be merged into yours to achieve something even more interesting.

---

First of all, assuming there is a ConditionManager contract that manages conditions for a particular kind (so we can later have many types of conditions).

```solidity

interface ConditionManager {

    uint public conditionType;
    function validateAndConsumeCondition(
        address operator,
        bytes inputData
        );
}

contract FastWithdrawalConditionManager is ConditionManager {
    uint public conditionType = 1;
    // exchange => sender => receiver => outstanding_amount

    function getConditionType() returns (uint);
    // Any address including the operator of the exchange can call this method to
    //  transfer token/Ether to another address and keep an outstanding balance.
    //  senders will have to set allowance first.
    function transfer(
        address exchange,
        address to,
        address token,
        uint amount
      )
    {
        require(exchanage != address(0));
        token.safeTransferFromAndVerify(msg.sender, to, amount);
        balances[exchange][msg.sender][to] += amount;
    }
    // This method actually checks the conditions and consumes a valid condition
    // and is only callable by the Exchange contract

   function validateAndConsumeCondition(
        address operator,
        bytes inputData
        )
    {
        address from = operator;
        address to = extractUser(inputData);
        address token = extractToken(inputData);
        uint    amount = extractAmount(inputData);
        validateAndConsumeConditionInternal(from, to, token, amount);
    }

    function validateAndConsumeConditionInternal(
        address from,
        address to,
        address token,
        uint amount
        )
        private
    {
        require(balances[msg.sender][from][to] >= amount, "INSUFFCIENT_BALANCE");
        balances[msg.sender][from][to] - = amount;
    }
}

```


In the exchange contract, we can allow the owner to set the CookieJar address

```solidity
contract Exchange {
    ...

    // conditionType => ConditionManager address.
    mapping (uint => address) conditionManagers;
    // Anyone can call this method to transfter token/Ether to another address
    // and keep an outstnding balance
    function setConditionManager(
        address conditionManager
        )
        onlyOwner
    {
        ConditionManager manager = ConditionManager(conditionManager);
        this.conditionManagers[manager.conditionType] = manager;
    }

    function getConditionManager(
      uint conditionType
    ) {
      address manager = conditionManagers[conditionType];
      require(manager != address(0));

      return ConditionManager(manager)
    }

}

```

And when we are processing blocks:

```solidity

   if (blockType == TRANSFER) {

    for (i = 0; i< numOfRequests; i++) {
       // perform other validation the same way
       bytes requestData = ....
       uint conditionType = extractCondition(requestData);

       if (conditionType != 0) {
          getConditionManager(t.conditionType).validateAndConsumeCondition(requestData)
       }
    }

   }

```

dong77:
I think the idea can go beyond conditional transfer. We can allow an exchange owner to register RequestCallbackHandlers as:

```solidity

 setRequestCallbackHandlers(address handlerAddress) onlyOwner {
   RequestCallbackHandler handler = RequestCallbackHandler(handlerAddress)
   require(handlers[handler.blockType][handler.requestType] == address(0));
   handlers[handler.blockType][handler.requestType] = handlerAddress;
  }
```  

Then for each request in commitBlock:

```solidity
 address handlerAddress = handlers[blockType][requestType];
 if (handlerAddress != address(0) {
   RequestCallbackHandler(handlerAddress).handle(requstData); // throw in case of error.
}
```

In theory, this design will make the exchange extensible -- the owner can create their own sub-request types with the auxiliary data you introduced; in reality, I don't know how much can be done with such extensibility though.

Brechtpd:
> @Brechtpd Your approach is interesting, but I feel we make the Exchange contract complicated. I had this idea so we don't have to change the Exchange contract a lot, let me know how you feel.

I don't think the Exchange contract gets more complicated than necessary. The only thing we really need for the fast withdrawals is a way to register the conditional transfers in the exchange and extra handling to check the conditions when included in a block. With any mechanism both are necessary. It's not particularly easy to understand how and why it's done like this, but the code is fairly easy. The other small changes (adding `from` addresses to the public API) are to support things similar to gas stations.

> 
> First of all, assuming there is a ConditionManager contract that manages conditions for a particular kind (so we can later have many types of conditions).

Actually I think a single condition is actually fine from the Exchange's point of view. Either the transfer can be done when allowed onchain or not (and in this case an EdDSA signature is needed). It doesn't matter what the condition was that allowed the transfer. So the `balances` data below can just be moved to the Exchange itself so all conditional transfer work on the same data (like this PR does). The actual reason the conditional transfer was allowed is not important.

> 
> ```solidity
> interface ConditionManager {
> 
>     uint public conditionType;
>     function validateAndConsumeCondition(
>         address operator,
>         bytes inputData
>         );
> }
> 
> contract FastWithdrawalConditionManager is ConditionManager {
>     uint public conditionType = 1;
>     // exchange => sender => receiver => outstanding_amount
> 
>     mapping(address => mapping(address => mapping(address => uint))) balances;
>     // Any address including the operator of the exchange can call this method to
>     //  transfer token/Ether to another address and keep an outstanding balance.
>     //  senders will have to set allowance first.
>     function transfer(
>         address exchange,
>         address to,
>         address token,
>         uint amount
>       )
>     {
>         require(exchanage != address(0));
>         token.safeTransferFromAndVerify(msg.sender, to, amount);
>         balances[exchange][msg.sender][to] += amount;
>     }
>     // This method actually checks the conditions and consumes a valid condition
>     // and is only callable by the Exchange contract
> 
>    function validateAndConsumeCondition(
>         address operator,
>         bytes inputData
>         )
>     {
>         address from = operator;
>         address to = extractUser(inputData);
>         address token = extractToken(inputData);
>         uint    amount = extractAmount(inputData);
>         validateAndConsumeConditionInternal(from, to, token, amount);
>     }
> 
>     function validateAndConsumeConditionInternal(
>         address from,
>         address to,
>         address token,
>         uint amount
>         )
>         private
>     {
>         require(balances[msg.sender][from][to] >= amount, "INSUFFCIENT_BALANCE");
>         balances[msg.sender][from][to] - = amount;
>     }
> }
> ```

It would be a bit simpler if we could `balances` like this, but I can't think of a way to support block reverts with this. If `validateAndConsumeConditionInternal` is called for request in a block, and that block is then reverted we cannot easily revert the state so the request can be done again in a new block. The mechanism in this PR does support reverts, but it is certainly less clean then the method above.

For the `FastWithdrawalConditionManager.transfer` function it is extremely important that only the owner of the cookie jar contract can actually execute the request. The owner of the cookie jar contract takes the risk that the user actually has the amount requested offchain AND that this amount will still be available when the conditional transfer is actually included in a block. So in practice the owner of the cookie jar and the operator will be the same entity so that this can be ensured. So a user can request a fast-withdrawal (with an ECDSA signed request in this PR) but the actual execution of the fast withdrawal (the onchain transfer) cannot be initiated by anyone other than the operator/cookie jar owner. (`to` will also be a different addresses for the onchain and internal transfers, but this can be fixed by always using the `to` address specified in the smart contract for the internal transfer).

> 
> In the exchange contract, we can allow the owner to set the CookieJar address
> 
> ```solidity
> contract Exchange {
>     ...
> 
>     // conditionType => ConditionManager address.
>     mapping (uint => address) conditionManagers;
>     // Anyone can call this method to transfter token/Ether to another address
>     // and keep an outstnding balance
>     function setConditionManager(
>         address conditionManager
>         )
>         onlyOwner
>     {
>         ConditionManager manager = ConditionManager(conditionManager);
>         this.conditionManagers[manager.conditionType] = manager;
>     }
> 
>     function getConditionManager(
>       uint conditionType
>     ) {
>       address manager = conditionManagers[conditionType];
>       require(manager != address(0));
> 
>       return ConditionManager(manager)
>     }
> 
> }
> ```

In my PR things like this are done in the Relayer. The Exchange doesn't need to know of any fast-withdrawals or multiple condition types or anything like that. The only thing the Exchange needs to know if it's conditional or not, which simplifies the Exchange contract.

> 
> And when we are processing blocks:
> 
> ```solidity
>    if (blockType == TRANSFER) {
> 
>     for (i = 0; i< numOfRequests; i++) {
>        // perform other validation the same way
>        bytes requestData = ....
>        uint conditionType = extractCondition(requestData);
> 
>        if (conditionType != 0) {
>           getConditionManager(t.conditionType).validateAndConsumeCondition(requestData)
>        }
>     }
> 
>    }
> ```

This done similarly in this PR, except that it just works on the conditional transfer data as stored in the exchange itself instead of in a separate condition manager, which I think is unnecessary (see above).

> I think the idea can go beyond conditional transfer. We can allow an exchange owner to register RequestCallbackHandlers as:
> 
> ```solidity
>  setRequestCallbackHandlers(address handlerAddress) onlyOwner {
>    RequestCallbackHandler handler = RequestCallbackHandler(handlerAddress)
>    require(handlers[handler.blockType][handler.requestType] == address(0));
>    handlers[handler.blockType][handler.requestType] = handlerAddress;
>   }
> ```
> 
> Then for each request in commitBlock:
> 
> ```solidity
>  address handlerAddress = handlers[blockType][requestType];
>  if (handlerAddress != address(0) {
>    RequestCallbackHandler(handlerAddress).handle(requstData); // throw in case of error.
> }
> ```
> 
> In theory, this design will make the exchange extensible -- the owner can create their own sub-request types with the auxiliary data you introduced; in reality, I don't know how much can be done with such extensibility though.

The Exchange only needs to know if it's conditional or not in my PR, the condition logic is handled in the Relayer using completely general modules (no need for an interface even). The Exchange doesn't need to know of any actual conditions.

I think your suggestions are actually quite similar to what is done in this PR, but some things are either done in the Exchange itself or outside the Exchange where I think it makes the most sense. And there is a bit of extra complexity to support block reverts unfortunately.

One of the things not handled in your code/comments is support for things like gas station networks. We kind of need to have a contract like the Relayer contract that can call the Exchange for a user so the user doesn't have to do the Ethereum transaction itself. Adding just an extra `bytes signature` to all methods where we want to support this produces a worse user experience because the user may need to sign multiple times (e.g. authorize the fee payment to the gas station network + authorize the call in the Exchange, so a user needs to sign some data twice for a single piece of work). So modules in the Relayer can not only be used to add different kinds of uses of conditional transfers, it can also be used to relay transactions for users or implement any kind of logic on top of an Exchange (instead of callbacks from inside the exchange itself).

packages/loopring_v3/contracts/impl/ExchangeV3.sol
```diff
@@ -419,7 +457,7 @@ contract ExchangeV3 is IExchangeV3
               revert(0, 0)
           }
         }
-        state.commitBlock(blockType, blockSize, blockVersion, decompressed, offchainData);
+        state.commitBlock(blockType, blockSize, blockVersion, decompressed, auxiliaryData, offchainData);
```
dong77: break to multiple lines.

packages/loopring_v3/contracts/impl/ExchangeV3.sol
```diff
@@ -519,7 +558,87 @@ contract ExchangeV3 is IExchangeV3
...
+        address from,
+        address to,
+        address token,
+        uint    amount
+        )
+        external
+        nonReentrant
+        onlyRelayer
+    {
+        require(token.safeTransferFrom(from, to, amount), "TRANSFER_FAILED");
```
dong77: so the point of introducing this `onchainTransferFrom` function is to avoid another user approval? Since this method doesn't have any condition checking, it means the caller, or the Relayer contract cannot be signed dynamically and the address must be provided during DEX initialization to reduce the risk (otherwise if the relayer contract can be set later, then the operator can still people's token)

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -420,6 +427,64 @@ library ExchangeBlocks
...
+        bytes  memory auxiliaryData
+        )
+        private
+    {
+        // Read how many conditional transfers are included in the block
+        uint numConditionalTransfers;
+        assembly {
+            numConditionalTransfers := and(mload(add(data, 104)), 0xFFFFFFFF)
+        }
+        require(auxiliaryData.length == numConditionalTransfers * 6, "INVALID_AUXILIARYDATA");
```
dong77: if the 'conditions' in all requests are not part of the OCDA, will the operator be able to change the conditions here to cheat? For example, change the salt for a much "easier" condition.

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -420,6 +427,64 @@ library ExchangeBlocks
...
+            // Check that this is a conditional transfer
+            require((transferData & 0xFF) == 1, "INVALID_AUXILIARYDATA");
+
+            // The key of the transfer is the combination of the transfer DA data and the salt
+            uint key = (transferData << 32) | salt;
+            ExchangeData.ConditionalTransferState memory transferState = S.conditionalTransfers[key];
+            require(transferState.approved, "TRANSFER_UNAUTHORIZED");
+
+            // Make sure the request was either not used yet, or used in a block that was reverted
+            if (transferState.blockIdx != 0 && transferState.blockIdx < S.blocks.length) {
```
dong77: I'm having a hard time to understand the reason we need `ethereumBlockNumber`... 

Assuming in Etheruem block X there is a Condition tx and no virtual blocks are in block X. Then if the Condition tx got reorganized out, such a reorg it will make all the virtual blocks after block X invalid. I think the exchange relayer should handle that scenario without trying to make the previous virtual blocks valid.

In other words, if the condition for a fast withdrawal becomes invalid, the DEX's virtual blocks need to be invalidated. The operator needs to wait long enough after the condition tx to include the offchain requests.

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -66,31 +66,34 @@ library ExchangeDeposits
         timestamp = request.timestamp;
     }
 
-    function depositTo(
+    function deposit(
```
dong77: I believe this will support meta tx.

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -134,7 +137,11 @@ library ExchangeDeposits
         );
         S.deposits.push(_deposit);
 
+        // Keep track how many tokens are deposited in the exchange
         S.tokenBalances[tokenAddress] = S.tokenBalances[tokenAddress].add(amount);
+        // Limit the total token balance to MAX_TOTAL_TOKEN_BALANCE.
+        // This way we can never have overflows when depositing, so users can never lose any funds this way.
+        require(S.tokenBalances[tokenAddress] <= ExchangeData.MAX_TOTAL_TOKEN_BALANCE(), "TOKEN_BALANCE_TOO_HIGH");
```
dong77: nice

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -66,31 +66,34 @@ library ExchangeDeposits
...
+        uint24 accountID = S.getAccountID(to);
         ExchangeData.Account storage account = S.accounts[accountID];
 
         // Total fee to be paid by the user
         uint feeETH = additionalFeeETH.add(S.depositFeeETH);
 
         // Transfer the tokens to this contract
         transferDeposit(
-            account.owner,
+            from,
```
dong77: so we need to change this method such that if msg.sender == from, we use safeTransfer, otherwise, we use safeTransferFrom. Right?

packages/loopring_v3/contracts/impl/relayer/Relayer.sol
```diff
@@ -0,0 +1,50 @@
...
+        (bool success, ) = module.delegatecall(transaction);
+        require(success, "MODULE_DELEGATECALL_FAILED");
+    }
+
+    function setAuthorized(
+        address module,
+        bool authorized
+        )
+        external
+        onlyOwner
```
dong77: will this ower be us or the DEX owner?

dong77: To increase security, we change the `authorized` to `uint sinceBlock` which must be at least 10000 blocks from the current block. Set to 0 means disabling it.

dong77:
@Brechtpd I read your code carefully and provided some feedback/questions. 

I do like the way you introduce the Relayer contract, I hope this contract is managed by us otherwise there will be security issues. I like that it can delegate calls to newly deployed modules for extensibility. I also suggested in my comment that newly added modules should only become valid after enough wait time, so if users feel the new modules are not secure, they can have enough time to withdraw all their funds.

This idea can be described as **A mechanism to transact on behalf of users**. But I don't think the fast withdrawal feature need to use the above mechanism, as fast withdrawals are essentially **An extension to the default exchange logics for validation of requests**, for which my approach of introducing a generic request sub-type may work better.

packages/loopring_v3/contracts/impl/ExchangeV3.sol
Brechtpd: Correct. The Relayer would be something that is also exclusively handled by Loopring and we could perhaps make it updatable using the protocol contract, though because the main Relayer contract doesn't do much (just forwards call to modules) I don't think that's even needed. We could make it so that exchanges could opt in or out of modules, though I also don't think that's necessary because modules will just call exchange functions that it would need to process anyway.

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
Brechtpd: The condition responsible for the approval of the internal transfer doesn't matter. The only thing that matters is that user approved some transfer to some account for some reason. The salt is only there to make the transfer unique so an identical transfer can be approved in multiple conditional requests. 

Example: tet's say the user has 2 conditional transfers.
- The user allows the transfer of 1 ETH to the operator for a fast withdrawal
- The user allows the transfer of 1 ETH as the fee for calling an on-chain function for the user

The condition of both request is different, but the resulting transfer is the same: 1 ETH needs to be transferred to the operator. The salt is used to make it unique so the uniqueness check when approving the transfer doesn't detect it as a duplicate. But the result is the same, so it doesn't matter what the condition was that resulted in the approval. The result is just that the 1 ETH is allowed to be transferred to the operator.

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
Brechtpd: It wasn't added for reorgs, it was added for reverts. For reorgs we don't have to do anything onchain because if a block was reorged out it was never part of the onchain state to begin with, so it works out.

But with an actual revert where the block gets reverted by calling `revertBlock` on the exchange we do have to make sure in the smart contract that all onchain requests that were handled in blocks that were reverted can be included in a new block again. This is also the reason we store `numDepositRequestsCommitted` and `numWithdrawalRequestsCommitted` in the block. If blocks get reverted we automatically also make sure that  onchain requests need to be handled again starting at `numDepositRequestsCommitted` and `numWithdrawalRequestsCommitted` for the latest block, no matter what request were already handled in the reverted blocks. But we don't have queue for the conditional transfers so now we have to store that state for every conditional transfer independently. This is where the `ethereumBlockNumber` comes in. This allows us to detect if the block where it was included was changed because of a block revert, which means we need to allow the inclusion of the conditional transfer in a block again.

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
Brechtpd: Yes

packages/loopring_v3/contracts/impl/libexchange/ExchangeDeposits.sol
Brechtpd: No, the funds are always need to be transferred from the `from` address, never from the contract itself (which safeTransfer would need). This change also fjust for the meta tx, `from` can authorize the deposit directly himself (`from == msg.sender`) or indirectly using the Relayer using an ECDSA signture/gas station network (`relayer == msg.sender`) implemented in a module.

packages/loopring_v3/contracts/impl/relayer/Relayer.sol
Brechtpd: This will be us. The Relayer needs to be secure because it can do exchange operations for users.

That's a good idea to add a delay. We could also allow exchanges to choose which modules they want 'support', though this would only be for security reasons they would want to disable a module. Modules will just use basic exchange functionality which they would have to do process anyway so it shouldn't matter for them.

Brechtpd:
> @Brechtpd I read your code carefully and provided some feedback/questions.
> 
> I do like the way you introduce the Relayer contract, I hope this contract is managed by us otherwise there will be security issues. I like that it can delegate calls to newly deployed modules for extensibility. I also suggested in my comment that newly added modules should only become valid after enough wait time, so if users feel the new modules are not secure, they can have enough time to withdraw all their funds.

I agree with everything here. :)

> This idea can be described as **A mechanism to transact on behalf of users**. But I don't think the fast withdrawal feature need to use the above mechanism, as fast withdrawals are essentially **An extension to the default exchange logics for validation of requests**, for which my approach of introducing a generic request sub-type may work better.

I see conditional transfers more as a building block of the exchange, like deposits, withdrawals an order cancellations. I don't see it as an extension. Fast withdrawals are just one use of this building block and I would not really call it a 'type' of conditional transfer, just a use of it (checking if it is approved would essentially always be the same, the circuits don't have any knowledge of 'types'). 
Another use would be to allow users to use their exchange account as a wallet from where they can also execute any Ethereum transactions. Another one could perhaps be leveraged trading (I have not idea if this would actually work, but users could lend extra tokens and sign a request that allows the exchange to liquidate those tokens when the price drops below a certain point). Yet another one could be to allow the user to buy tokens from something like uniswap behind the scenes (use the cookie jar contract to sell the tokens immediately and deposit the tokens bought this way directly to the user's account). So I think there can be many modules that can make use of it in many ways.

I do agree that it boils down to a mechanism to transact on behalf of users. And I think this is exactly the category where conditional transfers fall into.The user allows an internal transfer transaction **if** the user allows the transaction for some reason. The user approves of it in some way. The same as with gas stations, signatures or any other way of authorizing a transaction. This is what the relayer modules do. They check if the conditions of the user is fulfilled and if so the transaction on the exchange contract is allowed. So I think the Relayer is a good place to implement it, I don't really see a reason to have 2 different systems that would kind of do the same  (and Relayer modules have some extra flexibility because they are outside of the Exchange).

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -420,6 +427,64 @@ library ExchangeBlocks
...
+                // A block exists, but the block could have been changed.
+                // Check if the block has a newer Ethereum block number
+                require(
+                    S.blocks[transferState.blockIdx].ethereumBlockNumber > transferState.ethereumBlockNumber,
```
dong77: Now I understand the purpose of keeping the record of Ethereum block numbers in a virtual block.
Just to straight my mind:
1.  if `transferState.blockIdx == 0` => fine, this state is not used, so we can use it.
1.  otherwise, the block has been used, but if `transferState.blockIdx >= S.blocks.length`, it means the state is used by a block that has been reverted;
1.  otherwise, the state is used by a previous block, but that block on the current virtual chain may be different from the block that actually consumed the state in the first place.... to check if this is the case, we use the Ethereum block number, if for the same virtual block index, the Ethereum block height is larger than the one recorded in the state, we say the vblock is reverted so the state can be reused.

We are actually assuming that **a reverted vblock must have a bigger Ethereum height than the one it replaced**.

dong77: Another approach is to record the Ethereum block hash in our virtual block, then the above assumption doesn't have to hold. It will consume more gas though. 

Is it technically possible for the operator to commit a vblock and then immediately revert it in the same Ethereum block?  This will invalidate the above assumption. Therefore we may need to do certain checking in the `revertBlock` funciton.

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
dong77: I see your point: conditions with the same `from, to, token, amount` but different salts are actually equal, the operator has the freedom to choose one of them equliavantely.

My concern is that now the **definition of conditions** is basically the tuple of `from, to, token, amount`. It serves fast withdrawal well enough, but may not be as generalized as it can be. I was thinking that **conditions can be anything programmable using solidity**, for example, if *address X has a balance bigger than 10000 Ether, then transfer 0.1 Ether from my offchain account to accountId 123*. -- It doesn't have a matching onchain transfer the order direction.

packages/loopring_v3/contracts/impl/relayer/Relayer.sol
```diff
@@ -0,0 +1,50 @@
...
+
+import "./RelayerData.sol";
+
+import "../../lib/Claimable.sol";
+import "../../lib/ReentrancyGuard.sol";
+
+
+/// @title An Implementation of IRelayer.
+/// @author Brecht Devos - <brecht@loopring.org>
+contract Relayer is Claimable, ReentrancyGuard, RelayerData
```
dong77: Relayer => Agenct/Delegate

packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
```diff
@@ -0,0 +1,92 @@
...
+
+import "../../iface/IExchangeV3.sol";
+import "../../iface/ICookieJarContract.sol";
+
+import "../../lib/MathUint.sol";
+import "../../thirdparty/ECDSA.sol";
+
+/// @title An Implementation of IRelayer.
+/// @author Brecht Devos - <brecht@loopring.org>
+contract FastWithdrawal is RelayerData
```
dong77: FastWithdrawal should be Relayer module right?  so it doesn't need to extend RelayerData.

packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
```diff
@@ -0,0 +1,92 @@
...
+                to,
+                token,
+                uint(fAmount).decodeFloat(24),
+                feeToken,
+                uint(fFee).decodeFloat(16),
+                onchainFeePayment,
+                salt
+            )
+        );
+        require(hash.toEthSignedMessageHash().recover(signature) == from, "UNAUTHORIZED");
```
dong77: why not just check msg.sender == from?

packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
Brechtpd: > Now I understand the purpose of keeping the record of Ethereum block numbers in a virtual block.
> Just to straight my mind:
> 
>     1. if `transferState.blockIdx == 0` => fine, this state is not used, so we can use it.
> 
>     2. otherwise, the block has been used, but if `transferState.blockIdx >= S.blocks.length`, it means the state is used by a block that has been reverted;
> 
>     3. otherwise, the state is used by a previous block, but that block on the current virtual chain may be different from the block that actually consumed the state in the first place.... to check if this is the case, we use the Ethereum block number, if for the same virtual block index, the Ethereum block height is larger than the one recorded in the state, we say the vblock is reverted so the state can be reused.

Totally correct.

> We are actually assuming that **a reverted vblock must have a bigger Ethereum height than the one it replaced**.

True.





> Another approach is to record the Ethereum block hash in our virtual block, then the above assumption doesn't have to hold. It will consume more gas though.

Yes, I think that would also work.


> Is it technically possible for the operator to commit a vblock and then immediately revert it in the same Ethereum block? This will invalidate the above assumption. Therefore we may need to do certain checking in the `revertBlock` funciton.

Yes, it is currently possible to revert a block immediately (the delay until the proof generation period has expired was removed). The check is `require(S.blocks[transferState.blockIdx].ethereumBlockNumber > transferState.ethereumBlockNumber)` so the operator wouldn't even be able to reuse the request in a new block in the same Ethereum block, in this special case the operator needs to wait at least 1 Ethereum block before the conditional transfer can be reused in a new block, so I don't think it's a problem. But I didn't think of this and an extra check in `revertBlock` is a good idea.

packages/loopring_v3/contracts/impl/relayer/Relayer.sol
Brechtpd: > My concern is that now the definition of conditions is basically the tuple of from, to, token, amount. It serves fast withdrawal well enough, but may not be as generalized as it can be. I was thinking that conditions can be anything programmable using solidity, for example, if address X has a balance bigger than 10000 Ether, then transfer 0.1 Ether from my offchain account to accountId 123. -- It doesn't have a matching onchain transfer the order direction.

Why would it need a matching onchain transfer? What you described can be done like this:

```
function executeConditionalTransfer(
  address targetAddress,
  address to
  )
  external
{
  require(targetAddress.balance >= 10000 eth, "CONDITION_FAILED");

  // Approve the conditional transfer
  address token = 0x0; // ETH
  IExchangeV3(exchangeAddress).approveConditionalTransfer(
    msg.sender,
    to,
    token,
    1 eth,
    token,
    0,
    uint32(now)
  );
}
```

The tuple of `from`, `to`, `token`, `amount` is of the conditional transfer itself, which is completely independent of the logic done on-chain. It can even be a simple fee payment for the gas cost to caller in exchange for doing the  smart contract logic for the user. It doesn't have to match anything onchain, the module can do any logic it wants, it can even allow multiple conditional transfers if needed.

Brechtpd: > Relayer => Agenct/Delegate

I quite like Agent. :)

packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
Brechtpd: This was to allow sharing data between modules. And because the modules are called with delegatecall (to keep `msg.sender` the Relayer contract) they will share state (it's essentially a proxy with many implementation contracts). So it depends on what the process will be for adding a module, but it's certainly not needed. There are many ways to do it.

packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
Brechtpd: > For fast withdrawals/cookie jar use it is extremely important that only the owner of the cookie jar contract can actually execute the request. Or he has at least some guarantees from the operator. The owner of the cookie jar contract takes the risk that the user actually has the amount requested offchain AND that this amount will still be available when the conditional transfer is actually included in a block. So in practice the owner of the cookie jar and the operator will be the same entity or work closely together/come up with some kind of insurance so that this can be ensured. So a user can request a fast-withdrawal (with an ECDSA signed request in this PR) but the actual execution of the fast withdrawal (the onchain transfer) cannot be initiated by anyone other than the operator/cookie jar owner.

Essentially we don't know the offchain balance of a user onchain, so we can't allow a user to do a fast withdrawal of 100ETH if his offchain balance is only 1ETH.

Brechtpd:
> @Brechtpd Read the code again, I have a better understanding, but I need your help to fully understand the process of fast withdrawal. Can you tell me what the workflow would be in a step-by-step way (the interaction between user and operator) ?

1. The user signs the fast withdrawal request with ECDSA using his normal Ethereum keys.
2. The user sends the request to the operator/cookie jar owner.
3. The operator/cookie jar owner checks if
  a) The user actually has the amount requested in the fast withdrawal in his offchain balance
  b) The cookie jar contract has enough funds to actually do the transfer onchain for the amount requested by the user
  c) Any other criteria the cookie jar owner/operator decide on: how much fee the user wants to pay, how many tokens the user has of some specific token, ...

The request can be dropped entirely or it can actually be used. If it's used we go to the next steps.

5. The fast withdrawal function in the FastWithdrawal module is called by an address approved by the cookie jar owner. This will also approve the conditional transfer in the exchange so it can be processed by the operator.
6. The operator immediately 'locks' the amount of tokens the user gets in the fast withdrawal so he can be sure the user cannot spend them while trading/transferring/... 
7. The user has his tokens immediately so for the user it's not important when the conditional transfer is actually processed in a block or is finalized. For the cookie jar owner it is important because he wants to refill the cookie jar contract regularly.

packages/loopring_v3/contracts/impl/relayer/Relayer.sol
Brechtpd: Some other quick examples of modules:

Using onchain liquidity for simple trades without order matching:
```
function executeOnchainLiquidityTrade(
  address tokenSellerAddress,
  address accountOwner,
  address cookieJarOwner,
  address tokenS,
  address tokenB,
  uint    fAmount,
  uint32  salt,
  bytes   calldata signature
  )
  external
{
  // Check the signature
  bytes32 hash = keccak256(
    abi.encodePacked(
      tokenSellerAddress,
      accountOwner,
      cookieJarOwner,
      tokenS,
      tokenB,
      uint(fAmount).decodeFloat(24),
      salt
    )
  );
  require(hash.toEthSignedMessageHash().recover(signature) == accountOwner, "UNAUTHORIZED");

  // Transfer the tokens here from the cookie jar contract
  ICookieJarContract(cookieJarContractAddress).transfer(
    address(this),
    tokenS,
    uint(fAmount).decodeFloat(24)
  );
  // Alternatively, we could also transfer tokens directly from the user's own wallet like this if possible:
  // onchainTransferFrom(accountOwner, address(this), tokenS, uint(fAmount).decodeFloat(24));

  // Expects the tokens that are sold from 'this' and are also received on 'this' address
  // Approvals may need to be set etc..
  uint amountB = ITokenSeller(tokenSellerAddress).sellToken(
    tokenS,
    tokenB, 
    uint(fAmount).decodeFloat(24)
  );

  // Deposit the tokens from 'this' directly to the user's account
  exchange.depositTo(accountOwner, tokenB, amountB);

  // Approve the conditional transfer
  // (A fee can be charged in any token as well)
  IExchangeV3(exchangeAddress).approveConditionalTransfer(
    accountOwner,
    cookieJarOwner,
    token,
    fAmount,
    token,
    0,
    salt
  );
}
```

Using the Relayer's authorization capabilities to do a user's deposit:

```
function executeDeposit(
  address accountOwner,
  address token,
  uint    fAmount,
  uint32  salt,
  bytes   calldata signature
  )
  external
{
  // Check the signature
  bytes32 hash = keccak256(
    abi.encodePacked(
      accountOwner,
      token,
      uint(fAmount).decodeFloat(24),
      salt
    )
  );
  require(hash.toEthSignedMessageHash().recover(signature) == accountOwner, "UNAUTHORIZED");

  // Deposit the tokens from the user's wallet
  exchange.deposit(accountOwner, token, amount);

  // Charge a fee here for the gas cost with a conditional transfer from 
  // the user's offchain balances or do an onchain transfer from the user's wallet
  // ...
}
```

dong77:
@Brechtpd, thank you for answering all my questions. I think I now have a pretty good understanding of your design, but correct me if I'm still wrong.

## The Agent design

I do like the way the Agent contract is designed. This opens a door for authorized
modules to do wonderful things on behalf of users.

We need to be really careful about security, as a module supported by the Agent can
transfer all money away from a user's account if it passes authorization check.

I wonder if it is possible to 1) use a unified function for all modules so all modules look similar, and 2) check user authorization in the Agent contract before delegating the call. These are not important feedback though.

## Conditional Transfers

I guess we have some differences in the design of Conditional Transfers.
Your approach is to create approval records using the `approveConditionalTransfer` function
in the core Exchange contract, and requires each Conditional Transfer request to consumes a matching approval.

My suggestion is to use a function in another contract -- which is not part of the core protocol --
to validate if a condition is satisfied. You can think of it as **moving
your `approveConditionalTransfer` method to another contract**, so the Exchange contract
doesn't know HOW exactly the approval/condition checking is performed, it only knows which contracts to invoke for condition validation.

Assuming we have `RequestValidator`s defined as follows:

```solidity
interface RequestValidator {
    uint validatorID; // no-zero

    // throws in case of error
    function validate(
        uint exchangeId,
        unit blockType,
        bytes requestData // assuming each request is unique with help of the nonce
    );
}
```

Your `approveConditionalTransfer` funciton can be implemented  in a permissioned
RequestValidator contract, and the contract's `validate` function handles many logics
inside your current `validateConditionalTransfers` function.

With such an approach, we only need to tell the core Exchange contract which
`RequestValidator` needs to be called for which request. This can be easlily
done with a map-based registration in the Exchange contract, such as:

```solidity
    registerRequestValidator(address validatorAddress) onlyOwner;
    getRequestValidator(validatorId) internal returns (RequestValidator);
```

Since `RequestValidator`s don't move user's fund, we can allow exchanges to
develop and register their own validators without restriction.



### Onchain and Offchain Authorization

Your idea of allowing offchain requests not to have user-signatures is great, I think we should support it. If an offchain request's `validatorID` is 0, the circuit shall require the offchain EDdSA signature for the request must exist. Otherwise, we can rely on the `RequestValidator` for onchain authorization validation. (it is possible that both offchain and onchain signatures are checked).


### Fast withdrawal

The fast withdrawal feature can then be built on top of the Agent and the RequestValidator concept.


```solidity

contract FastWithdrawal is Claimable, RequestValidator, AgentModule {
    uint validatorId = 1;
    bool checkOwnerSignature = false; // allows two authrization checking modes.

    constructor(bool _checkOwnerSignature) {
        checkOwnerSignature = _checkOwnerSignature;
    }

    function validate(
        uint exchangeId,
        unit blockType,
        bytes requestData // assuming each request is unique with help of the nonce
    ) {
        
        // Consumes the approval record
        uint approvalId = ....
        require(consumeApproval(approvalId), "NOT_APPROVED");
    }

    function executeFastWithdrawal(...) {
        // 1. Check user authorization
        if (checkOwnerSignature) {
            // called by a user
            require(hash.toEthSignedMessageHash().recover(signature) == from, "UNAUTHORIZED");
        } else {
           // called by the operator
            require(msg.sender == owner);
        }
        // 2. Create approval record
        this.approveConditionalTransfer(...);
        // 3. Make the transfer
        Exchange(exchangeAddr).transferTokenFrom(...)
    }
}

```

If the operator deploys the above contract with `checkOwnerSignature = true`, it works exactly as your design; if `checkOwnerSignature = false`, then the operator will require and check the offchain EDdSA signature in the circuits.


#### Differences in User Interaction

- If `checkOwnerSignature = true`: a user asks the operator for an ECDSA signature, then **the user** will interact with RequestValidatorModule by calling `executeFastWithdrawal`. The operator will create an unsigned offchain transfer account. Here we need to give the operator's signature a timeout, to prevent the user from holding the signature for too long.

- If `executeFastWithdrawal = false`, the user initiates a fast withdrawal with a signed offchain transfer request, then the operator will decide whether or not to accept such request, and if the request is accepted, **the operator**, not the user, will call  `executeFastWithdrawal`.


### Independent CookieJar contract

The above implementation actually uses the cryptoassets that have an account ID in the DEX. But it's also possible (sometimes even preferred) to have a cookie jar contract to manage the fast withdrawal funds. In that case, the `FastWithdrawal` can choose not to extend `AgentModule`.

packages/loopring_v3/contracts/impl/relayer/FastWithdrawal.sol
Brechtpd: The user gives permission with a signature, the cookie jar owner gives permission by actually doing the fast withdrawal. In `CookieJarContract.transfer` this check is done:

```
// Only allow an authorized transaction origin to transfer tokens out
require(authorizedAddresses[tx.origin], "UNAUTHORIZED");
```

There are of course different ways to do this, but I think this would work okay.

Brechtpd:
> @Brechtpd, thank you for answering all my questions. I think I now have a pretty good understanding of your design, but correct me if I'm still wrong.
> ## The Agent design
> 
> I do like the way the Agent contract is designed. This opens a door for authorized
> modules to do wonderful things on behalf of users.
> 
> We need to be really careful about security, as a module supported by the Agent can
> transfer all money away from a user's account if it passes authorization check.

Yes, there are large risks when there are mistakes in the modules. So modules should be as easy as possible to write, and in general I think a single deployed module should be usable by all exchanges (as I don't think they will need to store a lot of exchange-specific state). This way it's easy for e.g. us to approve the module address and exchanges can just whitelist. 

> I wonder if it is possible to 1) use a unified function for all modules so all modules look similar, and 2) check user authorization in the Agent contract before delegating the call. These are not important feedback though.

1) A unified function isn't necessary I think (or this is already the case depending how you look at it) because you can just call a specified function using built in function encoding from solidity. If you check the Relayer contract the main entry point is always `executeTransaction`. All function calls in all modules always go through this function. This is very easy to do, the data in `bytes memory transaction` has the function and all function parameters encoded so the correct function in the module is called with the correct input parameters. This way we don't have to manually decode all inputs from some byte array ourselves, we just use the default decoding from solidity.

2) I think this is hard because there are many ways authorization can be done: `msg.sender`, `tx.origin`, ECDSA signatures, EdDSA signatures, external contracts, a state variable, the gas network API, ... And this is just for the authorization of a single user. In some cases it may be necessary to authenticate multiple users in multiple ways. I don't think we can settle for a single method (something like multihash like we did for protocol 2 could in theory work I guess, but I don't think that would make things simpler).

> ## Conditional Transfers
> 
> I guess we have some differences in the design of Conditional Transfers.
> Your approach is to create approval records using the `approveConditionalTransfer` function
> in the core Exchange contract, and requires each Conditional Transfer request to consumes a matching approval.
> 
> My suggestion is to use a function in another contract -- which is not part of the core protocol --
> to validate if a condition is satisfied. You can think of it as **moving your `approveConditionalTransfer` method to another contract**, so the Exchange contract
> doesn't know HOW exactly the approval/condition checking is performed, it only knows which contracts to invoke for condition validation.
> 
> Assuming we have `RequestValidator`s defined as follows:
> 
> ```solidity
> interface RequestValidator {
>     uint validatorID; // no-zero
> 
>     // throws in case of error
>     function validate(
>         uint exchangeId,
>         unit blockType,
>         bytes requestData // assuming each request is unique with help of the nonce
>     );
> }
> ```
> 
> Your `approveConditionalTransfer` funciton can be implemented in a permissioned
> RequestValidator contract, and the contract's `validate` function handles many logics
> inside your current `validateConditionalTransfers` function.
> 
> With such an approach, we only need to tell the core Exchange contract which
> `RequestValidator` needs to be called for which request. This can be easlily
> done with a map-based registration in the Exchange contract, such as:
> 
> ```solidity
>     registerRequestValidator(address validatorAddress) onlyOwner;
>     getRequestValidator(validatorId) internal returns (RequestValidator);
> ```

I think this is a nice design. Though I think this would mostly be useful to save some gas in some cases (like no replay protection/revert handling necessary onchain when using an EdDSA signature in the circuit because that already uses a nonce/is automatically reverted). I've been trying to think of different kinds of `RequestValidator`s but I always end up with something similar as in this PR. Could you give some examples of different `RequestValidator`s?

The EdDSA signatures we use in the circuits can also be validated onchain, but because we use Poseidon for hashing the data it would be expensive to actually do it though, but I would assume that precompiles for these SNARK friendly hash functions will be added in the future, just like the blake2 hash function is added in Instanbul for better interop with Zcash. But currently they are also quite limited because the request just approves some transfer between accounts and a fee to the operator using some `validatorID`, so additional inputs/parameters need to be authenticated in a different way by the user onchain.

> 
> Since `RequestValidator`s don't move user's fund, we can allow exchanges to
> develop and register their own validators without restriction.

I don't think that's true. The exchange can just write a `RequestValidator` that doesn't require an EdDSA signature and approves any transfer from a user's account to an account controlled by the exchange for example.

Writing a secure `RequestValidator` will generally not be very easy and you kind of need to understand how to handle replay attacks/reverts correctly. So having a single code path for the internal working of the conditional transfers would be nice.

And moving most of this code out the Exchange contract is a good idea of course!

> ### Onchain and Offchain Authorization
> 
> Your idea of allowing offchain requests not to have user-signatures is great, I think we should support it. If an offchain request's `validatorID` is 0, the circuit shall require the offchain EDdSA signature for the request must exist. Otherwise, we can rely on the `RequestValidator` for onchain authorization validation. (it is possible that both offchain and onchain signatures are checked).
> ### Fast withdrawal
> 
> The fast withdrawal feature can then be built on top of the Agent and the RequestValidator concept.
> 
> ```solidity
> contract FastWithdrawal is Claimable, RequestValidator, AgentModule {
>     uint validatorId = 1;
>     bool checkOwnerSignature = false; // allows two authrization checking modes.
> 
>     constructor(bool _checkOwnerSignature) {
>         checkOwnerSignature = _checkOwnerSignature;
>     }
> 
>     function validate(
>         uint exchangeId,
>         unit blockType,
>         bytes requestData // assuming each request is unique with help of the nonce
>     ) {
>         
>         // Consumes the approval record
>         uint approvalId = ....
>         require(consumeApproval(approvalId), "NOT_APPROVED");
>     }
> 
>     function executeFastWithdrawal(...) {
>         // 1. Check user authorization
>         if (checkOwnerSignature) {
>             // called by a user
>             require(hash.toEthSignedMessageHash().recover(signature) == from, "UNAUTHORIZED");
>         } else {
>            // called by the operator
>             require(msg.sender == owner);
>         }
>         // 2. Create approval record
>         this.approveConditionalTransfer(...);
>         // 3. Make the transfer
>         Exchange(exchangeAddr).transferTokenFrom(...)
>     }
> }
> ```
> 
> If the operator deploys the above contract with `checkOwnerSignature = true`, it works exactly as your design; if `checkOwnerSignature = false`, then the operator will require and check the offchain EDdSA signature in the circuits.

Any specific reason an EDdSA signature is used when `checkOwnerSignature = false`? There is some gas that can saved (like the ECDSA signature that doesn't need to be checked onchain for ~3000gas but that's not a large part of the actual cost of doing the fast withdrawal) so I'm wondering if there is any other extra benefit.

> #### Differences in User Interaction
> 
>     * If `checkOwnerSignature = true`: a user asks the operator for an ECDSA signature, then **the user** will interact with RequestValidatorModule by calling `executeFastWithdrawal`. The operator will create an unsigned offchain transfer account. Here we need to give the operator's signature a timeout, to prevent the user from holding the signature for too long.
> 
>     * If `executeFastWithdrawal = false`, the user initiates a fast withdrawal with a signed offchain transfer request, then the operator will decide whether or not to accept such request, and if the request is accepted, **the operator**, not the user, will call  `executeFastWithdrawal`.
> 
> 
> ### Independent CookieJar contract
> 
> The above implementation actually uses the cryptoassets that have an account ID in the DEX. But it's also possible (sometimes even preferred) to have a cookie jar contract to manage the fast withdrawal funds. In that case, the `FastWithdrawal` can choose not to extend `AgentModule`.

The `FastWithdrawal` contract could also be the account owner itself (I did that by letting the `CookieJarContract` inherit from `AccountContract`).

----------

So I think we basically agree on 90% of the design. :)

Here's a list of things I think we either don't agree on yet or just some general open questions:

1) I'm not 100% convinced multiple `RequestValidator`s are really needed.
  - There is a security risk 
  - They are not easy so make
  - If there are multiple implementations it makes it harder for 3rd party services to write modules that integrate with all Loopring exchanges.

So unless there are good use cases I would prefer a single code path that can be reused and can be used to implement pretty much any functionality in AgentModules.

2) I'm not the biggest fan of the EdDSA signature checks in the circuits for the conditional transfers.
  - They only authorize some very limited operation onchain (when we do the onchain transaction we actually have no idea the user actually authorized anything). So any additional inputs like a withdrawal address different from the accounts owner's address needs to be authorized differently.
  - The operator needs to receive the signature in an offchain method otherwise the request can't be used and the request remains unprocessed.
  - Things are a bit more complex
  - ... but they can save some gas in some cases.

So basically: Can you make a more efficient fast withdrawal with a custom `RequestValidator` and EdDSA? Yes! Should we support it? I'm not sure! :) I think most of the savings are also only possible for 'simple' fast withdrawals (so only withdrawing to the account owner address), otherwise there are extra costs as well.

For the general extensibility I think we'll have to move things out of the Exchange anyway, so the conditional transfer logic will certainly be a part of that. Some quick ideas: I'm thinking of something like keeping the Exchange really basic. Just storage of blocks, accounts and tokens with some functions to manipulate and verify them (get/set an account, add/get a block). We then have multiple `ExchangeModules` (which will mainly be build around the functionality of a circuit) that can have their own state and which will modify the Exchange. So instead of moving only the block processing out of the exchange like in #321 we can also move out all functionality related to it (like for deposits we also move out all deposit related functions). The functionality of an Exchange is than just the collection of `ExchangeModule`s plugged into the Exchange. If an Exchange wants to add some new feature they would write a new `ExchangeModule` (or modify an existing one), write a new circuit if necessary, and plug it into their exchange as well. If they want a different `approveConditionalTransfer` than they can add/modify an `ExchangeModule` for that like they want. `ExchangeModule`s could be written by the exchange as well and could be approved by us (but if they start modifying things it's mostly up to them to keep the exchange save I'd say).

dong77:
> Any specific reason an EDdSA signature is used when checkOwnerSignature = false? 

Anawer: Not at all, I just happened to use EDdsa as an example, and maybe I should use something like multihash.

> Could you give some examples of different RequestValidators?

This is hard, to be honest, I've been wondering about this too...

> I don't think that's true. The exchange can just write a RequestValidator that doesn't require an EdDSA signature and approves any transfer from a user's account to an account controlled by the exchange for example.

I agree. Seems we need to make sure enabling a validator shall also imply a waiting period.

> The FastWithdrawal contract could also be the account owner itself

Yes, that's possible.

> I'm not 100% convinced multiple RequestValidators are really needed.
To be honest, I'm not sure either. And even if we design the system as I suggested, it's not generic enough -- the Validator cannot accept additional parameters.

>  I'm thinking of something like keeping the Exchange really basic. 
This is really something I want too, but I thought the copying of large calldata to sub-contract is expensive. Unless for each type of block, **the relayer calls the module contract's function directly and that function then calls the Exchange's function to save some very small amount of block metadata ** (instead of the relayer calling Exchange contract which then calls one module).

Suggestions: So how about we implement fast withdrawal the way you implement it, but drop the Agent feature. Then in the next refractory, we put extensibility as a top priority.

We can also choose to NOT implement fast-withdrawal at all for this release and get right to the big refractory for extensibility.


-----

After I just wrote the above content, I realize that Fast Withdrawal is an example of **the same circuits with different onchain behaviors**. If the circuits are different, we can use a different ExchangeModule (which means each ExchangeModule has a matching circuit), but what if the circuits are the same and only there are additional onchain verification/validation, shall the design of ExchangeModule support **additional onchain validation logic**? In other words, will there be TransferModule and ConditionalTransferModule after the ExchangeModule concept is introduced.

Brechtpd:
> This is really something I want too, but I thought the copying of large calldata to sub-contract is expensive. Unless for each type of block, **the relayer calls the module contract's function directly and that function then calls the Exchange's function to save some very small amount of block metadata** (instead of the relayer calling Exchange contract which then calls one module).

This is exactly how I also saw it. Committing a block would just be something like this:

```
function commitBlock(
  bytes32 merkleRoot,
  bytes32 publicDataHash
  )
  external
```

Just enough to advance the chain of blocks and so that blocks are still available and verified on the Exchange itself (so all ExchangeModules have access to this important state by getting it from the Exchange contract). All other logic/data is in the ExchangeModules.

> Suggestions: So how about we implement fast withdrawal the way you implement it, but drop the Agent feature. Then in the next refractory, we put extensibility as a top priority.
> 
> We can also choose to NOT implement fast-withdrawal at all for this release and get right to the big refractory for extensibility.

I don't think it matters that much because there aren't that many changes for the conditional withdrawals. So maybe first the big refactor if we can decide on the design quickly because that's the bigger change.

> After I just wrote the above content, I realize that Fast Withdrawal is an example of **the same circuits with different onchain behaviors**. If the circuits are different, we can use a different ExchangeModule (which means each ExchangeModule has a matching circuit), but what if the circuits are the same and only there are additional onchain verification/validation, shall the design of ExchangeModule support **additional onchain validation logic**? In other words, will there be TransferModule and ConditionalTransferModule after the ExchangeModule concept is introduced.

I think we can do both. I think ExchangeModules should be as independent as possible, having the Exchange as the main dependency. If ExchangeModules depend on each other I think it gets more complicated real quickly. So you could do a TransferModule and a ConditionalTransferModule using the same circuit, but committing a single block with requests from both those ExchangeModules will probably be difficult, so having the all the logic for a single circuit inside a single ExchangeModule would be more efficient. But I certainly haven't worked out all the details, I only converted small parts of our contracts to see if it's even possible.

dong77:
Closing this PR as it is followed up by this one https://github.com/Loopring/protocols/pull/575

PR #614, opened 2019-10-22
emit BlockFinalized event only once

kongliangzhong:


packages/loopring_v3/contracts/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -170,10 +170,14 @@ library ExchangeBlocks
         uint idx = S.numBlocksFinalized;
         while (idx < S.blocks.length &&
             S.blocks[idx].state == ExchangeData.BlockState.VERIFIED) {
-            emit BlockFinalized(idx);
             idx++;
         }
-        S.numBlocksFinalized = idx;
+
+        // Only emit BlockFinalized event for the last verified block.
+        if (idx > S.numBlocksFinalized) {
```
dong77: 这个地方逻辑对吗？如果 idx === S.numBlocksFinalized,在之前的代码里面，`S.numBlocksFinalized = idx;`会被执行的，新的逻辑好像就不执行了。

dong77: 这个代码是不是有bug。
在while中，idx++会最多把idx变成S.blocks.length，因此下面的if永远都不会是true把？

dong77: 是不是应该做
```solidity
bool emitEvent = false;
while(..) {
   emitEvent = true;
}

if (emitEvent) emit ...

```

dong77: @Brechtpd please confirm if you also think that the current change is buggy.
I think `idx > S.numBlocksFinalized` will never be true.

Brechtpd: I think it's correct.  If `numBlocksFinalized` is 4 (blocks [0, 3] finalized) and block with `blockIdx == 4` is verified, `idx == 4` before the loop and `idx == 5` after the loop, so `idx > numBlocksFinalized` is true.

I kind of liked that each block got a `BlockFinalized` even if not necessary, but yeah it can save a bit of gas (1000 gas/block) in some cases if we don't do it.

dong77: Alright, if it is correct, that will be fine.

PR #615, opened 2019-10-22
force exchange to do withdraw distributions while in shutdown mode

kongliangzhong:


packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
```diff
@@ -196,8 +196,10 @@ library ExchangeAdmins
         // Another requirement is that the last block needs to be committed
         // longer than MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS so the exchange can still be fined for not
         // automatically distributing the withdrawals (the fine is paid from the stake)
+        uint shutdownBlocksCount = S.blocks.length - S.shutdownStartBlock - 1;
         require(
-            now > lastBlock.timestamp + ExchangeData.MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS(),
+                now > lastBlock.timestamp +
+                      shutdownBlocksCount.mul(ExchangeData.MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS()),
```
Brechtpd: For each block independently the operator has `ExchangeData.MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS()` seconds to do the withdrawals, after that a fine is paid. So this extra time is to allow other people (not the operator) to  do the withdrawals and get paid a fine for a much longer time? If `MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS()` is set to `1 day` than this will be an enormous amount of time even with a limited number of blocks used for the withdrawals (1000 blocks  * 1 day -> ~3 years locked up). This could lock up the stake pretty much indefinitely for even medium size exchanges. 
I think a good compromise would be to add some extra time, but still a fixed value. Something like `ExchangeData.MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS() * 10` regardless of the number of blocks used to do the withdrawals. This still gives plenty of time for other people to do the withdrawals and get paid if necessary.

dong77: I don't understand why we need this change. @kongliangzhong Lets talk offline.

dong77: The name `MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS` is a bit confusing, maybe rename it to
`MAX_TIME_TO_COMMIT_ALL_WITHDRAWAL_APPROVALS` because the operator is not really required to transfers all fund back to users address, he/she just need to make sure users can do the transfers, am I right, @Brechtpd

Brechtpd: The operator _is_ required to do the actual transfers, if not he pays a fine using the exchange stake. A user can do the withdrawal token transfer himself (or someone else) from the moment the block that produced the approved withdrawal is finalized (the approved withdrawal data is available when the block is committed). The distribution is the actual token transfer back to the user's address.

dong77: Maybe we should use two constants:
```
MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS_NORMAL_MODE  = 1 day
MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS_SHUTDOWN_MODE = 7 days
```

Brechtpd: There's nothing special about the actual distribution of withdrawals in shutdown mode so I don't think we need 2 different values. The only thing that can happen is that the exchange owner withdraws the exchange stake before all withdrawals are distributed (which is fine because users can still do the withdrawals themselves, and the token transfer while distributing can even fail because of the gas limit/withdrawal) so no reward can be given to the actual distributor using the exchange stake. While I think the code on the master is totally fine, there is limited time for other distributors to do the distributions (when the operator doesn't do it) while still getting paid. 

So we can just add some extra time of e.g. `MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS * 10` before the exchange stake can be withdrawn to give some extra time for distributors to still get paid before the exchange stake is completely withdrawn. This of course could also be a constant independent of `MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS`, but the constant would be something like `MIN_TIME_UNTIL_EXCHANGE_STAKE_IS_WITHDRAWABLE` [like I did for 3.1](https://github.com/Loopring/protocols/blob/8a34ef3f9e93f010ec8b756054216225ae26b4e9/packages/loopring_v3.1/contracts/impl/libexchange/ExchangeData.sol#L124) for different reasons.

kongliangzhong: Make a compromise, MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS_SHUTDOWN_MODE = MAX_TIME_TO_DISTRIBUTE_WITHDRAWALS * 10 seems better :)

Issue #603, opened 2019-10-17
[Protocol3] Finalize constants


dong77:
@letsgoustc Please lead the effort to discuss and finalize the constant variables in protocol 3.0. This has to go into beta4. (I assume some tests will fail, so please apply the change to a release branch, not the master).

Some discussion was captured in this doc: https://docs.google.com/spreadsheets/d/1XKOKUtLdh7PZgKWs5HRxK_wrCyt4FbHhRdL4SAW7_Cw/edit#gid=1428741813

yueawang:
#621

dong77:
Done

PR #621, opened 2019-10-23
[Protocol3] Add missed const cases

yueawang:
After constant value review, few const were not covered by the test.
Add 2 cases for MAX_GAP_BETWEEN_FINALIZED_AND_VERIFIED_BLOCKS and MIN_AGE_PROTOCOL_FEES_UNTIL_UPDATED.
However as MAX_GAP_BETWEEN_FINALIZED_AND_VERIFIED_BLOCKS now is 2500, it takes ~200s to complete, so far the case is skipped and it can be enabled after MAX_GAP_BETWEEN_FINALIZED_AND_VERIFIED_BLOCKS becomes small.

### DIFF
--- a/packages/loopring_v3/test/testExchangeBlocks.ts
+++ b/packages/loopring_v3/test/testExchangeBlocks.ts
@@ -307,6 +307,99 @@ contract("Exchange", (accounts: string[]) => {
           }
         });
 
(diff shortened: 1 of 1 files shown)


yueawang:
part of #603

PR #657, opened 2019-11-05
add a precondition for withdrawal non-LRC tokens

dong77:
We now require a user must hold ILoopring.minLRCBalanceForUserToWithdraw in order to send onchain withdrawal requests to withdraw non-LRC tokens.

### DIFF
--- a/packages/loopring_v3/contracts/iface/ILoopringV3.sol
+++ b/packages/loopring_v3/contracts/iface/ILoopringV3.sol
@@ -71,13 +71,14 @@ contract ILoopringV3 is ILoopring
     uint    public totalStake;
     address public blockVerifierAddress;
     address public downtimeCostCalculator;
-    uint    public maxWithdrawalFee;
     uint    public withdrawalFineLRC;
     uint    public tokenRegistrationFeeLRCBase;
     uint    public tokenRegistrationFeeLRCDelta;
     uint    public minExchangeStakeWithDataAvailability;
     uint    public minExchangeStakeWithoutDataAvailability;
     uint    public revertFineLRC;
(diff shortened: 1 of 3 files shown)


Brechtpd:
> In normal operation, this requirement will ensure users who left a Loopring-DEX end up with LRC in their accounts.

In normal operations users will just do off-chain withdrawal requests because that's cheaper (which won't have to pass this check). On-chain withdrawal requests are mainly for security so won't be used normally.
It also doesn't seem very user friendly to force users to buy LRC, withdraw the LRC to their wallet, and only then would they be able to withdraw any other token if they didn't have any LRC to begin with. That'd be a strange requirement and normal users would be forced to buy a random token (from their point of view) just to be able to use an exchange for no particular reason.

> In early product launch, this requirement ensures getting onboard only users with enough LRC (but we need to make this requirement clear to users.)

What if we do the check while depositing instead? You can't deposit unless you have a minimum amount of LRC. Users would need to buy the LRC somewhere else but there are enough options to do that.

dong77:
To be honest, I want each user to possess some LRC to use the system... any suggestion how this can be achieved in a non-offensive way?

Brechtpd:
Isn't this what the protocol fee indirectly does? For using the system users have to pay a fee and this fee is then given to users in the form of LRC that stake LRC or is burned. You can see both as users indirectly buying LRC to use the system, but completely transparent for normal exchange users.

It's not completely the same as owning X amount of LRC up front, but it's arguably much better in the long term because that X amount of LRC a user would need to own would have to be low to not lock users out. The protocol fee scales nicely with how much the user trades, so also with how much tokens a user has.

PR #689, opened 2019-11-14
[TokenBank] add core smart wallet related contracts

dong77:
The implementation is still in progress, but the overall design can be reviewed.

Each Wallet can have multiple Modules but need to have at least one Module otherwise its asset will be locked forever.  Adding new modules or deleting existing modules can not be done by the owner directly, but need to be called from an existing module, this is to guarantee no third-party modules can be added to a wallet by the wallet owner. 

Each module needs to be registered in ao ModuleRegistry, this is not done yet.


packages/tokenbank_v1/contracts/lib/Authorizable.sol
```diff
@@ -0,0 +1,113 @@
...
+        require(pos != 0, "NOT_FOUND");
+
+        uint size = authorizedAddresses.length;
+        if (pos != size) {
+            address lastOne = authorizedAddresses[size - 1];
+            authorizedAddresses[pos - 1] = lastOne;
+            positionMap[lastOne] = pos;
+        }
+
+        authorizedAddresses.length -= 1;
```
adamchen0510: Don't delete authorizedAddresses[size-1]?
And authorizedAddresses not using now, seems positionMap is enough now, for future?

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -0,0 +1,125 @@
...
+            if (msg.value > 0) {
+                emit Received(msg.sender, msg.value, msg.data);
+            }
+            return;
+        }
+
+        require(moduleIdx[module] !=0, "MODULE_UNAUTHORIZED");
+        assembly {
+            calldatacopy(0, 0, calldatasize())
+            let result := staticcall(gas, module, 0, calldatasize(), 0, 0)
```
Brechtpd: This is probably some temporary code (`staticcall` can't do any state changes), but will this be `delegatecall` or just `call`?

dong77: I copied the code from Argent and I also have the same questions. Take a look at their comment here:

https://github.com/dong77/argent-contracts/blob/34da49e6defded4cce9602ea30baa53ef99ae76a/contracts/wallet/BaseWallet.sol#L119

Maybe you can figure it out.

packages/tokenbank_v1/contracts/lib/Authorizable.sol
dong77: authorizedAddresses is now made as public so it can be queried.

packages/tokenbank_v1/contracts/impl/BaseModule.sol
```diff
@@ -0,0 +1,77 @@
...
+    function transact(
+        address wallet,
+        address to,
+        uint    value,
+        bytes   memory data
+        )
+        internal
+        returns (bytes memory result)
+    {
+        return Wallet(wallet).transact(to, value, data);
```
dong77: @Brechtpd if the data is large, I guess calling the wallet directly will cost more gas, is this a legit optimization in Argent's code?
https://github.com/dong77/argent-contracts/blob/34da49e6defded4cce9602ea30baa53ef99ae76a/contracts/modules/common/BaseModule.sol#L94

Brechtpd: I think solidity will generate comparable code, so I think the current code is fine. The only reason I can think of to do `abi.encodeWithSignature("invoke(address,uint256,bytes)", ...)` is so there's no need to define an interface.

One thing the Argent code can detect better is if the wallet address doesn't actually have the `invoke` function. The `call` will then return `success == false` which allows a better error message (I don't really know what the error message will be for the solidity code, but probably something more cryptic).

dong77: Thanks. 
The wallet contract is controlled by us so it will have the required method.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
Brechtpd: Ah interesting! So this will forward all `view` function calls on the wallet to the modules which some other smart contract using the wallet may depend on. Because it's a `staticcall` there's isn't really any risk of doing this because it's impossible to do any state changes. Because it seems riskless to me, I wonder why those functions need to be enabled for this forwarding to happen. I assume it's difficult otherwise to know when to forward the call or not (e.g. just a simple ETH transfer into the wallet should not do anything, but `msg.data.length > 0` is true).

dong77: Let's talk about sometime later. The whole idea of method registration in a wallet seems wired to me as all read-only methods can be invocated directly in modules.

Brechtpd: A function call done through `transact` originates from the wallet address (`msg.sender == wallet`), so if the dapp depends on `msg.sender` to implement something, it needs to go through the wallet first (the dapp doesn't know about any modules, it only knows the wallet's address). At least that's my understanding.

packages/tokenbank_v1/contracts/impl/WalletFactory.sol
```diff
@@ -0,0 +1,110 @@
...
+
+
+    function createWalletInternal(
+        address   _owner,
+        address[] memory _modules
+        )
+        internal
+        returns (address walletAddress)
+    {
+        SimpleProxy proxy = new SimpleProxy(walletImplementation);
```
Brechtpd: (Copied from older PR that was closed:) Will we (eventually) allow wallets to be created with `create2`? So that, for example, we can generate wallet addresses for free for all users without paying any gas if the address remains unused?

packages/tokenbank_v1/contracts/impl/WalletFactoryWithENS.sol
```diff
@@ -0,0 +1,84 @@
...
+            extendedModules[0] = address(this);
+            for(uint i = 0; i < _modules.length; i++) {
+                extendedModules[i + 1] = _modules[i];
+            }
+            walletAddress = createWalletInternal(_owner, extendedModules);
+
+            Wallet wallet = Wallet(walletAddress);
+            registerSubdomain(wallet, _subdomain);
+
+            wallet.removeModule(address(this));
```
Brechtpd: (Copied from older PR that was closed: ) `this` is added as a module to the wallet and then removed. I guess it will be necessary for the wallet to temporarily have `this` as a module?

dong77:
Yes, otherwise the factory cannot trigger transaction on the wallet itself for registration of ENS.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -0,0 +1,242 @@
...
+        require(_owner != address(0), "ZERO_ADDRESS");
+        require(_modules.length > 0, "EMPTY_MODULES");
+
+        owner = _owner;
+
+        emit Initialized(owner);
+
+        for(uint i = 0; i < _modules.length; i++) {
+            address module = _modules[i];
+            require(module != address(0), "NULL_MODULE");
```
Brechtpd: Seems like this does the same as `addModule` so it may not be a bad idea to create an `addModuleInternal` so the logic can be shared.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -0,0 +1,242 @@
...
+    {
+        return isAddressInSet(MODULE, _module);
+    }
+
+    function bindStaticMethod(bytes4 _method, address _module)
+        external
+        onlyModule
+        nonReentrant
+    {
+        require(methodToModule[_method] != address(0), "BAD_MODULE");
```
Brechtpd: Maybe this needs to be `require(methodToModule[_method] == address(0)`? Or it may not be a bad idea to remove the check all together so that this method can also be used to unbind methods by setting `_module == address(0)`?

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
```diff
@@ -0,0 +1,255 @@
...
+/// https://github.com/argentlabs/argent-contracts
+contract MetaTxModule is BaseModule
+{
+    using MathUint for uint;
+    uint constant public BLOCK_BOUND = 100;
+    uint constant public GAS_OVERHEAD = 30000;
+
+    struct WalletState {
+        uint nonce;
+        mapping (bytes32 => bool) execuitedMetaTxHash;
```
Brechtpd: Typo: `executedMetaTxHash`

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
```diff
@@ -0,0 +1,255 @@
...
+        uint    nonce,
+        bytes32 metaTxHash
+        )
+        private
+    {
+        if (nonce == 0) {
+            require(!wallets[wallet].execuitedMetaTxHash[metaTxHash], "DUPLICIATE_SIGN_HASH");
+            wallets[wallet].execuitedMetaTxHash[metaTxHash] = true;
+        } else {
+            require(nonce <= wallets[wallet].nonce, "NONCE_TOO_SMALL");
```
Brechtpd: I think this needs to be `require(nonce > wallets[wallet].nonce, "NONCE_TOO_SMALL");`

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
```diff
@@ -0,0 +1,255 @@
...
+            uint nonceBlock = (nonce & 0xffffffffffffffffffffffffffffffff00000000000000000000000000000000) >> 128;
```
Brechtpd: I don't think you need to do the AND if you shift to the right anyway, but I guess it can't hurt. :)

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: That's right, Brecht. The `transact` method is the killing feature for the wallet to transact on any arbitrary contracts with msg.sender == the_wallet_address.

packages/tokenbank_v1/contracts/impl/WalletFactory.sol
dong77: We will need to support CREATE2. I'll add a TODO here.

packages/tokenbank_v1/contracts/impl/WalletFactoryWithENS.sol
dong77: That's also my understanding. `registerSubdomain` will call the wallet's `transact` method, so this module must be in the initial module list.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: done.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: As you suggested:
```solidity
require(_method != bytes4(0), "BAID_METHOD");
        require(
            _module == address(0) || methodToModule[_method] == address(0),
            "BAD_MODULE"
);

```
this will allow unbinding of methods.

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
dong77: DONE

packages/tokenbank_v1/contracts/impl/ERC1271Module.sol
```diff
@@ -0,0 +1,105 @@
...
+    /// @dev This method is expected to be called from a wallet's default function.
+    function isValidSignature(
+        bytes memory _data,
+        bytes memory _signature
+        )
+        public
+        view
+        returns (bytes4)
+    {
+        return isValidSignatureFrom(address(this), _data, _signature);
```
Brechtpd: `address(this)` can't sign anything so it seems like this function will always simply return `false`.

packages/tokenbank_v1/contracts/lib/SignatureUtil.sol
```diff
@@ -0,0 +1,52 @@
...
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+
+/// @title Claimable
+/// @author Brecht Devos - <brecht@loopring.org>
+/// @dev Extension for the Ownable contract, where the ownership needs
+///      to be claimed. This allows the new owner to accept the transfer.
+contract SignatureUtil
```
Brechtpd: We can also just use the Open Zeppelin contract here:
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol

Brechtpd: Oh, I see it doesn't have the `index` functionality that is used here. Then I guess not.

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
```diff
@@ -0,0 +1,254 @@
...
+            );
+
+        return keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
+    }
+
+    /// @dev   Recovers the signer at a given index from a list of concatenated signatures.
+    /// @param metaTxHash The signed hash
+    /// @param signatures The concatenated signatures.
+    /// @param index The index of the signature to recover.
+    function recoverSigner(
```
Brechtpd: Should use the library function in `SignatureUtil`?

dong77: Done.

packages/tokenbank_v1/contracts/lib/SignatureUtil.sol
dong77: Made it a library.

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
dong77: done

packages/tokenbank_v1/contracts/impl/ERC1271Module.sol
dong77: I think it should now be fixed, but I need your confirmation.

dong77:
The way static methods are bound has been changed a bit. please take a look, @Brechtpd

packages/tokenbank_v1/contracts/impl/BaseModule.sol
```diff
@@ -0,0 +1,131 @@
...
+        address module
+        )
+        external
+        nonReentrant
+        onlyStricklyWalletOwner(wallet)
+    {
+        Wallet(wallet).addModule(module);
+    }
+
+    /// @dev Adds a module to a wallet. Callable only by the wallet owner.
```
adamchen0510: comments error

packages/tokenbank_v1/contracts/impl/BaseModule.sol
```diff
@@ -0,0 +1,131 @@
...
+    ///      Note that the current module must have been added to the wallet.
+    function initialize(address wallet)
+        external
+        onlyWallet(wallet)
+    {
+        bindStaticMethods(wallet);
+        emit Initialized(wallet);
+    }
+
+    /// @dev Initializes the module for the given wallet address.
```
adamchen0510: comments error

packages/tokenbank_v1/contracts/impl/ERC1271Module.sol
```diff
@@ -0,0 +1,91 @@
...
+    {
+        if (_address == address(0) || _signature.length != 65) {
+            return false;
+        }
+
+        bytes32 signHash = keccak256(
+            abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(_data))
+        );
+
+        return signHash.recoverSigner(_signature, 0) != _address;
```
adamchen0510: seems it should use == _address

dong77: fixed

packages/tokenbank_v1/contracts/impl/BaseModule.sol
dong77: fixed

packages/tokenbank_v1/contracts/impl/BaseModule.sol
dong77: fixed

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
```diff
@@ -0,0 +1,226 @@
...
+        bytes32 metaTxHash
+        )
+        private
+    {
+        if (nonce == 0) {
+            require(!wallets[wallet].executedMetaTxHash[metaTxHash], "DUPLICIATE_SIGN_HASH");
+            wallets[wallet].executedMetaTxHash[metaTxHash] = true;
+        } else {
+            require(nonce > wallets[wallet].nonce, "NONCE_TOO_SMALL");
+            require((nonce >> 128) <= (block.number + BLOCK_BOUND), "NONCE_TOO_LARGE");
```
letsgoustc: Why need such check or block.number?

packages/tokenbank_v1/contracts/lib/Managed.sol
```diff
@@ -0,0 +1,69 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+import "./NamedAddressSet.sol";
+
+
+contract Managed is NamedAddressSet
+{
```
letsgoustc: The contract name "Managed" looks confusing. Suggest to use another name

packages/tokenbank_v1/contracts/iface/Wallet.sol
```diff
@@ -0,0 +1,150 @@
...
+    ///      not been added to this wallet.
+    /// @param _method The method's 4-byte selector.
+    /// @param _module The module's address. Use address(0) to unbind the method.
+    function bindStaticMethod(bytes4 _method, address _module) external;
+
+    /// @dev Returns the module the given method has been bound to.
+    /// @param _method The method's 4-byte selector.
+    /// @return _module The address of the bound module. If no binding exists,
+    ///                 returns address(0) instead.
+    function staticMethodModule(bytes4 _method) public view returns (address _module);
```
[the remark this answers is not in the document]
dong77: 是的，因此不同的module的只读方法，如果bind到wallet里面，必须名字是不一样的。

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
dong77: 这个是防止用户的nonce一下子设置的非常大，然后就没有空间了。比如一下子是知道uint256.max -2，这样以后就空间了。

packages/tokenbank_v1/contracts/lib/Managed.sol
dong77: 这个是这样用的，和Owned, Claimable
```solidity
contract Abc is Managed {
}
```
当然，也可以叫Managable... 你觉得那个好。

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -0,0 +1,250 @@
...
+        require(to != address(this), "SAME_ADDRESS");
+        bytes memory result;
+        if (token == address(0)) {
+            result = transactInternal(to, value, "");
+        } else {
+            bytes memory data = abi.encodeWithSignature(ERC20_TRANSFER, to, value);
+            result = transactInternal(token, 0, data);
+        }
+
+        // TODO(daniel): Not sure if this will work, this need to be tested!!!
```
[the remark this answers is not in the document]
dong77: 这个地方我也不是很确定，不过是实现的细节，晚点回来再看。

dong77:
@letsgoustc 具体方法的实现细节可以先不管，肯定有很多bug，后续要通过测试逐一排查。
目前需要review的主要是核心设计，特别是合约之间的调用关系。目前调用角色主要包括这些：

- 钱包(Wallet)
- 钱包的Owner，这个可能是一个普通地址，也可能是任何一个合约地址，甚至包括另一个的Wallet。
- 钱包的Guardian，这个可能是一个普通地址，也可能是任何一个合约地址，甚至包括另一个的Wallet。
- 模块（module)，用来给wallet提供只读方法，或者允许钱包的owner直接调用这个模块的写方法；
- 任意第三方合约。整个设计应该允许把钱包作为msg.sender和任何一个第三方合约交互，也包括把钱包作为msg.sender和任何一个钱包模块交互）
- 钱包owner和钱包Guardian可以通过meta transaction来和任意一个第三方（包括一个module）交易，任何一个人都可以帮助这个元交易来付以太油费（前提是线下拿到这个元交易）；
- 钱包可以添加或者去掉Module，或者通过调用Module的方法（包括使用元交易）间接添加或者去掉Module。
- 钱包可以通过特殊的UpgraderModule升级（也可以使用元交易）。

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
letsgoustc: 那是不是可以参考以太的设计，下一次的TX只允许是上一次的nonce+1？

packages/tokenbank_v1/contracts/iface/Wallet.sol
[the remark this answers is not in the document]
letsgoustc: > 是的，因此不同的module的只读方法，如果bind到wallet里面，必须名字是不一样的。

这个感觉很难保证啊，我觉得可以设计成用sha256(函数编码的bytes4+moduleName)，再取前4个字节，也就是先做一个带上模块名字的Hash，这样就不用担心名字会重了

packages/tokenbank_v1/contracts/lib/Managed.sol
letsgoustc: > 这个是这样用的，和Owned, Claimable
> 
> ```solidity
> contract Abc is Managed {
> }
> ```
> 
> 当然，也可以叫Managable... 你觉得那个好。

Prefer Managable

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -0,0 +1,296 @@
...
+        returns (uint)
+    {
+        if (token == address(0)) {
+            return address(this).balance;
+        } else {
+            return ERC20(token).balanceOf(address(this));
+        }
+    }
+
+    function transferToken(
```
letsgoustc: 我们是不是也考虑类似Argent的设计，把转账操作也都放到Transfer模块里面去实现，这样保持BaseWallet合约极简化

packages/tokenbank_v1/contracts/modules/LockModule.sol
```diff
@@ -0,0 +1,188 @@
...
+
+import "../impl/MetaTxModule.sol";
+
+import "../storage/GuardianStorage.sol";
+
+
+/// @title LockModule
+/// @dev  A module for managing wallet locking and unlocking by guardians.
+///       Guardians of a wallet can use a meta-transaction to lock/unlock a wallet,
+///       or they can can the lock/unlock method directly.
```
letsgoustc: can call

letsgoustc:
> @letsgoustc 具体方法的实现细节可以先不管，肯定有很多bug，后续要通过测试逐一排查。
> 目前需要review的主要是核心设计，特别是合约之间的调用关系。目前调用角色主要包括这些：
> 
> * 钱包(Wallet)
> * 钱包的Owner，这个可能是一个普通地址，也可能是任何一个合约地址，甚至包括另一个的Wallet。
> * 钱包的Guardian，这个可能是一个普通地址，也可能是任何一个合约地址，甚至包括另一个的Wallet。
> * 模块（module)，用来给wallet提供只读方法，或者允许钱包的owner直接调用这个模块的写方法；
> * 任意第三方合约。整个设计应该允许把钱包作为msg.sender和任何一个第三方合约交互，也包括把钱包作为msg.sender和任何一个钱包模块交互）
> * 钱包owner和钱包Guardian可以通过meta transaction来和任意一个第三方（包括一个module）交易，任何一个人都可以帮助这个元交易来付以太油费（前提是线下拿到这个元交易）；
> * 钱包可以添加或者去掉Module，或者通过调用Module的方法（包括使用元交易）间接添加或者去掉Module。
> * 钱包可以通过特殊的UpgraderModule升级（也可以使用元交易）。

packages/tokenbank_v1/contracts/iface/Wallet.sol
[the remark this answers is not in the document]
dong77: Sorry I have to type English (switched to another computer).
We allow modules to bind static methods to a Wallet so we can add more "readonly" on the fly. For example, if we bind "getAbc(...)" then anyone can call "wallet.getAbc(...", this is done by going through the deafult functions. In the default function, we only know's function's 4-byte selector, not the target module. So we have to do a lookup with the selector. The selector's generation method is pre-defined by Ethereum, we cannot change it. Therefore, there is no way we can bind the same method (same selector) from two modules.

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: We can do this. I'll later create an issue so someone else can do it. I believe the process will help him to learn the design. I'll add a TODO.

dong77:
I'll do the renaming "Managed => Managable", "OwnerManaged => OwnerManagable" later.

packages/tokenbank_v1/contracts/lib/Managed.sol
dong77: done

packages/tokenbank_v1/contracts/impl/MetaTxModule.sol
dong77: 这个也是可以的，不过由于使用了meta transaction，就很难保证严格+1。

PR #690, opened 2019-11-14
[Protocol3] Fixed dummy rings

Brechtpd:
This works around the limitation in the new circuits that the fills cannot be 0. I forgot that this is actually a case where that would be useful.

Luckily  the workaround isn't very difficult. We can can just create orders for the dummy account that for example self-trades 1 wei LRC for 1 wei ETH. So the only difference for the relayer should be that the dummy orders have some different properties and that the dummy account needs e.g. 1 wei LRC and 1 wei ETH in its account (this only needs to be deposited once). The dummy ring can then be used, as is, `2**96` times, so there will be no need to ever create another dummy ring or another dummy account.

A minor downside is that the dummy ring now cost 24 gas more/dummy trade (but the compression could help with this), but this is pretty insignificant.

### DIFF
--- a/packages/loopring_v3/test/testExchangeBlockPermutations.ts
+++ b/packages/loopring_v3/test/testExchangeBlockPermutations.ts
@@ -147,6 +147,25 @@ contract("Exchange", (accounts: string[]) => {
       }
     });
 
+    it("Ring Settlement (dummy rings)", async () => {
+      const bDataAvailabilities = [true];
+      for (const bDataAvailability of bDataAvailabilities) {
+        await createExchange(bDataAvailability);
+        await exchangeTestUtil.commitDeposits(exchangeId);
+        const blockSizes = exchangeTestUtil.ringSettlementBlockSizes;
+        for (const blockSize of blockSizes) {
+          for (let i = 0; i < blockSize; i++) {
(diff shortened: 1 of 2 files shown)


dong77:
Hmmm... I really wished we have truely 0-padding

Brechtpd:
It wasn't possible before the order matching change because:
- `accountID == 0` is the protocol fee account, which can't sign any orders (and a valid order signature is always needed).
- It's invalid to trade between identical tokens, so at least one order needs `tokenIdS != 0`.

Both were (and are) not possible and makes it impossible for the on-chain data to be all zeros.

To make this possible we would have to reserve `accountID == 0` for this (with perhaps a publicly known public/private EdDSA key) or make it possible to put orders in a block without a valid signature. 

We would also have to allow orders with `tokenS == tokenB` (so both tokenIDs can be 0) and mark those trades as invalid and enforce the fills to be 0. Another possibility here would be to zero out the on-chain data if the ring is invalid, but we would still have to have different logic to detect these invalid things, instead of doing the `assert`-like logic we do now which is easier to do with constraints.

It's certainly an inconvenience, but I would not expect these dummy trades to be used a lot?

PR #691, opened 2019-11-14
[TokenBank] 2: implement a simple WalletFactory

dong77:


packages/tokenbank_v1/contracts/impl/WalletFactory.sol
```diff
@@ -0,0 +1,100 @@
...
+    function createWallet(
+        address   _owner,
+        address[] calldata _modules
+        )
+        external
+        payable
+        onlyManager
+        returns (address walletAddress)
+    {
+        SimpleProxy proxy = new SimpleProxy(walletImplementation);
```
Brechtpd: Will we allow wallets to be created with `create2`? So that, for example, we can generate wallet addresses for free for all users without paying any gas if the address remains unused?

PR #692, opened 2019-11-14
[TokenBank] 3: Add an unimplemented WalletFactoryWithENS

dong77:


packages/tokenbank_v1/contracts/impl/WalletFactoryWithENS.sol
```diff
@@ -0,0 +1,84 @@
...
+            extendedModules[0] = address(this);
+            for(uint i = 0; i < _modules.length; i++) {
+                extendedModules[i + 1] = _modules[i];
+            }
+            walletAddress = createWalletInternal(_owner, extendedModules);
+
+            Wallet wallet = Wallet(walletAddress);
+            registerSubdomain(wallet, _subdomain);
+
+            wallet.removeModule(address(this));
```
Brechtpd: `this` is added as a module to the wallet and then removed. I guess it will be necessary for the wallet to temporarily have `this` as a module?

PR #697, opened 2019-11-15
Update ring settlement test.

yueawang:
For issue https://github.com/Loopring/protocol3-circuits/issues/31

Update the test framework.
1. Use real trade for ring settlement block padding.
2. Make a test case cover padding case.

Test is running, review in parallel.


Brechtpd:
I made a similar PR #690 yesterday. :)

I think my very similar method has some advantages because there is no need to create new rings/orders every time (the same dummy ring can be reused `2**96` times) and the orderID of both orders is always 0 (which saves on onchain data but also means we don't need to worry about the limited number of trading history slots in an account).

yueawang:
Aha, yes, that's truly an advantage 👍

PR #711, opened 2019-11-18
[TokenBank] Added a BankRegistry

dong77:


packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -69,7 +77,7 @@ contract BaseWallet is Wallet, NamedAddressSet, ReentrancyGuard
         require(_modules.length > 0, "EMPTY_MODULES");
 
         owner = _owner;
-
+        bankRegistry.registerWallet(address(this));
```
Brechtpd: I think this will have to be moved to `WalletFactory` because `registerWallet` is `onlyFactory` (which makes sense).

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -53,10 +56,15 @@ contract BaseWallet is Wallet, NamedAddressSet, ReentrancyGuard
 
     modifier onlyModule
     {
-        require(isAddressInSet(MODULE, msg.sender), "NOT_A_MODULE");
+        require(isAddressInSet(MODULE, msg.sender), "MODULE_UNAUTHORIZED");
+        require(bankRegistry.isModuleRegistered(msg.sender), "INVALID_MODULE");
```
Brechtpd: Maybe slightly strange that a module cannot be removed from a wallet when it's deregistered. This may also make the wallet unusable because the wallet may not have any registered modules any more even if `numAddressesInSet(MODULE) > 1`.

packages/tokenbank_v1/contracts/impl/BaseBankRegistry.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    modifier onlyFactory()
+    {
+        require(msg.sender == factory, "UNAUTHORIZED");
+        _;
+    }
+
+    constructor() public Claimable() {}
+
+    function setWalletFacgtory(address _factory)
```
Brechtpd: Typo

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: You are right! Fixed

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
dong77: Sharp eyes. I changed it so only adding new module will check if the module is registered; all modules that are already registered are still valid even if they are all removed from the registry.

My concern is if a module is buggy, there is no global way of disabling it. Any suggestions?

packages/tokenbank_v1/contracts/impl/BaseBankRegistry.sol
dong77: solved.

Brechtpd:
@dong77 It looks like some of the fixes haven't been pushed to this branch yet?

packages/tokenbank_v1/contracts/impl/BaseWallet.sol
Brechtpd: Maybe there can be some kind of fail-safe method that allows some kind of default module to be added if all modules on a wallet are not registered any more which just restores the very basic workings of a wallet. Because any module can remove another module this could also be used to remove the buggy module completely from a wallet. Currently `removeModule` calls `terminate` directly on the module that is removed, but if needed a custom clean up contract could also be used (passed into `removeModule`) if even `terminate` is buggy, but that may take things to far.

Another idea that may work is that if the module is disabled some kind of replacement contract can be set which is then queried and is used to do all operations normally done on the original module contract. This replacement contract would just allow the module to be safely removed without breaking the wallet (but this leans very closely to just making the module upgradeable).

Not really sure how this is best solved, the ideas above are certainly not super easy or elegant.

dong77: I'll need time to think about it.

PR #716, opened 2019-11-19
[tokenbank] add UpgraderModule

dong77:


packages/tokenbank_v1/contracts/impl/BaseWallet.sol
```diff
@@ -160,6 +155,21 @@ contract BaseWallet is Wallet, NamedAddressSet, ReentrancyGuard
         emit StaticMethodBound(_method, _module);
     }
 
+    function supportsInterface(bytes4 _method)
```
Brechtpd: You're probably already aware of this, but because an interface can XOR multiple function selectors together it's possible a module implements a certain interface, but because this looks at every function individually it would still return false, even if all methods of an interface are implemented. A possible solution is that a module would also register it's supported interfaces in the wallet. Querying the modules every time would also be possible, but is a bit more expensive when `supportsInterface` is used a lot.

dong77: Brecht, I guess I didn't get your point...

Brechtpd: Example from the spec:
```
interface Simpson {
    function is2D() external returns (bool);
    function skinColor() external returns (string);
}

contract Homer is ERC165, Simpson {
    function supportsInterface(bytes4 interfaceID) external view returns (bool) {
        return
          interfaceID == this.supportsInterface.selector || // ERC165
          interfaceID == this.is2D.selector
                         ^ this.skinColor.selector; // Simpson
}
```

Let's say a module implements the `Simpson` interface. Using ERC165, someone may check if the `Simpson` interface is implemented using `supportsInterface(Simpson.is2D.selector ^ Simpson.skinColor.selector)`. This will however always return false with the current implementation in the wallet (even if the wallet has a module that implements the `Simpson` interface) because the `methodToModule[_method] != address(0))` check for functions in modules only supports interfaces with a single function (`supportsInterface(Simpson.is2D.selector) == true` and `supportsInterface(Simpson.skinColor.selector) == true`, but the actual `Simpson` interface will return false). Interfaces with multiple functions are not supported because the wallet has no way of knowing which functions together also define an interface.

dong77: I see. So maybe we can do something like this:
```
bytes4 id = interfaceID;
for (i = 0; i < methods; i++) {
   id &= ^mehods[i];
   if (id == 0) return true;
} 
return id == 0;
```

dong77: I guess it doesn't work.

Brechtpd: I think this would work if the methods of an interface are in the `methods` array are grouped together in the 
```
for (i = 0; i < methods; i++) {
  id = 0;
  for (j = i; j < methods; j++) {
   id ^= mehods[i];
   if (id == interfaceID) return true;
} 
return false;
```

But this still has O(n!) complexity. I think letting the module just manually specify this some way would be better.

dong77: I would argue that we treat each method as one single Interface.

Brechtpd: If we assume each module implements a complete interface (so an interface doesn't need multiple modules to be a complete implementation), which seems to be a very reasonable assumption, having a `supportsInterface` on the module as well seems like a pretty clean solution to me. It seems unlikely that `supportsInterface` will be called on-chain many times any way, so caching the supported interfaces in the wallet itself may not even make sense (though we would have to watch out for the 30,000 gas limit).

I think the problem with not supporting EIP165 correctly is that it may not make a lot of sense to implement this exact interface if it doesn't work as expected for 3rd parties. Maybe we can instead define the method as `supportsMethod` to make that clear.

dong77: I change the method name to supportsMethod and do not make the wallet compatible with EIP-165.

PR #719, opened 2019-11-20
[Protocol3] Add owner contracts with mandatory delays

Brechtpd:
On our current smart contracts there are some critical values that the owner of the contract could immediately change (like changing the verification keys so proofs can be constructed using any rules) which is a big security issues for users.

This PR has a simple base contract that enforces a mandatory delay on specified function calls. This means that if the owner of a contract is malicious, users can observe the malicious intent and act before the function call actually can be executed.

In the future this can be done with more advanced mechanisms like a DAO, but this seems to be a good compromise between security/simplicity in the short/medium term.

### DIFF
--- a/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
+++ b/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
@@ -27,7 +27,7 @@ contract BlockVerifierOwner is DelayedOwner
     constructor(
         IBlockVerifier blockVerifier
         )
-        DelayedOwner(address(blockVerifier))
+        DelayedOwner(address(blockVerifier), 3 days)
         public
     {
         setFunctionDelay(blockVerifier.transferOwnership.selector, 7 days);
--- a/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
+++ b/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
@@ -25,12 +25,11 @@ import "../../iface/IBlockVerifier.sol";
 contract BlockVerifierOwner is DelayedOwner
 {
     constructor(
-        address blockVerifierAddress
+        IBlockVerifier blockVerifier
         )
-        DelayedOwner(blockVerifierAddress)
+        DelayedOwner(address(blockVerifier))
         public
     {
-        IBlockVerifier blockVerifier = IBlockVerifier(blockVerifierAddress);
         setFunctionDelay(blockVerifier.transferOwnership.selector, 7 days);
         setFunctionDelay(blockVerifier.registerCircuit.selector, 7 days);
         setFunctionDelay(blockVerifier.disableCircuit.selector, 1 days);
--- a/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
+++ b/packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
@@ -1,0 +1,38 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+import "./DelayedOwner.sol";
+import "../../iface/IBlockVerifier.sol";
+
+
+/// @title  BlockVerifierOwner
+/// @author Brecht Devos - <brecht@loopring.org>
+contract BlockVerifierOwner is DelayedOwner
+{
+    constructor(
+        address blockVerifierAddress
+        )
+        DelayedOwner(blockVerifierAddress)
+        public
+    {
+        IBlockVerifier blockVerifier = IBlockVerifier(blockVerifierAddress);
+        setFunctionDelay(blockVerifier.transferOwnership.selector, 7 days);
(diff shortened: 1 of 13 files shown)


packages/loopring_v3/contracts/lib/BytesUtil.sol
```diff
@@ -65,6 +65,20 @@ library BytesUtil
         return uint16(bytesToUintX(b, offset, 2) & 0xFFFF);
     }
 
+    function bytesToBytes4(
```
dong77: I think we should make this function more gereric, like sliceBytes(offset, size)...

packages/loopring_v3/contracts/impl/owners/DelayedOwner.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    event DelayedTransactionExecuted(
+        uint    delayedTransactionIdx,
+        uint    timestamp,
+        uint    value,
+        bytes   data
+    );
+
+    address public contractAddress;
+    mapping (bytes4 => uint) public functionDelays;
```
dong77: I think we should make the  delay configurations ar queryable, something like `delayedFunctions()`

packages/loopring_v3/contracts/impl/owners/DelayedOwner.sol
```diff
@@ -0,0 +1,211 @@
...
+    {
+        (success, returnData) = contractAddress.call.value(transaction.value)(transaction.data);
+    }
+
+    function disableTransaction(
+        Transaction storage transaction
+        )
+        internal
+    {
+        transaction.timestamp = 0;
```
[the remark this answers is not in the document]
Brechtpd: Done both.

packages/loopring_v3/contracts/impl/owners/BlockVerifierOwner.sol
```diff
@@ -0,0 +1,38 @@
...
+/// @author Brecht Devos - <brecht@loopring.org>
+contract BlockVerifierOwner is DelayedOwner
+{
+    constructor(
+        address blockVerifierAddress
+        )
+        DelayedOwner(blockVerifierAddress)
+        public
+    {
+        IBlockVerifier blockVerifier = IBlockVerifier(blockVerifierAddress);
```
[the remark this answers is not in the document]
Brechtpd: Done

packages/loopring_v3/contracts/lib/BytesUtil.sol
Brechtpd: The reason for the specific `bytesToBytes4` is that otherwise we would always have to cast to `bytes4` after calling a more general function. I modified it to make use of a more general function internally (similar to the uint functions).

packages/loopring_v3/contracts/impl/owners/DelayedOwner.sol
Brechtpd: Done

packages/loopring_v3/contracts/impl/owners/DelayedOwner.sol
```diff
@@ -0,0 +1,341 @@
...
+    constructor(
+        address _ownedContract
+        )
+        public
+    {
+        require(_ownedContract != address(0), "ZERO_ADDRESS");
+        ownedContract = _ownedContract;
+    }
+
+    function executeDelayedTransaction(
```
dong77: executeDelayedTransaction => executeDelayed ?
Make some methods shorter would be nice.

dong77:
I believe the point of DalayedOwner is to express the intentions first then invoke the methods. Maybe we should restrict each deployable function has at most one delayed tx, otherwise, the user will submit multiple transactions for one method and then decide which one to actually invoke, which introduces confusing and uncertainty. In other words, when a delyedFunction is called by the user, it should replace the old pending tx if there is any.

The other improvement would be giving all delayed method the same TTL so the method must be called within that TTL or otherwise it' will expire.

Brechtpd:
> I believe the point of DalayedOwner is to express the intentions first then invoke the methods. Maybe we should restrict each deployable function has at most one delayed tx, otherwise, the user will submit multiple transactions for one method and then decide which one to actually invoke, which introduces confusing and uncertainty. In other words, when a delyedFunction is called by the user, it should replace the old pending tx if there is any.

I believe this would be too strict because calling the same function multiple times can be valid. If we create a new circuit and need to register multiple new circuits then these would all need to call the same function. Of course we could instead create a batch function which does multiple things at once, but this would need additional logic and it would be hard to make a system that is completely transparent even if we do it like that (see below).

The most robust method I think would be to keep a strict ordering:
- Disallow cancellations
- Force all delayed transactions to be done in the order they were originally submitted (which could extend the delay of some function calls if they come after a function call with a longer delay), otherwise the resulting state could still be different.

But the reason for the delays is that malicious intent can be detected by users. If the owner wants to submit multiple transactions to confuse users then that would be the same as a single transaction that is malicious. In any scenario (no matter the specifics) the owner needs to make sure his transactions cannot be seen as malicious (and this could mean following additional rules not enforced on-chain). The on-chain part is just to make it public, the reaction needs to be taken off-chain by the users. So I think the extra flexibility given by the current contract to the owner is fine and does not worsen the security provided.

> The other improvement would be giving all delayed method the same TTL so the method must be called within that TTL or otherwise it' will expire.

That's a good idea!

dong77:
After a second thought I think allowing executing of pending transactions of the same type out of order should also be fine as long as a ttl is applied. The order of different functions cannot be enforced at all as they have different delay settings.

kongliangzhong:
Maybe we should move contracts Ownable, Claimable, Authorizable, and Killable from lib to owners directory.

Brechtpd:
> Maybe we should move contracts Ownable, Claimable, Authorizable, and Killable from lib to owners directory.

Possibly. But the contracts mentioned above are actual core protocol contracts. The contracts currently in the owners folder are more like non-core utility contracts that can be used with the protocol, but are not really a part of it. So maybe the contracts in owners should be moved out to a 'tools' or 'utitilies' folder, and all core exchange contracts can be moved to an 'protocol' or 'exchange' folder to make this more clear. But that's a lot of files moving around so perhaps this is something that can be done for version 3.1. :)

PR #722, opened 2019-11-20
[tokenbank] fix  MetaTxModule's design

dong77:


dong77:
The previous implementations requires a bind(wallet, guardian) function just for the relayed meta transaction. Now the signer is passed to executeSigned method as a parameter but the data signed by the signer doesn’t need to include the guardian param.

dong77:
@letsgoustc please let me know if you have any feedback. Will merge only after your review.

letsgoustc:
> @letsgoustc please let me know if you have any feedback. Will merge only after your review.

The argent's origin design only allow wallet owner to sign the meta transaction. Looks like we add one more signer variable to allow non-wallet owner?

dong77:
The guardian can also sign meta-transactions.
Who can sign meta-transactions is totally up to the module, the point is: the singer can be any arbitrary contract address, not just external owned addresses.

letsgoustc:
> The guardian can also sign meta-transactions.
> Who can sign meta-transactions is totally up to the module, the point is: the singer can be any arbitrary contract address, not just external owned addresses.

Got it. Agreed that our current design is more flexible to let the module to decide who can sign metaTx.

PR #732, opened 2019-11-23
[tokenbank] implement security related modules (#731)

dong77:


packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -0,0 +1,155 @@
...
+        require(
+            method == this.startRecovery.selector ||
+            method == this.completeRecovery.selector ||
+            method == this.cancelRecovery.selector,
+            "INVALID_METHOD"
+        );
+
+        // data layout: {length:32}{sig:4}{_wallet:32}{signer1:32}{signer2:32}{...}
+        require((data.length - 68) % 32 == 0, "INVALID_DATA");
+        uint numSigners = (data.length - 68) / 32;
```
Brechtpd: I think this should be `(data.length - 36)` because the 32 bytes length value isn't part of `data.length`.

packages/tokenbank_v1/contracts/modules/security/SecurityStorage.sol
```diff
@@ -0,0 +1,126 @@
...
+{
+    struct Wallet
+    {
+        uint128   lock;
+        address   locker; // the module locked/unlocked this wallet
+        address[] guardians;
+        mapping   (address => uint) guardianIdx;
+    }
+
+    mapping (address => Wallet) public cells;
```
Brechtpd: Why is this called `cells` and not simply `wallets` if the struct is named `Wallet`?

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -0,0 +1,155 @@
...
+        delete wallets[wallet];
+        securityStorage.setLock(wallet, 0);
+
+        emit RecoveryCancelled(wallet);
+    }
+
+    function completeRecovery(address wallet)
+        external
+        nonReentrant
+        onlyFromMetaTx
```
Brechtpd: Does it need to be `onlyFromMetaTx`?

packages/tokenbank_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -0,0 +1,184 @@
...
+        pendingPeriod = _pendingPeriod;
+        confirmPeriod = _confirmPeriod;
+    }
+
+    function addGuardian(
+        address wallet,
+        address guardian
+        )
+        external
+        nonReentrant
```
Brechtpd: Because `executeMetaTx` is `nonReentrant`, I guess all functions called by a metaTx can't be `nonReentrant`? Not really sure how to best fix this. Maybe if `msg.sender == address(this)` the reentrancy check is skipped?

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
dong77: The current design does not maintain the signature collection process in smart contract, the server/relayer will coordinate the collection of signatures, once enough signatures are collected, it will invoke the `startRecovery` method. But the transactor doesn't have to be a specific address, anyone including any guardian who has enough signatures can transact the meta-transaction.

packages/tokenbank_v1/contracts/modules/security/SecurityStorage.sol
dong77: fixed. The struct used to call WalletCell

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
dong77: // data layout: {length:32}{sig:4}{_wallet:32}{signer1:32}{signer2:32}{...}

The first signer address is preceded with 32 bytes length, 4 bytes method selector, and 32 bytes wallet address, so it has 68 bytes ahead of it.

dong77: But I think there is a bug, I'll fix it and please review this part again.

packages/tokenbank_v1/contracts/modules/security/GuardianModule.sol
dong77: I added  `nonReentrantExceptFrom` and `nonReentrantExceptFromThis`

packages/tokenbank_v1/contracts/base/MetaTxModule.sol
```diff
@@ -241,4 +254,24 @@ contract MetaTxModule is BaseModule
             wallets[wallet].nonce = nonce;
         }
     }
+
+    // TODO (daniel): return false in case of ERC1271 error, not throw exception
```
dong77: @Brechtpd can you help me do this? I'm not familiar with this technique. 

> TODO (daniel): return false in case of ERC1271 error (for example the contract doesn't support 1271 at all), not throw an exception.

Brechtpd: I don't there's anything really new to do this. This should work:

```
bytes memory callData = abi.encodeWithSelector(
    ERC1271(signer).isValidSignature.selector,
    hash,
    extractSignature(signatures, idx)
);
(bool success, bytes memory data) = signer.staticcall(callData);
if (success && data.length == 32) {
    return (data.bytesToBytes4(0) == ERC1271_MAGICVALUE);
} else {
    return false;
}
```

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -0,0 +1,164 @@
...
+        signers = new address[](numSigners);
+
+        address signer;
+        for (uint i = 0; i < numSigners; i++) {
+            uint start = skipBytes + 32 * i;
+            assembly {signer := mload(add(data, start)) }
+            signers[i] = signer;
+        }
+    }
+}
```
Brechtpd: The `length` is the array length, right? Then, for example, with data like this:
```
{length:32}{sig:4}{_wallet:32}{_newOwner:32}{signer1:32}
```
`data.length == 100` and **not** `data.length == 132`, because the number of bytes used by `length` at the start is just some internal data for the array. But for reading the data using assembly you do need to use a correct offset that takes those `length` bytes into account.

Probably a good idea to add a utility method in `BytesUtil.sol` to read addresses stored in 32 bytes (there's already a 20 bytes one, but the current function names are pretty bad, so some renaming is a good idea anyway):
```
// Can be used to extract an address packed in 20 bytes
function readPackedAddress(
        bytes memory b,
        uint  offset
        )
        internal
        pure
        returns (address)
{
    return address(bytesToUintX(b, offset, 20) & 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF);
}

// Can be used to extract an address packed in 32 bytes
function readAddress(
        bytes memory b,
        uint  offset
        )
        internal
        pure
        returns (address)
{
    return address(bytesToUintX(b, offset, 32) & 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF);
}
```

This way we don't have to think about it:
```
function extractMetaTxSigners(
    address /*wallet*/,
    bytes4  method,
    bytes   memory data
    )
    internal
    view
    returns (address[] memory signers)
{
    uint skipBytes = 0;
    if (method == this.startRecovery.selector) {
        // data layout: {sig:4}{_wallet:32}{_newOwner:32}{signer1:32}{signer2:32}{...}
        skipBytes = 68;
    } else if (method == this.cancelRecovery.selector) {
        // data layout: {sig:4}{_wallet:32}{signer1:32}{signer2:32}{...}
        skipBytes = 36;
    } else if (method == this.completeRecovery.selector){
        // data layout: {sig:4}{_wallet:32}
        skipBytes = 36;
        require(data.length == skipBytes, "INVALID_DATA");
    } else {
        revert("UNKNOWN_DATA");
    }

    uint numSigners = (data.length.sub(skipBytes)) / 32;
    signers = new address[](numSigners);
    for (uint i = 0; i < numSigners; i++) {
        signers[i] = data.readAddress(skipBytes + 32 * i);
    }
}
```

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
Brechtpd: But `completeRecovery` doesn't need any signatures, no? I was just curious because it seems like anyone can complete the recovery now, so it seems like it doesn't matter if `completeRecovery` is called with a MetaTx or a direct call.

dong77: You are right, I should remove onlyFromMetaTx so this method can be called by anybody.

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
dong77: Actually I like that we know all the details :)

packages/tokenbank_v1/contracts/base/MetaTxModule.sol
dong77: done

packages/tokenbank_v1/contracts/modules/security/GuardianModule.sol
Brechtpd: Hmmm seems like there are some subtleties with this solution. With `nonReentrantExceptFrom` the `_guardValue` is set to `0` when exiting the function which would reset is, which means that afterwards calling a method like that you could actually do reentrancy again. So a function stack like this would be possible:
```
nonReentrant (_guardValue -> 1)
  nonReentrantExceptFrom (_guardValue -> 0 after function end)
  nonReentrant (_guardValue -> 1)
```

With some extra logic this could perhaps be solved, but it seems very error prone because it can be very subtle how things can go wrong.

I think it would probably be easier to just keep the normal `nonReentrant` everywhere (easy to understand), while making  `executeMetaTx` **not** `nonReentrant`. Instead,  `executeMetaTx` needs to follow the Checks-Effects-Interactions pattern that makes it inherently safe against reentrancy. I think that is currently already the case (`saveExecutedMetaTx` is done before the function call).

dong77: Yeah, the previous implementation does have the bug. Fixing the modifiers is trivial, please let me know if you like the fix or not.

Brechtpd: I think it works correctly now, but functions now need to be a little bit aware of these more complex interactions which does make things more difficult and error-prone. I think just making ` executeMetaTx` not `nonReentrant` would be easier, but the current method with modifiers seems to be work fine as well.

dong77: executeMetaTx should be nonReentrant. A couple of legit call stacks:

- `executeMetaTx` (in `moduleA`) calls` foo` (in `ModuleA`) which calls `bar` (in a `Wallet`) which calls `xyz` (in `moduleA`).


- `executeMetaTx` (in `moduleA`) calls` foo` (in `ModuleA`) which calls `bar` (in a `Wallet`) which calls `xyz` (in `moduleB`).

Brechtpd: `executeMetaTx` should be safe for reentrancy, but using `nonReentrant` is just one way to do it. Using the Checks-Effects-Interactions pattern is another way to do. And the benefit with the pattern method is that we don't modify the state of the contract, so we don't influence any other methods.

> * `executeMetaTx` (in `moduleA`) calls` foo` (in `ModuleA`) which calls `bar` (in a `Wallet`) which calls `xyz` (in `moduleA`).

I don't really see how a state-modifying method can be called back on a module (`xyz` in this example) that isn't a `view` method (view methods don't need to be `nonReentrant` so are irrelevant). I guess it's possible by using `transact`, but the `msg.sender` for these is the wallet, so contracts that are called can't reach the original module (unless it's a view method of course, which doesn't matter). And even if it would be possible to do it, I think `moduleA.foo` and `moduleA.xyz` wouldn't be safe against reentrancy even when using a variant of `nonReentrant` unless they both follow the Checks-Effects-Interactions pattern because they can be used for reentrancy by external contracts via the wallet and that could be dangerous.

dong77: @Brechtpd please feel free to submit a PR for this :)

Brechtpd: Will do. :)

dong77:
@Brechtpd I'll merge this one for now :)

PR #738, opened 2019-11-24
[tokenbank] implement transfer-related modules

dong77:


packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -144,30 +149,26 @@ contract RecoveryModule is SecurityModule
...
+        require (
+            method == this.startRecovery.selector ||
+            method == this.cancelRecovery.selector ||
+            method == this.completeRecovery.selector,
+            "INVALID_METHOD"
         );
 
-        uint numSigners = (data.length - skipBytes) / 32;
+        // ASSUMPTION:
+        // data layout: {data_length:32}{wallet:32}{signers_length:32}{signer1:32}{signer2:32}
```
Brechtpd: Arrays are ABI encoded slightly differently with an offset to the data:
- https://solidity.readthedocs.io/en/v0.5.13/abi-spec.html#examples

So I think a call to e.g. `completeRecovery` needs data like this:
```
{wallet:32}{signers_offset:32}{signers_length:32}{signer1:32}{signer2:32}
```

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -16,11 +16,259 @@
...
+            return;
+        }
+
+        uint allowance = ERC20(token).allowance(wallet, to);
+        if (allowance >= amount) {
+            approveInternal(wallet, token, to, amount);
+            return;
+        }
+
+        allowance -= amount;
```
Brechtpd: I think this needs to be `amount -= allowance;` to have the delta increased value to use in the following calculation.

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -16,11 +16,259 @@
...
+        uint valueInCNY = priceProvider.getValueInCNY(token, allowance);
+
+        if (quotaStore.checkAndAddToSpent(wallet, valueInCNY)) {
+            approveInternal(wallet, token, to, allowance);
```
Brechtpd: Needs to be `approveInternal(wallet, token, to, amount);`?

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -16,11 +16,259 @@
...
+        address wallet,
+        bytes32 pendingTxId
+        )
+        private
+        returns (bool foundPendingTx)
+    {
+        if (msg.sender != Wallet(wallet).owner()) {
+            if (isPendingTxValid(wallet, pendingTxId)) {
+                foundPendingTx = true;
+                emit PendingTxExecuted(wallet, pendingTxId, now);
```
Brechtpd: Pending transactions aren't marked as executed, so it's valid that pending transactions can be executed an unlimited amount of times within `pendingExpiry` seconds of creating the pending transaction?

packages/tokenbank_v1/contracts/modules/security/RecoveryModule.sol
dong77: This is really helpful. Please take another look.

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
dong77: it is `allowance` but it needs to be updated.

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
dong77: fixed.

packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
```diff
@@ -0,0 +1,70 @@
...
+        internal
+        view
+        returns (address[] memory signers)
+    {
+        require (
+            method == this.createDEXAccount.selector,
+            "INVALID_METHOD"
+        );
+        // ASSUMPTION:
+        // data layout: {data_length:32}{wallet:32}{signers_offset:32}{signers_length:32}{signer1:32}{signer2:32}
```
Brechtpd: Looks correct, but the function selector is still missing? I thought it was stripped off some place, but that doesn't seem to be the case.

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -16,11 +16,257 @@
...
+        }
+
+        uint allowance = ERC20(token).allowance(wallet, to);
+        if (allowance >= amount) {
+            approveInternal(wallet, token, to, amount);
+            callContractInternal(wallet, to, 0, data);
+            return;
+        }
+
+        allowance -= amount;
```
Brechtpd: Still needs to be changed to `allowance = amount - allowance;` like above.

packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
dong77: sorry about that 👍

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
dong77: fixed

PR #740, opened 2019-11-25
[tokenbank] add price oracles

dong77:


packages/tokenbank_v1/contracts/modules/stores/PriceCacheStore.sol
```diff
@@ -0,0 +1,76 @@
...
+        uint            amount,
+        uint            value,
+        uint            timestamp
+    );
+
+    struct TokenPrice
+    {
+        uint amount;
+        uint value;
+        uint timestmap;
```
Brechtpd: typo

packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -258,6 +262,20 @@ contract QuotaTransfers is TransferModule
...
     }
 
+    function getTokenValue(address token, uint amount)
+        private
+        returns (uint value)
+    {
+        if (amount == 0) return 0;
+        value = priceCacheStore.tokenPrice(token, amount);
+        if (value == 0) {
+            value = priceOracle.tokenPrice(token, amount);
```
Brechtpd: I don't think this is much of a problem probably, but small `amount` values can have significant rounding errors compared to the real price. Maybe `decimals` on a token could be used for some standard amount so this can't be the case, not sure.

dong77: I think it should be fine. Using 'decimals' will probably complicate things...

packages/tokenbank_v1/contracts/modules/stores/PriceCacheStore.sol
dong77: fixed

PR #743, opened 2019-11-25
[tokenbank] implement ExchangeModule

dong77:


packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
```diff
@@ -0,0 +1,247 @@
...
+    function getDEXAccount(IExchangeV3 exchange)
+        public
+        view
+        returns (
+            uint24 accountId,
+            uint   pubKeyX,
+            uint   pubKeyY
+        )
+    {
+        return exchange.getAccount(address(this));
```
Brechtpd: Should use the wallet's address.

packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
```diff
@@ -0,0 +1,247 @@
...
+    function checkAndReimburse(
+        address payable wallet,
+        uint            etherFee,
+        address         feeToken,
+        uint            feeAmount
+        )
+        private
+    {
+        if (wallet.balance >= etherFee) {
+            msg.sender.transfer(msg.value);
```
Brechtpd: `executeMetaTx` will have to be payable so the ETH can be received in the module.

Brechtpd: I think the ETH will also have to be resent in the `call` so that `msg.value` is the correct value. I think it will be `0` now because the actual `msg.value` is lost in `executeMetaTx` when doing a new `call`.

packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
```diff
@@ -0,0 +1,247 @@
...
+        address         feeToken,
+        uint            feeAmount
+        )
+        external
+        nonReentrantExceptFromThis
+        onlyFromMetaTxOrWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+    {
+        (uint creationFee, uint updateFee, , ) = exchange.getFees();
+        (uint24 accountId, , ) = exchange.getAccount(wallet);
```
Brechtpd: This will actual throw when the account doesn't exist. Maybe we should change that so it's easier to use from smart contracts.

packages/tokenbank_v1/contracts/modules/loopring/ExchangeModule.sol
```diff
@@ -0,0 +1,247 @@
...
+
+        uint fee = accountId == 0 ? creationFee : updateFee;
```
Brechtpd: A deposit fee is always charged as well:
```
uint fee = depositFee + (accountId == 0 ? creationFee : updateFee);
```

PR #750, opened 2019-11-25
[tokenbank] implement LoopringDEXModule

dong77:


packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
```diff
@@ -0,0 +1,247 @@
...
+    function getDEXAccount(IExchangeV3 exchange)
+        public
+        view
+        returns (
+            uint24 accountId,
+            uint   pubKeyX,
+            uint   pubKeyY
+        )
+    {
+        return exchange.getAccount(address(this));
```
Brechtpd: Should use the wallet's address.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
```diff
@@ -0,0 +1,247 @@
...
+    function checkAndReimburse(
+        address payable wallet,
+        uint            etherFee,
+        address         feeToken,
+        uint            feeAmount
+        )
+        private
+    {
+        if (wallet.balance >= etherFee) {
+            msg.sender.transfer(msg.value);
```
Brechtpd: `executeMetaTx` will have to be payable so the ETH can be received in the module.

I think the ETH will also have to be resent in the `call` so that `msg.value` is the correct value. I think it will be `0` now because the actual `msg.value` is lost in `executeMetaTx` when doing a new `call`.

packages/tokenbank_v1/contracts/modules/transfers/TokenBalances.sol
```diff
@@ -35,22 +35,20 @@ contract TokenBalances is BaseModule
...
         public
         view
         returns (uint)
     {
         if (token == address(0)) {
-            return wallet.balance;
+            return address(this).balance;
         } else {
-            return ERC20(token).balanceOf(wallet);
+            return ERC20(token).balanceOf(address(this));
```
dong77: @Brechtpd , my understanding is that since this method is bound to a wallet, so the "address(this)" should be the wallet.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
```diff
@@ -0,0 +1,247 @@
...
+        address         feeToken,
+        uint            feeAmount
+        )
+        external
+        nonReentrant
+        onlyFromMetaTxOrWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+    {
+        (uint creationFee, uint updateFee, , ) = exchange.getFees();
+        (uint24 accountId, , ) = exchange.getAccount(wallet);
```
Brechtpd: This will actual throw when the account doesn't exist. Maybe we should change that so it's easier to use from smart contracts.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
```diff
@@ -0,0 +1,247 @@
...
+
+        uint fee = accountId == 0 ? creationFee : updateFee;
```
Brechtpd: A deposit fee is always charged as well:
```
uint fee = depositFee + (accountId == 0 ? creationFee : updateFee);
```

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
dong77: change the executeMetaTx method.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
dong77: or can we use assembly and check the result?

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
dong77: the method is bound to the wallet, so I think address(this) will be the wallet's address.

packages/tokenbank_v1/contracts/modules/transfers/TokenBalances.sol
Brechtpd: `address(this)` is the address of the `TokenBalances` module. `transact` uses `call` and not `delegatecall` so there is nothing special with how the functions are called even through the wallet.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
dong77: even if the user is not changing the trading pub-key?

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
Brechtpd: Yes, that's also possible, but a little bit more complicated than needed. But now that solidity has support for it maybe it's not that bad actually.

packages/tokenbank_v1/contracts/modules/exchanges/LoopringDEXModule.sol
Brechtpd: If the pub-key remains unchanged no `updateFee` is charged, but `depositFee` is always charged, no exceptions (`depositFee` is charged for the cost of including the request in a block, which always needs to be done).

packages/tokenbank_v1/contracts/modules/transfers/TokenBalances.sol
dong77: then several static methods will have to be updated.

PR #754, opened 2019-11-25
[tokenbank] enable batch token transfer with all token balances

dong77:


packages/tokenbank_v1/contracts/modules/transfers/QuotaTransfers.sol
```diff
@@ -112,6 +112,29 @@ contract QuotaTransfers is TransferModule
...
+        bytes calldata     logdata
+        )
+        external
+        nonReentrant
+        onlyFromWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+    {
+        bool allowed = whitelistStore.isWhitelisted(wallet, to);
+
+        if (allowed) {
```
Brechtpd: Maybe this shouldn't silently do nothing, and just throw instead?

dong77: made it throw.

PR #755, opened 2019-11-25
[tokenbank] implement time-based inheritance.

dong77:


packages/tokenbank_v1/contracts/modules/stores/SecurityStore.sol
```diff
@@ -120,4 +122,31 @@ contract SecurityStore is DataStore
         wallets[wallet].lock = _lock;
         wallets[wallet].locker = msg.sender;
     }
+
+    function touchLastActive(address wallet)
+        public
+        onlyOwner
```
Brechtpd: `onlyManager`

packages/tokenbank_v1/contracts/modules/stores/SecurityStore.sol
```diff
@@ -120,4 +122,31 @@ contract SecurityStore is DataStore
...
+            uint    lastActive
+        )
+    {
+        who = wallets[wallet].inheritor;
+        lastActive = uint(wallets[wallet].lastActive);
+    }
+
+    function setInheritor(address wallet, address who)
+        public
+        onlyOwner
```
Brechtpd: `onlyManager`

packages/tokenbank_v1/contracts/modules/security/SecurityModule.sol
```diff
@@ -38,6 +38,26 @@ contract SecurityModule is MetaTxModule
         securityStore = _securityStore;
     }
 
+    // overriding
+    modifier onlyFromWalletOwner(address wallet) {
+        require(Wallet(wallet).owner() == msg.sender, "NOT_FROM_WALLET_OWNER");
+        _;
+        securityStore.touchLastActive(wallet);
```
Brechtpd: Any reason why only in SecurityModules? Or is it just because `SecurityModule` is basically the base class of all modules anyway?

dong77: SecurityModule has access to SecurityStorage, so it knows if the wallet is locked, who are the guardians.  SecurityModule also supports meta-tx. So basically most modules should inherit from SecurityModule.

PR #762, opened 2019-11-27
[tokenbank] implement a multi-signature vault.

dong77:


packages/tokenbank_v1/contracts/base/BaseVault.sol
```diff
@@ -0,0 +1,185 @@
...
+        require(
+            newRequirement > 0 &&
+            newRequirement <= numOwners(),
+            "INVALID_REQUIREMENT"
+        );
+        _requirement = newRequirement;
+        emit RequirementChanged(_requirement);
+    }
+
+    function requirement()
```
Brechtpd: `_requirement` -> `requirement` and solidity will automatically generate a `view`method for it if it's public (which `_requirement` is). Or is there another reason to do it like this?

dong77: no it's not working as `requirement()` is an interface method.

PR #774, opened 2019-11-29
[hebao] add missing logics for managing quota and whitelist

dong77:


packages/hebao_v1/contracts/modules/security/WhitelistModule.sol
```diff
@@ -0,0 +1,113 @@
...
+    function addToWhitelist(
+        address wallet,
+        address addr
+        )
+        external
+        nonReentrant
+        onlyFromMetaTxOrWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+    {
+        whitelistStore.addToWhitelist(wallet, addr);
```
Brechtpd: I would think this would need a time delay or `numGuardiants + 1 / 2` signatures to be safe.

packages/hebao_v1/contracts/modules/security/WhitelistModule.sol
```diff
@@ -0,0 +1,113 @@
...
+
+import "./SecurityModule.sol";
+
+
+/// @title WhitelistModule
+/// @dev Manages whitelisted addresses.
+contract WhitelistModule is SecurityModule
+{
+    WhitelistStore  public whitelistStore;
+    SecurityStore   public securityStore;
```
Brechtpd: `securityStore` is already defined in SecurityModule. Having a variable here with the same name is actually a new variable, unrelated to the one in the super class. This means `securityStore` is undefined in WhitelistModule. To make `securityStore` public in WhitelistModule a getter function needs to be used .

```
contract SuperContract
{
    uint internal a;
    
    constructor(uint _a) public {
        a = _a;
    }
    
    function getA() external view returns (uint) {
        return a;
    }
}

contract BaseContract is SuperContract
{
    uint public a;
    
    constructor() public 
        SuperContract(20)
    {
        a = 10;
    }
}
```

`BaseContract.a() == 10`
`BaseContract.getA() == 20`

packages/hebao_v1/contracts/modules/security/WhitelistModule.sol
dong77: I'll create an issue to apply a delay period.

packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -30,7 +30,6 @@ import "./SecurityModule.sol";
 /// @title RecoveryModule
 contract RecoveryModule is SecurityModule
 {
-
```
letsgoustc: In extractMetaTxSigners function.

require(data.length >= (100 + 32 * numSigners), "DATA_INVALID");

looks like should be: "require(data.length >= (132 + 32 * numSigners), "DATA_INVALID");"

Brechtpd: When using `.length` you don't count `data_length` as part of the array length, that's just how arrays store their length internally (so when you just use solidity you don't need to know this). It is important to keep `data_length` in mind when using assembly however.

Issue #780, opened 2019-11-30
[hebao] support Multihash so more types of signatures can be supported


letsgoustc:
https://ethresear.ch/t/pragmatic-signature-aggregation-with-bls/2105
https://medium.com/cryptoadvance/bls-signatures-better-than-schnorr-5a7fe30ea716
https://electriccoin.co/blog/new-snark-curve/

As eth2 will use BLS instead of ECDSA for verify multiple signature, we can follow to use it in our multiple signature implementation.

Brechtpd:
We should certainly make sure we can support multiple signature types (EIP-712 being the most important one right now I think).

Supporting BLS right now isn't really important I feel, maybe in 1-2 years when ETH2 is actually being used for smart contracts or when verifying BLS signatures gets substantially cheaper on ETH 1.X. BLS signatures are currently expensive to check on-chain (I don't have a good estimate, [this still has issues](https://github.com/kfichter/solidity-bls) and [SKALE (BLS-rollup) hasn't open sourced their smart contracts yet](https://github.com/skalenetwork/skale-network/blob/master/OPEN_SOURCE_PLAN.md) which would contain a good implementation. A rough estimate is that verifying a BLS signature costs 100,000 - 200,000 gas on Ethereum right now (which could do a bunch of signature verifications at once if the same data is hashed, which would be the case for us). Compared to ECDSA which costs 4,000 gas/signature, which has more flexibility for different weights/address and is supported in everything (hardware wallets/metamask/...). In most cases we'll also just check just a single signature because multiple signatures are only used for special operations which should be pretty rare.

dong77:
Supporting more types of signing algorithms should be important, so the ABI should be designed to cope with other signing algorithms. Maybe we should still use [Multihash](https://github.com/Loopring/protocols/blob/master/packages/loopring_v2/contracts/lib/MultihashUtil.sol)  which supports EIP-712.

For multiple signature support (one tx with multiple signatures), the current ABI should be fine, such as this one:

```solidity
metaTxHash.verifySignatures(signers, signatures);
```

The `signatures` is a `bytes` so it should be *extensible*.

The current implementation of the `verifySignatures` method does have an issue. It assumes all signatures are 65 bytes and are ECDSA signatures. We should convert it to support Multihash and multiple signatures.

dong77:
@Brechtpd @hosschao please take a look at https://github.com/Loopring/protocols/pull/782

PR #782, opened 2019-11-30
[hebao] support multiple types of signatures.

dong77:
Try to solve issue https://github.com/Loopring/protocols/issues/780

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseVault.sol
+++ b/packages/hebao_v1/contracts/base/BaseVault.sol
@@ -15,6 +15,7 @@
   limitations under the License.
 */
 pragma solidity ^0.5.11;
+pragma experimental ABIEncoderV2;
 
 import "../lib/AddressSet.sol";
 import "../lib/SignatureUtil.sol";
@@ -74,7 +75,7 @@ contract BaseVault is Vault, AddressSet
         uint      value,
         bytes     calldata data,
         address[] calldata signers,
-        bytes     calldata signatures
+        bytes[]   calldata signatures
         )
         external
     {
--- a/packages/hebao_v1/contracts/base/ERC1271Module.sol
+++ b/packages/hebao_v1/contracts/base/ERC1271Module.sol
@@ -79,6 +79,6 @@ contract ERC1271Module is BaseModule, ERC1271
             abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(_data))
         );
 
-        return signHash.recoverSigner(_signature) == _address;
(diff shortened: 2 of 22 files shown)


Brechtpd:
There's a couple of ways to do this and this is certainly a good way to do it. Some other options:

1. [EIP-2126](https://github.com/ethereum/EIPs/blob/202d578acb76bb4b8d0f46630eff4965ca61c092/EIPS/eip-2126.md): There is a sort of standard (still a draft) to support multiple signature schemes. The signature type is simply added to the end of the `bytes` array which simplifies the code a bit (no need to extract data). It doesn't support multiple signatures in a single `bytes`, but we don't need that if we simply do `bytes[] memory signatures` which is supported with `ABIEncoderV2`.
2. Do something similar to [gnosis](https://github.com/gnosis/safe-contracts/blob/78494bcdbc61b3db52308a25f0556c42cf656ab1/contracts/GnosisSafe.sol#L181). All signatures are still 65 bytes so we still have O(1) access to any signature (and you can still do `signatures.length/65` to make sure enough data can be read from the array). If a signature type needs more data it can be added to the end of the `bytes` array.

I kind of like option 1. because it keeps things very simply (no need to extract signatures from a larger `bytes` array, easy O(1) access), but all options are very similar of course so the method used in this PR is also fine with me (although it is a bit more complex and doesn't follow any standards (well, a draft standard...)).

dong77:
Assuming each signature is 65 bytes is really a bad idea. I also like the first approach. Actually I had tried to create a PR to use `bytes[] memory signature`, but then I did some research and found that there are even [bugs](https://blog.ethereum.org/2019/03/26/solidity-optimizer-and-abiencoderv2-bug/) in ABIEncoderV2 which scares me from using it. `bytes[] memory signature` will make the interface much cleaner.

Brechtpd:
I'm aware of the bugs with `ABIEncoderV2`, but because we don't ever store anything in the contract using `ABIEncoderV2` (which seems to be the case for these hard to detect `ABIEncoderV2` bugs as far as I'm aware) I think it's safe to use for things like this (if it's in memory and the data isn't correct it would be easy to detect while testing). There's lots of projects using `ABIEncoderV2` for a long time already (0x was already using it in V2 of their protocol more than a year ago for example).

dong77:
I changed the code to use ABIEncoderV2 and supported https://github.com/ethereum/EIPs/blob/202d578acb76bb4b8d0f46630eff4965ca61c092/EIPS/eip-2126.md

PR #788, opened 2019-11-30
[hebao] Changes to module activation/deactivation

Brechtpd:
I'm still not completely happy with this PR (adding/removing modules is less clean than before), but I do think the positives far outweigh the negatives.

With the current code the module can call into the module for activation/deactivation, which kind of complicates things a bit. In the best case, which makes things easier, everything is callable using `transact` or the fallback function without any limitations or caveats. The current exceptions are `activation` and `deactivation` which should not be callable by the wallet except when adding/removing modules. 

To solve this in this PR the wallet doesn't call `activate` and `deactivate` any more, those can now be called by the module adding/removing the module. This also means that the wallet itself has zero dependencies on modules. A module is now just an address for the wallet. The interface of a module is now just for other modules to add/remove the wallet, which could in theory change.

This also allows us to safely forward any call in the fallback function (not just staticcalls). If a module wants to catch a call that has passed through the wallet it can simply bind the method and check that `msg.sender` is the wallet address. This is different from before where we had to be careful not to call the wrong function from the wallet address. Using this PR any call to any contract can be made from the wallet.

This can be helpful for certain token standards (see [EIP-1155](`https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1155.md) which calls `onERC1155Received` on the smart contract). By having a `call` instead of a `staticcall` we can now more than before when catching these (like logging an event, which is not possible with `staticcall`). We could even append the original `msg.sender` so the module can check the origin of the call into the fallback function.

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseWallet.sol
+++ b/packages/hebao_v1/contracts/base/BaseWallet.sol
@@ -45,7 +45,7 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
     event OwnerChanged          (address indexed newOwner);
     event ModuleAdded           (address indexed module);
     event ModuleRemoved         (address indexed module);
-    event StaticMethodBound     (bytes4  indexed method, address indexed module);
+    event MethodBound           (bytes4  indexed method, address indexed module);
 
     event WalletSetup(address indexed owner);
 
@@ -145,7 +145,7 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
         return isAddressInSet(MODULE, _module);
     }
 
-    function bindStaticMethod(bytes4 _method, address _module)
+    function bindMethod(bytes4 _method, address _module)
         external
         onlyModule
     {
@@ -154,10 +154,10 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
         require(bankRegistry.isModuleRegistered(_module), "UNREGISTERED_MODULE");
 
         methodToModule[_method] = _module;
-        emit StaticMethodBound(_method, _module);
+        emit MethodBound(_method, _module);
     }
 
-    function staticMethodModule(bytes4 _method)
+    function boundMethodModule(bytes4 _method)
         public
         view
         returns (address)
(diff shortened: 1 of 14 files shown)


dong77:
I'm fine with this approach. Should we change `staticMethods` to `boundMethods`?

Brechtpd:
> Should we change `staticMethods` to `boundMethods`?

Yes, I'll remove the static everywhere.

> or maybe we should support `bindMethod(module, selector, isStatic)` then depends on `isStatic` to call `call` or `staticcall`

`call` works for both so I would just use `call`. I don't see a need to have different code paths.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -244,7 +239,7 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
         assembly {
             let ptr := mload(0x40)
             calldatacopy(ptr, 0, calldatasize())
-            let result := staticcall(gas, module, ptr, calldatasize(), 0, 0)
+            let result := call(gas, module, 0, ptr, calldatasize(), 0, 0)
```
dong77: Should we enable a flag for `bindMehtod` to indicate if a method should be called using `call` or `staticcall`? I am not sure what's the side effect if the target method is a veiw/pure function but we call it using `call`. Will there be an empty transaction?

Brechtpd: I really think there's no reason to do so.

- All our smart contracts using a proxy also use `delegatecall` to call all view functions. `delegatecall` would also allows state changes.
- Solidity didn't even use `staticcall` for view functions [before 0.5](https://solidity.readthedocs.io/en/v0.5.13/050-breaking-changes.html#interoperability-with-older-contracts):
  > This is due to the fact that starting with Solidity v0.5.0 staticcall is used to call view functions
- Only functions on modules can be called. The wallet address itself has no permissions on a module (which is what this PR changed). So even accidentally binding a non-view method of a module isn't a security issue. The functions of the modules must of course be properly guarded for meta-tx/... but using `call` here instead of `staticcall` makes no difference for that.
- If there's a security issue with the wallet that allows modules to be added that were not approved (which would allow the `call` to be exploited), then they would also be able to exploit `transact` directly. So that also doesn't make it worse security wise.

Issue #753, opened 2019-11-25
[hebao] fix issue in getAccountCreationOrUpdateFee in LoopringDEXModule.sol


dong77:
```solidity
 // TODO: This will actually throw when the account doesn't exist. Maybe we should change that so it's easier to use from smart contracts.
        (uint24 accountId, , ) = exchange.getAccount(wallet);

```

PR #790, opened 2019-12-01
[hebao] fix issue in calling Loopring getAccount method

dong77:
Fixing https://github.com/Loopring/protocols/issues/753

### DIFF
--- a/packages/hebao_v1/contracts/modules/exchanges/LoopringModule.sol
+++ b/packages/hebao_v1/contracts/modules/exchanges/LoopringModule.sol
@@ -117,9 +117,9 @@ contract LoopringModule is SecurityModule
         (bool success, bytes memory result) = address(exchange).staticcall(callData);
         if (success && result.length == 96) {
             assembly {
-                accountId := mload(result)
-                pubKeyX := mload(add(result, 32))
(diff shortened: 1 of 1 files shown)


packages/hebao_v1/contracts/modules/exchanges/LoopringModule.sol
```diff
@@ -110,7 +110,18 @@ contract LoopringModule is SecurityModule
...
+        bytes memory callData = abi.encodeWithSelector(
+            exchange.getAccount.selector,
+            wallet
+        );
+        (bool success, bytes memory result) = address(exchange).staticcall(callData);
+        if (success && result.length == 96) {
+            assembly {
+                accountId := mload(result)
+                pubKeyX := mload(add(result, 32))
+                pubKeyY := mload(add(result, 64))
```
Brechtpd: Needs an additional offset of 32 to skip over the length:
```
accountId := mload(add(result, 32))
pubKeyX := mload(add(result, 64))
pubKeyY := mload(add(result, 96))
```

dong77: I keep forgetting this

Issue #786, opened 2019-11-30
[hebao] More flexible/accurate gas costs


Brechtpd:
Currently the gas cost charged to the user is `gasSpent + GAS_OVERHEAD`, with `gasSpent` the gas used to execute the meta-tx.

There are some gas costs as well to put all data necessary on-chain for the meta-tx (the `calldata` costs). For example, for protocol V3 committing a block may only use 100,000 gas, but there are a lot of costs to put the data-availability data on-chain. This gas cost isn't part of `gasSpent` because the cost is only charged on the transaction itself.

While we could try to calculate this extra gas cost on-chain by checking the length of all the data passed in, I don't think this really worth it because gas costs may change and other gas costs can still exist (initial 21,000 gas cost/transaction for example).

I think putting the `GAS_OVERHEAD` inside the data signed by the user so it can simply be used on-chain as-is (and the value used calculated off-chain) is a more flexible and easy way to charge these extra gas costs to the user.

dong77:
fixed

PR #792, opened 2019-12-01
[hebao] add gasOverhead parameter to executeMetaTx

dong77:
addressing  https://github.com/Loopring/protocols/issues/786

### DIFF
--- a/packages/hebao_v1/contracts/base/MetaTxModule.sol
+++ b/packages/hebao_v1/contracts/base/MetaTxModule.sol
@@ -47,8 +47,6 @@ contract MetaTxModule is BaseModule
     using SignatureUtil for bytes32;
     using BytesUtil     for bytes;
 
-    uint constant public MIN_GAS_RESERVE = 20000;
-
     struct WalletState
     {
         uint nonce;
@@ -88,9 +86,9 @@ contract MetaTxModule is BaseModule
     /// @param gasToken The token to pay the relayer as fees. Use address(0) for Ether.
     /// @param gasPrice The amount of `gasToken` to pay per gas. 0 is a valid value.
     /// @param gasLimit The max amount of gas that can be used by this meta transaction,
-    ///                including the `gasReserve`.
-    /// @param gasReserve The amount of `gasToken` reserved for the meta data storage and
-    ///                   extra logic in the meta-transaction.
+    ///                 excluding the `gasOverhead`.
+    /// @param gasOverhead The extra amount of `gasToken` paid for data storage and
+    ///                   extra fee-payment internal transaction.
     /// @param signatures Signatures.
     function executeMetaTx(
         bytes   calldata data,
@@ -98,17 +96,16 @@ contract MetaTxModule is BaseModule
         address gasToken,
         uint    gasPrice,
         uint    gasLimit,
-        uint    gasReserve,
+        uint    gasOverhead,
         bytes[] calldata signatures
         )
         external
         payable
     {
-        require(gasReserve >= MIN_GAS_RESERVE, "INVALID_GAS_RESERVE");
-        require(gasLimit > gasReserve, "INVALID_GAS_LIMIT");
+        require(gasLimit > 0, "INVALID_GAS_LIMIT");
 
         uint startGas = gasleft();
-        require(startGas >= gasLimit, "OUT_OF_GAS");
+        require(startGas >= gasLimit.add(gasOverhead), "OUT_OF_GAS");
 
         address wallet = extractWalletAddress(data);
         bytes32 metaTxHash = getSignHash(
@@ -120,7 +117,7 @@ contract MetaTxModule is BaseModule
             gasToken,
             gasPrice,
             gasLimit,
-            gasReserve
+            gasOverhead
         );
 
         address[] memory signers = extractMetaTxSigners(
@@ -142,7 +139,7 @@ contract MetaTxModule is BaseModule
             gasToken,
             gasPrice,
             gasLimit,
-            gasReserve,
+            gasOverhead,
             startGas - gasleft()
         );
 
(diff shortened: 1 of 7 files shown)


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -137,17 +154,18 @@ contract MetaTxModule is BaseModule
         address gasToken,
         uint    gasPrice,
         uint    gasLimit,
+        uint    gasReserve,
         uint    gasSpent
         )
         private
     {
-        if (gasPrice == 0) return;
+        uint gasUsed = gasSpent + gasReserve;
```
Brechtpd: Maybe do `gasSpent.add(gasReserve)` just to be sure.

dong77: done

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -88,19 +88,25 @@ contract MetaTxModule is BaseModule
...
         uint    gasPrice,
         uint    gasLimit,
+        uint    gasReserve,
         bytes[] calldata signatures
         )
         external
         payable
     {
+        require(gasReserve >= MIN_GAS_RESERVE, "INVALID_GAS_RESERVE");
+        require(gasLimit > gasReserve, "INVALID_GAS_LIMIT");
```
Brechtpd: I don't think we need to do this check. I would just charge `gasReserve` on top of `gasSpent` without any other rules (`gasReserve` could be a lot higher than the gas needed to actually execute the code on-chain). So just using `gasLimit` as the actual gas limit on-chain seems easier.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -127,7 +137,14 @@ contract MetaTxModule is BaseModule
         // solium-disable-next-line security/no-call-value
         (bool success,) = address(this).call.value(msg.value)(data);
```
Brechtpd: Best to change this to 
```
(bool success,) = address(this).call.gas(gasLimit.sub(startGas.sub(gasleft())).value(msg.value)(data);
```
so the function call can't use more gas than allowed. If it uses more `success == false` and the relayer still gets paid.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -137,17 +154,18 @@ contract MetaTxModule is BaseModule
...
         uint    gasLimit,
+        uint    gasReserve,
         uint    gasSpent
         )
         private
     {
-        if (gasPrice == 0) return;
+        uint gasUsed = gasSpent.add(gasReserve);
+        require(gasUsed <= gasLimit, "EXCEED_GAS_LIMIT");
+        require(gasReserve <= gasleft(), "OUT_OF_GAS");
```
Brechtpd: These checks seem unnecessary with the changes above.

packages/hebao_v1/contracts/base/MetaTxModule.sol
dong77: I added a TODO. The change will cause a stack-too-deep error and I didn't manage to fix it.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -137,22 +156,21 @@ contract MetaTxModule is BaseModule
         address gasToken,
         uint    gasPrice,
         uint    gasLimit,
+        uint    gasOverhead,
         uint    gasSpent
         )
         private
     {
-        if (gasPrice == 0) return;
+        require(gasSpent <= gasLimit, "EXCEED_GAS_LIMIT");
```
Brechtpd: I would remove this check because the gas is now limited in the `call`, so if that call uses up all gas it still costs a bit of gas to call `reimburseGasFee` before this check is done, which could still fail the transaction. This would mean the relayer pays for the complete failing transactions even though he did everything right.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -88,21 +86,26 @@ contract MetaTxModule is BaseModule
...
         bytes[] calldata signatures
         )
         external
         payable
     {
+        require(gasLimit > 0, "INVALID_GAS_LIMIT");
+
         uint startGas = gasleft();
-        require(startGas >= gasLimit, "OUT_OF_GAS");
+        require(startGas >= gasLimit.add(gasOverhead), "OUT_OF_GAS");
```
Brechtpd: I would just do `require(startGas >= gasLimit, "OUT_OF_GAS");` because the `gasOverhead` gas is already used up before `gasleft()` is stored in `startGas`.

packages/hebao_v1/contracts/base/MetaTxModule.sol
dong77: If we remove this line, there there is no place we ever check against `gasLimit`.

dong77: The gasLimit is used to measure the gas used since the first line of code in `executeMetaTx` but before we call `reimburseGasFee`:

```solidity
    function executeMetaTx(
        bytes   calldata data,
        uint    nonce,
        address gasToken,
        uint    gasPrice,
        uint    gasLimit,
        uint    gasOverhead,
        bytes[] calldata signatures
        )
        external
        payable
    {
        require(gasLimit > 0, "INVALID_GAS_LIMIT");

        uint startGas = gasleft();
        require(startGas >= gasLimit, "OUT_OF_GAS");

         [code measured by gasLimit]

        reimburseGasFee(
            wallet,
            gasToken,
            gasPrice,
            gasLimit,
            gasOverhead,
            startGas - gasleft()
        );
    }

```

The gasOverhead includes both the storage of calldata and the `reimburseGasFee` method.

Brechtpd: > If we remove this line, there there is no place we ever check against `gasLimit`.

We now do `address(this).call.gas(gasLimit.sub(startGas.sub(gasleft()))).value(msg.value)(data)` which makes it impossible to use more than `gasLimit` until after the call ended. But a bit more gas is used  for emitting the `ExecutedMetaTx` event and preparing the stack to call `reimburseGasFee` that is also measured until `startGas - gasleft()`. So it's possible `gasSpent` is slightly higher than `gasLimit`, even though this is completely valid (the call may fail having using up all requested gas and the relayer still gets paid, but the user is ultimately responsible to have a high enough `gasLimit`)

If we would do `require(gasSpent <= gasLimit, "EXCEED_GAS_LIMIT");` that would mean the transaction would always fail if the `call` uses it's complete amount of gas allowed of `gasLimit.sub(startGas.sub(gasleft()))`, which is certainly not what we want because the relayer would have simply wasted it's own gas for a failed transaction. It would be possible to let the `call` use a little bit less gas (e.g. subtract another 2,000 gas from `gaslimit`), but with changing gas prices for different operations on ETH 1.x I don't think that's a good idea.

Perhaps from a user point of view it makes more sense to just use `gasLimit` directly for the gas limit of the call itself (so `address(this).call.gas(gasLimit).value(msg.value)(data)`). Everything else (signature checking, emitting events, ...) can also be part of `gasOverhead`.  Users may know better how much the actual call costs (from having used Ethereum directly perhaps) and not really know how much the meta-tx part costs (as this will be different depending on the number of signatures etc... which the user may not really know, a different number of signatures may even be needed for doing the same operation in some cases). As `gasOverhead` is something that is automatically generated for users and `gasLimit` is something that users may be able to set themselves I think that would make sense.

dong77:
please take another look.

Issue #785, opened 2019-11-30
[hebao] Support delegate calls


Brechtpd:
This is something that is done in GnosisSafe which I like. `delegatecall` is supported which can be used to support libraries. The idea is that some extra logic can be done on a call which still executes in the wallet (so `msg.sender` is still the wallet address).

This allows us to do some nice things  which makes things more future proof as well (support new EVM opcodes):
- Allows ue to support more rarely used functions without modifying the wallet. Things like `create` and `create2` may still be useful to be callable by the smart wallet (so the smart wallet is immediately the owner of the created smart contract for example). A module can just let the data in `transact` pass through the library contract that implements this logic.
- Can also be used to do things like #784 by doing the multiple calls logic in a `delegatecall`

letsgoustc:
> This is something that is done in GnosisSafe which I like. `delegatecall` is supported which can be used to support libraries. The idea is that some extra logic can be done on a call which still executes in the wallet (so `msg.sender` is still the wallet address).
> 
> This allows us to do some nice things which makes things more future proof as well (support new EVM opcodes):
> 
> * Allows ue to support more rarely used functions without modifying the wallet. Things like `create` and `create2` may still be useful to be callable by the smart wallet (so the smart wallet is immediately the owner of the created smart contract for example). A module can just let the data in `transact` pass through the library contract that implements this logic.
> * Can also be used to do things like #784 by doing the multiple calls logic in a `delegatecall`

How about we add one more API delegateTransact for Wallet in which we call delegatecall instead of current transact API with call.

Brechtpd:
Yes, I think another external function would work nicely. Internally I would use the same `transactInternal` for normal and delegate calls though because the code can largely be reused (as it's basically the same).

dong77:
Not sure I followed the idea. Are you suggesting when we bind methods, we can give it a type whichi decides whether we should use `staticcalll`, `call`, or `delegatecall` in the default function?

Brechtpd:
It would be mainly for `transact` in the wallet so we can do more logic in the wallet itself without changing `msg.sender` (doing the logic in another contract would change msg.sender for subsequent calls which isn't something we want in some cases). I'll make a PR for this.

dong77:
I see what you mean. So add a `mode` parameter to `transact`? 

```solidity

    enum TransactMode {CALL, DELEGATECALL}
    function transact(
        address to,
        TransactMode mode,
        uint    value,
        bytes   calldata data
        )
        external
[...]

PR #793, opened 2019-12-01
[hebao] support delegatecall in wallet transact

dong77:
addressing https://github.com/Loopring/protocols/issues/785

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseModule.sol
+++ b/packages/hebao_v1/contracts/base/BaseModule.sol
@@ -80,7 +80,7 @@ contract BaseModule is Module, ReentrancyGuard
         internal
         returns (bytes memory)
     {
-        return Wallet(wallet).transact(uint8(0), to, value, data);
+        return Wallet(wallet).transact(uint8(1), to, value, data);
     }
 
     function transact2(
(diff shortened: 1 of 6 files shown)


packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -168,15 +168,16 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
     }
 
     function transact(
-        address to,
-        uint    value,
-        bytes   calldata data
+        uint8    mode,
```
Brechtpd: Why no enum?

packages/hebao_v1/contracts/base/BaseModule.sol
```diff
@@ -155,4 +155,28 @@ contract BaseModule is Module, ReentrancyGuard
             w.bindStaticMethod(methods[i], address(0));
         }
     }
+
+    function transact1(
```
Brechtpd: Instead of `transact1` and `transact2`, why not something more clear like `transactCall` and `transactDelegateCall`?

packages/hebao_v1/contracts/base/BaseWallet.sol
dong77: No specific reason, I just prefer simple uint :)

packages/hebao_v1/contracts/base/BaseModule.sol
dong77: done

Issue #775, opened 2019-11-29
[hebao] apply a delay when an address is whitelisted.


dong77:
WhitelistStore: When an address is added to the whitelist, it needs to wait for 1 day to take effect.

dong77:
fixed

PR #794, opened 2019-12-01
[hebao] Delayed whitelisting

dong77:
addressing  issue https://github.com/Loopring/protocols/issues/775

### DIFF
--- a/packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
+++ b/packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
@@ -89,7 +89,7 @@ contract WhitelistStore is DataStore
         )
     {
         timestamp = timestampMap[wallet][addr];
-        isWhitelistedAndEffective = timestamp > 0 && timestamp < now;
+        isWhitelistedAndEffective = timestamp > 0 && timestamp <= now;
     }
 
     function whitelistSize(address wallet)
--- a/packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
+++ b/packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
@@ -26,7 +26,7 @@ import "../../base/DataStore.sol";
 contract WhitelistStore is DataStore
 {
     // wallet => whitelisted_addr => effective_since
-    mapping(address => mapping(address => uint)) public timestampMap;
+    mapping(address => mapping(address => uint)) public effectiveTimeMap;
 
     event Whitelisted(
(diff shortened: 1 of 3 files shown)


packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
```diff
@@ -52,25 +60,37 @@ contract WhitelistStore is DataStore
...
         view
-        returns (bool)
+        returns (
+            bool isWhitelistedAndEffective,
+            uint timestamp
+        )
     {
-        return isAddressInSet(walletKey(wallet), addr);
+        timestamp = timestampMap[wallet][addr];
+        isWhitelistedAndEffective = timestamp > 0 && timestamp <= now;
```
Brechtpd: I think this should be `isWhitelistedAndEffective = timestamp >= now;`

packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
```diff
@@ -25,13 +25,20 @@ import "../../base/DataStore.sol";
 /// @dev This store maintains a wallet's whitelisted addresses.
 contract WhitelistStore is DataStore
 {
+    uint delayToDffective;
```
Brechtpd: typo

packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
```diff
@@ -41,6 +48,7 @@ contract WhitelistStore is DataStore
         onlyManager
     {
         addAddressToSet(walletKey(wallet), addr, true);
+        timestampMap[wallet][addr] = now + delayToDffective;
```
Brechtpd: Maybe it's a good idea to just let the modules calculate the timestamp the address is actually effective? This way there's no need to configure this variable in this storage contract and modules could use extra logic because they have more context: something like the more signatures that were used to whitelist an address the less time it takes to make it effective.

packages/hebao_v1/contracts/modules/stores/WhitelistStore.sol
dong77: The effectiveness timestamp is a time in the future when addToWhitelist is called, for exmaple:

```addToWhitelist(wallet, address, now + 1 day)```

. so the current logic is correct.

Brechtpd: Oh yes of course, for some reason I thought the check did the opposite.

PR #795, opened 2019-12-02
[hebao] Allow multiple calls in a single meta-tx

Brechtpd:
This can still be improved.

### DIFF
--- a/packages/hebao_v1/contracts/base/MetaTxModule.sol
+++ b/packages/hebao_v1/contracts/base/MetaTxModule.sol
@@ -101,25 +101,31 @@ contract MetaTxModule is BaseModule
         uint startGas = gasleft();
         require(startGas >= gasSetting[2], "OUT_OF_GAS");
 
-        address wallet = extractWalletAddress(data);
+        // Copy data to memory
+        bytes memory _data = data;
+
+        address wallet = extractWalletAddress(_data);
         bytes32 metaTxHash = getSignHash(
             wallet, // from
             address(this),  // to. Note the relayer can only call its own methods.
             msg.value, // value
-            data,
+            _data,
             nonce,
             gasSetting
         );
 
-        bytes4 method = extractMethod(data);
+        bytes4 method = extractMethod(_data);
         // If we exectute multiple transactions using `executeTransactions` we don't need to check
         // any signatures now, we will check them for each transaction independently
         if (method != this.executeTransactions.selector) {
-            address[] memory signers = extractMetaTxSigners(wallet, method, data);
+            address[] memory signers = extractMetaTxSigners(wallet, method, _data);
             metaTxHash.verifySignatures(signers, signatures);
         } else {
-            // Make sure the hash passed into executeTransactions is correct.
-            require(metaTxHash == data.toBytes32(4 + 32), "INVALID_DATA");
+            // As `metaTxHash` is the hash of `data` it can't contain the hash of `data`, so set it here
+            require(metaTxHash == 0x0, "INVALID_DATA");
+            assembly {
+                mstore(add(_data, 68), metaTxHash)
+            }
         }
 
         // Mark the transaction as used before doing the call to guard against re-entrancy
@@ -127,9 +133,7 @@ contract MetaTxModule is BaseModule
         saveExecutedMetaTx(wallet, nonce, metaTxHash);
 
         // solium-disable-next-line security/no-call-value
-        // (bool success,) = address(this).call.value(msg.value)(data);
-
-        (bool success,) = address(this).call.gas(gasSetting[2]).value(msg.value)(data);
+        (bool success,) = address(this).call.gas(gasSetting[2]).value(msg.value)(_data);
 
         emit ExecutedMetaTx(msg.sender, wallet, nonce, metaTxHash, success);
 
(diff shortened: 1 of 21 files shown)


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -131,6 +139,45 @@ contract MetaTxModule is BaseModule
...
+        uint totalValue = 0;
+        for (uint i = 0; i < data.length; i++) {
+            // Check that the wallet is the same for all transactions
+            address txWallet = extractWalletAddress(data[i]);
+            require(txWallet == wallet, "INVALID_DATA");
+
+            // Verify the signatures for this transaction
+            bytes4 method = extractMethod(data[i]);
+            address[] memory signers = extractMetaTxSigners(wallet, method, data[i]);
+            metaTxHash.verifySignatures(signers, signatures);
```
dong77: the `signers` and `signatures` won't match correctly. For example, if signatures are from address, A, B, and C. but singers can be A, and C.

Brechtpd: Yeah, but I think having special ordering for signers will be pretty rare I think? Well, the wallet order at signers[0] is pretty common, but otherwise it doesn't seem common that signers need a fixed order.

I will try to find a way to make it as general as possible, but having to check all the signatures for each transaction (or needing complex logic to work around it) seems like a waste of gas for some very uncommon uses. But, I'm still trying to find better ways to do it.

Brechtpd: This is now fixed in an acceptable/cheap way I think. No order of signers is enforced any more.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -101,27 +101,39 @@ contract MetaTxModule is BaseModule
...
-        metaTxHash.verifySignatures(signers, signatures);
+        bytes4 method = extractMethod(_data);
+        // If we exectute multiple transactions using `executeTransactions` we don't need to check
+        // any signatures now, we will check them for each transaction independently
+        if (method != this.executeTransactions.selector) {
+            address[] memory signers = extractMetaTxSigners(wallet, method, _data);
+            metaTxHash.verifySignatures(signers, signatures);
+        } else {
+            // As `metaTxHash` is the hash of `data` it can't contain the hash of `data`, so set it here
+            require(metaTxHash == 0x0, "INVALID_DATA");
```
dong77: metaTxHash  is calculated by getSignHash, so it's unlikely it will be 0x0.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -101,27 +101,39 @@ contract MetaTxModule is BaseModule
...
+            assembly {
+                mstore(add(_data, 68), metaTxHash)
```
dong77: why save the hash, seems it is not used anywhere.

dong77: does it really just replace the value of the `metaTxHash` parameter in `executeTransactions `?

Brechtpd: It's an input parameter for `executeTransactions`, so the calldata here is modified directly so the `call` can be done as usual. That's where the signatures are currently checked so we need to know the hash of the complete meta transaction.

I want to make as few changes to `executeMetaTx` as possible, so I'm trying to find a better way to do this.

dong77: This is actually fine.

packages/hebao_v1/contracts/base/MetaTxModule.sol
Brechtpd: It needs to be temporarily set to `0x0` because `data` is used to calculate `metaTxHash`. So `metaTxHash` can't be part of `data` otherwise we modify the data that is used to calculate the hash, which isn't possible of course. :)

So my solution to this was to set it to `0x0` for the hash calculation and then set it to the correct value on-chain. I would also like to find another way to do this.

dong77: is it possible to do it this way. 

```
if (method != this.executeTransactions.selector) {
   metaTxHash = getSignHash()...
} else {
  //do nothing
}

```
then in executeTransactions, drop the metaTxHash parameter, and calculate the hash inside the function using it's own selector and parameters...

Brechtpd: I have found a way to make it completely general without any changes to `executeMetaTx` which I think is very nice because `executeTransactions` is now a module function like any other. If a module doesn't want to support or a different function for multiple calls is wanted for some reason that's very easy now.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -162,7 +204,27 @@ contract MetaTxModule is BaseModule
         )
         internal
         view
-        returns (address[] memory signers);
+        returns (address[] memory signers)
```
dong77: I'd prefer creating another private method for this and keep extractMetaTxSigners as an abstract method so we do't have to use the following boilerplate code:

```
signers = super.extractMetaTxSigners(wallet, method, data);
        if (signers.length > 0) {
            return signers;
}
```

PR #798, opened 2019-12-02
[hebao] Sub account module

dong77:
The objective of this PR is to introduce a SubAccount contract interface to make it simpler for the App to interact with DeFi modules. The balance query and wallet<->subacounts transfer will be easier (note that inter-subaccount transfers are not allowed).

The getBalances/getBalance function will return `int` instead of `uint` to reflect negative borrowing balances.

The LoopringModule probably will not become a SubAccount as the transact methods are more complicated and the balances cannot be queried onchain.


packages/hebao_v1/contracts/iface/SubAccount.sol
```diff
@@ -0,0 +1,44 @@
...
+{
+    function tokenBalance (address wallet, address token) public view returns (int);
+    function tokenBalances(address wallet, address[] memory tokens) public view returns (int[] memory balances);
+
+    /// @dev Deposits Ether/token from the wallet to this sub-account.
+    function deposit(
+        address wallet,
+        address token,
+        uint    amount,
+        bytes[] calldata signatures
```
Brechtpd: Why does `deposit`/`withdraw` have `signatures` as input?

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -36,14 +37,15 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
     address internal _owner;
 
     bytes32 internal constant MODULE = keccak256("__MODULE__");
+    bytes32 internal constant SUB_ACCOUNTS = keccak256("__SUB_ACCOUNT__");
```
Brechtpd: Do we need to keep track of modules that are sub accounts in the wallet? Seems like this is mainly useful off-chain where we can just loop over all modules and not something we really need to store in the wallet on-chain.

packages/hebao_v1/contracts/iface/SubAccount.sol
dong77: this should be signers :) my bad

PR #807, opened 2019-12-02
[hebao] Work around Istanbul gas cost changes for send/transfer

Brechtpd:
See here for more info:
- https://github.com/gnosis/safe-contracts/issues/149

This is to ensure the smart wallet can still be used to receive ETH with `send`/`transfer` after Istanbul.

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseWallet.sol
+++ b/packages/hebao_v1/contracts/base/BaseWallet.sol
@@ -224,7 +224,12 @@ contract BaseWallet is Wallet, AddressSet, ReentrancyGuard
         payable
     {
         if (msg.value > 0) {
-            emit Received(msg.sender, msg.value, msg.data);
+            // If this was called from an EOA or we have enough gas we should emit an event
+            // Note: we check the origin to avoid that EOA transactions are estimated too low to emit the event
+            // solium-disable-next-line security/no-tx-origin
+            if (msg.sender == tx.origin || gasleft() > 1500) {
+                emit Received(msg.sender, msg.value, msg.data);
+            }
             return;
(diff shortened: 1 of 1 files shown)


dong77:
Sorry I  didn’t get it. Why not emit if msg.sender is a contract?

Brechtpd:
Before Istanbul, which changes certain important gas prices (like SLOAD), it was good practice to send ETH in an account using `send`/`transfer` which limited the amount of gas to 2300. This was good practice because that's not enough gas to make re-entrancy attacks possible. So a lot of contracts do just that.

If a contract uses `send`/`transfer` to send ETH to our smart wallet address we have the following costs:
- First the proxy contract is called. This loads the implementation contract from storage (SLOAD). Cost: 800 gas
- Then the implementation contract is called with `delegatecall`. Cost: 700 gas
- In the fallback function we then (before this PR) always emit an event. Cost: ~1000 gas

Adding all these gas costs together we get ~2500 gas (which doesn't yet contain all gas costs). This would mean that `send`/`transfer` for our smart wallet would always fail, which would be a big problem to maintain compatibility with a lot of existing contracts. This PR works around this by not emitting the event when not enough gas is left, which is the case with `send`/`transfer`. By not emitting the event the gas costs will be lower than the 2300 gas available in `send`/`transfer` which allows them to still work as expected.

This isn't really important when sending ETH from a normal address because in that case you can set the gas limit yourself.

dong77:
Good to know all the details. But I think if we do let emit the event consistently, then maybe we don't want to emit the event at all...

Brechtpd:
Also good for me. :)

PR #812, opened 2019-12-03
[hebao]  refactor Controller

dong77:


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -100,6 +109,10 @@ contract MetaTxModule is BaseModule
 
         uint startGas = gasleft();
         require(startGas >= gasSetting[2], "OUT_OF_GAS");
+        require(
+            controller.relayerRegistry().isRelayerRegistered(msg.sender),
+            "INVALID_RELAYER"
+        );
```
Brechtpd: I'm not yet convinced that only allowing registered relayers is a good idea:
- The design requires meta transaction for critical operations: it is the only way to send multiple signatures. If a registered relayer doesn't want to processes the meta tx (or simply stopped working) the user is out of luck. This means the wallet owner is not in full control of his wallet and needs to depend on a third party for things like recovering his wallet.
- We will operate the wallet software and a relayer. This means that we not only have easier access to the private keys of users, we're also the relayer that can drain full balances of users with just a single private key. That means users really have to trust us for no particular reason (as the smart wallet code alone does not ensure safety for users).
- Even if a relayer is an independent third party the user still needs to trust the relayer because there's still a risk of him working together with an attacker for his own benefit.

I think the relayer should be completely trustless. Argent does this by requiring at least 2 signatures for a meta transaction that can do gas reimbursement which in theory is a good solution I guess, but the user may not have quick access to another keypair on his phone for example. I still think the best solution is if it's just the wallet owner that signed the meta-tx the cost should deducted from his quota, but I understand that's tedious and expensive.

dong77: I'll remove the relayer registry.

dong77: The MetaTxModule is supposed to be independent of the QuoaTransfers module and is more fundamental than QuoaTransfers. Using QuoaTransfers inside MetaTxModule doesn't make sense to me :(

Let's keep this in mind and see if we can find some other solutions. One approach is using the price feed to make sure the user doesn't pay token fees more than a pre-defined Ether value, but this is not going to help tokens without a price feed (in such case its Ether value is treaded as 0, same as in QuotaTransfers). So this apporach is no ideal.

PR #816, opened 2019-12-03
[hebao] trace meta-tx gas

dong77:
A possible improvement to this PR is to have two quotas per wallet, one for transfers, and one for gas.


Brechtpd:
> A possible improvement to this PR is to have two quotas per wallet, one for transfers, and one for gas.

I don't think that would be necessary. In normal use the quota used by gas will be low or zero. In an attack the attacker would be able to drain both quota, so I think just a single quota is fine.

Perhaps the quota shouldn't be enforced for gas reimbursement when the wallet owner signs and `(n+1)*2` signatures are provided, but that many signatures will rarely be provided so it's probably not worth it to create a special case for it.

PR #823, opened 2019-12-04
[hebao] Allow the wallet to re-enable a module that was already added

Brechtpd:
This allows a wallet owner or guardian to re-enable a module that was already added to the wallet when it was still registered. This does not allow any module to be registered, it only allows module to still be used that was already being used.

The problem with the registry is that any module can be disabled. So if the wallet only has a single module (that was just disabled) this renders the wallet unusable. The other problem is that wallets need to trust the owner of the ModuleRegistry to not act maleficent (and the keys controlled by the owner (contract) could also be stolen).

So this allows the module to still be disabled immediately for safety by a central party, but it can be overridden when this mechanism is used in bad faith.

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseWallet.sol
+++ b/packages/hebao_v1/contracts/base/BaseWallet.sol
@@ -42,6 +42,8 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
 
     mapping (bytes4  => address) internal methodToModule;
 
+    mapping (address => bool) public moduleForceEnabled;
+
     event OwnerChanged          (address indexed newOwner);
     event ModuleAdded           (address indexed module);
     event ModuleRemoved         (address indexed module);
@@ -64,12 +66,19 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
 
     modifier onlyModule
     {
-        require(isAddressInSet(MODULE, msg.sender), "MODULE_UNAUTHORIZED");
-        require(
-            controller.moduleRegistry().isModuleRegistered(msg.sender),
-            "INVALID_MODULE"
-        );
+        requireModuleEnabled(msg.sender);
         _;
+    }
+
+    function requireModuleEnabled(address module)
+        private
+        view
+        returns (bool)
+    {
+        require(isAddressInSet(MODULE, module), "MODULE_UNAUTHORIZED");
+        if(!controller.moduleRegistry().isModuleRegistered(module)) {
+            require(moduleForceEnabled[module], "INVALID_MODULE");
+        }
     }
 
     function owner() public view returns (address)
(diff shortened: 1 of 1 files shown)


dong77:
Brecht, I think in my original code, we don't check anymore whether a module is still registered in the registry once it's been added.  So maybe we should simply remove the second `require`

```solidity

    /// @dev This default function can receive Ether or perform queries to modules
    ///      using bound methods.
    function()
        external
        payable
    {
        if (msg.data.length == 0 || msg.value > 0) {
            // Note: don't do anything here so send/transfer still works
            // -- the 2300 gas limit for send/transfer may case out-of-gas
            // issue after Istanbul.
            return;
        }

        address module = methodToModule[msg.sig];
        require(isAddressInSet(MODULE, module), "MODULE_UNAUTHORIZED");
        // REMOVE THIE FOLLOWIGN LINE?
        require(controller.moduleRegistry().isModuleRegistered(module), "INVALID_MODULE");

        assembly {
            let ptr := mload(0x40)
            calldatacopy(ptr, 0, calldatasize())
            let result := call(gas, module, 0, ptr, calldatasize(), 0, 0)
            returndatacopy(ptr, 0, returndatasize())

            switch result
            case 0 { revert(ptr, returndatasize()) }
            default { return(ptr, returndatasize()) }
        }
    }
```

Brechtpd:
The original code also checks this for all incoming module calls:
```
modifier onlyModule
{
    require(isAddressInSet(MODULE, msg.sender), "MODULE_UNAUTHORIZED");
    require(
        controller.moduleRegistry().isModuleRegistered(msg.sender),
        "INVALID_MODULE"
    );
    _;
}
```

Removing the checks would also be okay for me, so `isModuleRegistered` is only used when modules are added.

dong77:
Yes, please just remove the check in the default function to keep it simple.

Brechtpd:
> Yes, please just remove the check in the default function to keep it simple.

I removed all checks except when adding the module. The check that was done in `onlyModule` is much more impactful than the one in the default function.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -151,10 +147,6 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
     {
         require(_method != bytes4(0), "BAD_METHOD");
         require(methodToModule[_method] == address(0), "METHOD_BOUND_ALREADY");
-        require(
-            controller.moduleRegistry().isModuleRegistered(_module),
```
dong77: Maybe we should keep this one.

Brechtpd: `bindMethod` is also used for unbinding methods, so I removed it here as well to allow modules to be completely removed. Because the module will probably be completely setup, with all methods necessary bound, when it's added I don't think it matters much.

PR #828, opened 2019-12-06
[hebao] Guardian grouping

Brechtpd:
The basic idea is to store data about the guardian and let modules use that data for more complex logic. This way we do not limit ourselves to a single system which we're than stuck with. So storing the guardians and adding/removing guardians remains just as easy as before, there's just a single extra `uint` value stored for each guardian. The modules could store any data in it, but currently it's used as a bitmask to store some extra data about the guardian.

This means there's a bit more work to 'prepare' the guardians, but that's insignificant with checking the actual signatures of the guardians. It allows a lot of flexibility in the modules (like allow a self controlled hardware guardian to do transfers with the same quota as the wallet owner or something like that).

I also changed some requirements in respect to when the owner needs or is allowed to sign.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
@@ -36,8 +36,8 @@ contract GuardianModule is SecurityModule
     mapping (address => mapping(address => mapping(uint => uint))) public pendingAdditions;
     mapping (address => mapping(address => uint)) public pendingRemovals;
 
-    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint info, uint confirmAfter);
-    event GuardianAdded             (address indexed wallet, address indexed guardian, uint info);
+    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint types, uint confirmAfter);
+    event GuardianAdded             (address indexed wallet, address indexed guardian, uint types);
     event GuardianAdditionCancelled (address indexed wallet, address indexed guardian);
 
     event GuardianRemovalPending    (address indexed wallet, address indexed guardian, uint confirmAfter);
@@ -59,7 +59,7 @@ contract GuardianModule is SecurityModule
     function addGuardian(
         address wallet,
         address guardian,
-        uint    info
+        uint    types
         )
         external
         nonReentrant
@@ -74,47 +74,47 @@ contract GuardianModule is SecurityModule
         require(count < MAX_GUARDIANS, "TOO_MANY_GUARDIANS");
 
         if (controller.securityStore().numGuardians(wallet) == 0) {
-            controller.securityStore().addGuardian(wallet, guardian, info);
-            emit GuardianAdded(wallet, guardian, info);
+            controller.securityStore().addOrUpdateGuardian(wallet, guardian, types);
+            emit GuardianAdded(wallet, guardian, types);
         } else {
-            uint confirmStart = pendingAdditions[wallet][guardian][info];
+            uint confirmStart = pendingAdditions[wallet][guardian][types];
             require(confirmStart == 0 || now > confirmStart + confirmPeriod, "ALREADY_PENDING");
-            pendingAdditions[wallet][guardian][info] = now + pendingPeriod;
-            emit GuardianAdditionPending(wallet, guardian, info, now + pendingPeriod);
+            pendingAdditions[wallet][guardian][types] = now + pendingPeriod;
+            emit GuardianAdditionPending(wallet, guardian, types, now + pendingPeriod);
         }
     }
 
     function confirmGuardianAddition(
         address wallet,
         address guardian,
-        uint    info
+        uint    types
         )
         external
         nonReentrant
         onlyFromMetaTxOrWalletOwner(wallet)
         onlyWhenWalletUnlocked(wallet)
     {
-        uint confirmStart = pendingAdditions[wallet][guardian][info];
+        uint confirmStart = pendingAdditions[wallet][guardian][types];
         require(confirmStart != 0, "NOT_PENDING");
         require(now > confirmStart && now < confirmStart + confirmPeriod, "EXPIRED");
-        controller.securityStore().addGuardian(wallet, guardian, info);
-        delete pendingAdditions[wallet][guardian][info];
-        emit GuardianAdded(wallet, guardian, info);
+        controller.securityStore().addOrUpdateGuardian(wallet, guardian, types);
+        delete pendingAdditions[wallet][guardian][types];
+        emit GuardianAdded(wallet, guardian, types);
     }
 
     function cancelGuardianAddition(
         address wallet,
         address guardian,
-        uint    info
+        uint    types
         )
         external
         nonReentrant
         onlyFromMetaTxOrWalletOwner(wallet)
         onlyWhenWalletUnlocked(wallet)
     {
-        uint confirmStart = pendingAdditions[wallet][guardian][info];
+        uint confirmStart = pendingAdditions[wallet][guardian][types];
         require(confirmStart > 0, "INVALID_GUARDIAN");
-        delete pendingAdditions[wallet][guardian][info];
+        delete pendingAdditions[wallet][guardian][types];
         emit GuardianAdditionCancelled(wallet, guardian);
     }
 
--- a/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
@@ -33,11 +33,11 @@ contract GuardianModule is SecurityModule
     uint public pendingPeriod;
     uint public confirmPeriod;
 
-    mapping (address => mapping(address => uint)) public pendingAdditions;
+    mapping (address => mapping(address => mapping(uint => uint))) public pendingAdditions;
     mapping (address => mapping(address => uint)) public pendingRemovals;
 
-    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint confirmAfter);
-    event GuardianAdded             (address indexed wallet, address indexed guardian);
+    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint info, uint confirmAfter);
+    event GuardianAdded             (address indexed wallet, address indexed guardian, uint info);
     event GuardianAdditionCancelled (address indexed wallet, address indexed guardian);
 
     event GuardianRemovalPending    (address indexed wallet, address indexed guardian, uint confirmAfter);
@@ -58,7 +58,8 @@ contract GuardianModule is SecurityModule
 
     function addGuardian(
         address wallet,
-        address guardian
+        address guardian,
+        uint    info
         )
         external
         nonReentrant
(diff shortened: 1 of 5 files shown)


packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return (numSigners >= (numTotal + 1) / 2);
+    }
+
+    function countGuardians(
+        SecurityStore.Guardian[] memory guardians
+        )
+        internal
+        pure
+        returns (uint numSelfControlled, uint numFriends, uint numFamily, uint numSocialOther)
```
dong77: multilines

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+    {
+        for (uint i = 0; i < guardians.length; i++) {
+            if(guardians[i].info & GuardianType.SelfControlled() != 0) {
```
dong77: `if(` => `if (`

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -65,15 +82,20 @@ contract SecurityStore is DataStore
...
+    function addGuardian(address wallet, address guardian, uint info)
         public
         onlyManager
     {
         require(guardian != address(0), "ZERO_ADDRESS");
         Wallet storage w = wallets[wallet];
         require(w.guardianIdx[guardian] == 0, "GUARDIAN_EXISTS");
 
-        w.guardians.push(guardian);
+        Guardian memory g = Guardian(
```
dong77: if we allow a guardian to have different roles, then maybe we should do:

`info |= existing.info` and remove the `require` above.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+                numSelfControlled++;
+            } else {
+                if(guardians[i].info & GuardianType.Friend() != 0) {
```
dong77: the & operator indicates one guardian may have different types.
I think we can rename `info` to `types` or `type`

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -65,15 +82,20 @@ contract SecurityStore is DataStore
         return wallets[wallet].guardians.length;
     }
 
-    function addGuardian(address wallet, address guardian)
+    function addGuardian(address wallet, address guardian, uint info)
```
dong77: should we check `info != 0`

packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -45,6 +46,16 @@ contract RecoveryModule is SecurityModule
     uint public recoveryPeriod;
     uint public lockPeriod;
 
+    modifier onlySufficientSigners(address wallet, address[] memory signers, GuardianUtils.SigRequirement requirement) {
```
dong77: multiline

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    function isGroupCriteriaValid(
+        uint numSigners,
+        uint numTotal
+        )
+        internal
+        pure
+        returns (bool)
+    {
+        return (numSigners >= (numTotal + 1) / 2);
```
dong77: should we use (numTotal / 2) + 1

so for an even-number of signers, we need more than 50%.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+                numGroupsActive++;
+                numGroupsSatisfied += isGroupCriteriaValid(numFamily, totalNumFamily) ? 1 : 0;
+            }
+            if (totalNumSocialOther > 0) {
+                numGroupsActive++;
+                numGroupsSatisfied += isGroupCriteriaValid(numSocialOther, totalNumSocialOther) ? 1 : 0;
+            }
+
+            require(numGroupsActive > 0, "SOCIAL_AUTHENTICATION_IMPOSSIBLE");
+            require(numGroupsSatisfied >= (numGroupsActive*2 + 2) / 3, "NOT_ENOUGH_SIGNERS");
```
dong77: we can also use isGroupCriteriaValid here if it is changed as I suggested below.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+            uint numSocialOther
+        ) = countGuardians(signingGuardians);
+
+        if (requirement == SigRequirement.OwnerRequired) {
+            require(walletOwnerSigned, "WALLET_OWNER_SIGNATURE_REQUIRED");
+        }
+
+        if (numFriends + numFamily + numSocialOther == 0) {
+            /* Self authentication */
+            uint numSelfControlledSignersRequired = 2;
```
dong77: if there is only 1 self-controlled address, we should also allow it to recover. 
I think this is handled in the last two `require`s, is my understanding correct?

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+            // If the owner is allowed to sign:
+            // - increase the total number of self controlled guardians
+            // - increase the number of required self controlled guardians
+            if (requirement != SigRequirement.OwnerNotAllowed) {
+                totalNumSelfControlled += 1;
+                numSelfControlledSignersRequired += 1;
+            }
+            // Count the wallet owner as a self controlled guardian
+            if (walletOwnerSigned) {
```
dong77: I think this should move into the above if-statement

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+                numSelfControlled += 1;
+            }
+            require(totalNumSelfControlled > 0, "SELF_AUTHENTICATION_IMPOSSIBLE");
+            if (totalNumSelfControlled >= numSelfControlledSignersRequired) {
+                require(numSelfControlled >= numSelfControlledSignersRequired, "NOT_ENOUGH_SIGNERS");
+            } else {
+                require(numSelfControlled == totalNumSelfControlled, "NOT_ENOUGH_SIGNERS");
```
dong77: can we simplify it as:

```
require(
        numSelfControlled >= numSelfControlledSignersRequired ||
        numSelfControlled == totalNumSelfControlled && totalNumSelfControlled > 0,
        "NOT_ENOUGH_SIGNERS"
);
```

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -73,45 +74,47 @@ contract GuardianModule is SecurityModule
...
     {
-        uint confirmStart = pendingAdditions[wallet][guardian];
+        uint confirmStart = pendingAdditions[wallet][guardian][info];
         require(confirmStart != 0, "NOT_PENDING");
         require(now > confirmStart && now < confirmStart + confirmPeriod, "EXPIRED");
-        controller.securityStore().addGuardian(wallet, guardian);
-        delete pendingAdditions[wallet][guardian];
-        emit GuardianAdded(wallet, guardian);
+        controller.securityStore().addGuardian(wallet, guardian, info);
+        delete pendingAdditions[wallet][guardian][info];
```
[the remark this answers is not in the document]
Brechtpd: A guardian can only be added a single time. I have added code to update the guardian. In the edge case a guardian is added with different `types` values it's still the wallet owner that needs to confirm it in a certain order so that seems okay for me.

packages/hebao_v1/contracts/stores/SecurityStore.sol
Brechtpd: When updating a guardian the value is currently just overwritten (as `|` can only add bits, which would be a bit strange).

packages/hebao_v1/contracts/stores/SecurityStore.sol
Brechtpd: I would definitely not do it in the storage (storage shouldn't need to care), and even in the module I wouldn't do it because it's valid and may be useful one day.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
Brechtpd: Yes! So in a the case with only 1 social guardian and only the wallet owner the single social guardian could start recovery, but the wallet owner can cancel it.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+  limitations under the License.
+*/
+pragma solidity ^0.5.11;
+
+import "./SecurityModule.sol";
+
+
+/// @title GuardianType
+/// @author Brecht Devos - <brecht@loopring.org>
+library GuardianType
```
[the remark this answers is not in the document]
Brechtpd: [Solidity doesn't allow assigning values to enums unfortunately](https://github.com/ethereum/solidity/issues/4638). The alternative would be to use an enum and do the bit shift every time or create a function for it and maybe wrap it again in new functions, but I wouldn't really call that an improvement compared to how it's done now really. :)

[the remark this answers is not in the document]
dong77: I see. Leave it as-is then.

packages/hebao_v1/contracts/stores/SecurityStore.sol
dong77: somewhere we may want to make it clear the types are overridden. And maybe we need to change `types` to `type` as there won't be more than one type.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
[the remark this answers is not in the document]
dong77: if we don't allow a guardian address to have more than one type, then we can use "==" to compare instead of "& TYPE != 0", then we don't need the types to have the specific shifted values then.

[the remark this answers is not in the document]
Brechtpd: I would allow more types, so things like `SelfControlled` + `Hardware` are possible. But a guardian can only be put in the list once. This way a guardian has only a single `types` value (with potentially multiple bits set) which allows for easy filtering.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -0,0 +1,176 @@
...
+            uint numFriends,
+            uint numFamily,
+            uint numSocialOther
+        )
+    {
+        for (uint i = 0; i < guardians.length; i++) {
+            if (guardians[i].types & GuardianType.SelfControlled() != 0) {
+                numSelfControlled++;
+            } else {
+                if (guardians[i].types & GuardianType.Friend() != 0) {
```
dong77: if we simply do `if (guardians[i].types == GuardianType.Friend())` can we use enum for types?

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -33,11 +33,11 @@ contract GuardianModule is SecurityModule
     uint public pendingPeriod;
     uint public confirmPeriod;
 
-    mapping (address => mapping(address => uint)) public pendingAdditions;
+    mapping (address => mapping(address => mapping(uint => uint))) public pendingAdditions;
     mapping (address => mapping(address => uint)) public pendingRemovals;
 
-    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint confirmAfter);
-    event GuardianAdded             (address indexed wallet, address indexed guardian);
+    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint types, uint confirmAfter);
```
dong77: recommend renaming 'types' to type as it will only be used to represent one type.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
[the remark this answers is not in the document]
dong77: ok

PR #832, opened 2019-12-07
[hebao] Better guardian rules

Brechtpd:
The current logic has a flaw in that a wallet could have many social guardians, but the wallet owner could always cancel a recovery of the wallet if the wallet didn't have any other self controlled guardians. 

This PR tries to fix that by also making the logic a bit easier by having the self controlled guardians also as a group, but I'm not sure yet if it works in all cases, so some additional changes are quite likely.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -65,7 +65,7 @@ library GuardianUtils
         bool walletOwnerSigned = false;
         SecurityStore.Guardian[] memory signingGuardians = new SecurityStore.Guardian[](signers.length);
         uint numGuardians = 0;
-        for (uint i = 0; i < 0; i++) {
+        for (uint i = 0; i < signers.length; i++) {
             if (signers[i] == walletOwner) {
                 walletOwnerSigned = true;
             } else {
@@ -89,6 +89,7 @@ library GuardianUtils
         if (numFriends + numFamily + numSocialOther == 0) {
             /* Self authentication */
             uint numSelfControlledSignersRequired = 2;
+
             // If the owner is allowed to sign:
             // - increase the total number of self controlled guardians
             // - increase the number of required self controlled guardians
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -112,8 +112,8 @@ library GuardianUtils
             numVotes += hasMajority(signed.numSelfControlled, total.numSelfControlled) ? 1 : 0;
         }
         if (total.numHardware > 0) {
-            totalNumVotes += total.numHardware;
-            numVotes += signed.numHardware;
+            totalNumVotes += 1;
+            numVotes += 1;
         }
 
         // We need a majority of votes
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -143,14 +143,12 @@ library GuardianUtils
                 groups.numHardware++;
             } else if (guardians[i].types & GuardianType.SelfControlled() != 0) {
                 groups.numSelfControlled++;
+            } else if (guardians[i].types & GuardianType.Friend() != 0) {
+                groups.numFriends++;
+            } else if (guardians[i].types & GuardianType.Family() != 0) {
+                groups.numFamily++;
             } else {
-                if (guardians[i].types & GuardianType.Friend() != 0) {
-                    groups.numFriends++;
-                } else if (guardians[i].types & GuardianType.Family() != 0) {
-                    groups.numFamily++;
-                } else {
-                    groups.numSocialOther++;
-                }
+                groups.numSocialOther++;
             }
         }
     }
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -44,6 +44,7 @@ library GuardianUtils
 
     struct GuardianGroupSizes
     {
+        uint numHardware;
         uint numSelfControlled;
         uint numFriends;
         uint numFamily;
@@ -87,46 +88,36 @@ library GuardianUtils
             require(!walletOwnerSigned, "WALLET_OWNER_SIGNATURE_NOT_ALLOWED");
         }
 
-        uint numGroups = 0;
-        uint numGroupsWithMajority = 0;
+        uint totalNumVotes = 0;
+        uint numVotes = 0;
         // Count the number of active groups and see which ones have a majority of signers
         if (requirement != SigRequirement.OwnerNotAllowed) {
-            numGroups++;
-            numGroupsWithMajority += walletOwnerSigned ? 1 : 0;
+            totalNumVotes += 1;
+            numVotes += walletOwnerSigned ? 1 : 0;
         }
         if (total.numFriends > 0) {
-            numGroups++;
-            numGroupsWithMajority += hasMajority(signed.numFriends, total.numFriends) ? 1 : 0;
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numFriends, total.numFriends) ? 1 : 0;
         }
         if (total.numFamily > 0) {
-            numGroups++;
-            numGroupsWithMajority += hasMajority(signed.numFamily, total.numFamily) ? 1 : 0;
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numFamily, total.numFamily) ? 1 : 0;
         }
         if (total.numSocialOther > 0) {
-            numGroups++;
-            numGroupsWithMajority += hasMajority(signed.numSocialOther, total.numSocialOther) ? 1 : 0;
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numSocialOther, total.numSocialOther) ? 1 : 0;
         }
         if (total.numSelfControlled > 0) {
-            numGroups++;
-            numGroupsWithMajority += hasMajority(signed.numSelfControlled, total.numSelfControlled) ? 1 : 0;
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numSelfControlled, total.numSelfControlled) ? 1 : 0;
+        }
+        if (total.numHardware > 0) {
+            totalNumVotes += total.numHardware;
+            numVotes += signed.numHardware;
         }
 
-        // Social authentication: Require a majority of groups to have a majority of signing guardians
-        // Or self authentication: When a wallet has enough self controlled guardians a majority
-        // of only the self controlled guardians (inlcuding the wallet owner when allowed) is enough.
-        if (!hasMajority(numGroupsWithMajority, numGroups)) {
-            // We need at least 3 self controlled guardians
-            require(total.numSelfControlled >= 3, "NOT_ENOUGH_SIGNERS");
-            // Count the wallet owner as a self controlled guardian
-            if (walletOwnerSigned) {
-                signed.numSelfControlled += 1;
-            }
-            // If the owner is allowed to sign increase the total number of self controlled guardians
-            if (requirement != SigRequirement.OwnerNotAllowed) {
-                total.numSelfControlled += 1;
-            }
-            require(hasMajority(signed.numSelfControlled, total.numSelfControlled), "NOT_ENOUGH_SIGNERS");
-        }
+        // We need a majority of votes
+        require(hasMajority(numVotes, totalNumVotes), "NOT_ENOUGH_SIGNERS");
     }
 
     function hasMajority(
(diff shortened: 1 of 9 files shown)


dong77:
I think we should not allow the owner to sign the cancelRecovery tx.

Brechtpd:
That could be a solution for for the cancellation problem, but I think there are other problems as well. For example, the same problem is true for `ApprovedTransfers`. If you only have social guardians (which may be true for a lot of people) a wallet owner could still do anything without any limits at any time (and not allowing the owner to sign for these doesn't make a lot of sense I think). So it seems like we need to strike a balance between social authentication and self authentication, where complete self authentication is only possible for people that have enough self controlled addresses (who are probably the more advanced users).

dong77:
I see. We need to makes sure the collective power of guardians are greater than owners so the guardians can reset the owner in case someone still the owner’s private key.

We can also treat self-controlled addresses and social guardians as two independent recovery approaches. And maybe in the future we can introduce another recovery module.

dong77:
@Brechtpd Just wondering if you are still working on this.

Brechtpd:
@dong77 I was indecisive about this so I left it to work on other things first. I'll come back to this and write down some cases so we can see what makes sense to do. The defi apps seemed more important to do first. In any case the change shouldn't be more than a small tweak to the current implementation.

Brechtpd:
- Minimum amount  of signers are given in the following cases.
- Cases mostly use a single guardian in each group for simplicity. This is of course the same as the majority in a group with multiple guardians.

WIP! Will see if more cases are needed.

#### `HasMajority` lookup table (total -> required number):
- 1 -> 1
- 2 -> 2
- 3 -> 2
- 4 -> 3

### 1. 1 guardian
- Start recovery: 1 guardian
- Cancel recovery: owner + 1 guardian
- Use `ApprovedTransfers`: owner + 1 guardian

-> A single **social** guardian in general is not really secure because a single person could recover your wallet.
-> A self controlled guardian is acceptable if it's a secure method.

### 2. 1 self + 1 social
- Start recovery: 1 self + 1 social
- Cancel recovery: 
  - owner + (1 self || 1 social)
  - 1 self + 1 social
- Use `ApprovedTransfers`: owner + (1 self || 1 social)

### 3. 1 family + 1 friend
- Start recovery: 1 family + 1 friend
- Cancel recovery: 
  - owner + (1 family || 1 friend)
  - 1 family + 1 friend
- Use `ApprovedTransfers`: owner + (1 family || 1 friend)

### 4. 1 self + 1 family + 1 friend
- Start recovery:
  - 1 self + 1 family
  - 1 self + 1 friend
  - 1 family + 1 friend
- Cancel recovery:
  - owner + 1 self + 1 family
  - owner + 1 self + 1 friend
  - owner + 1 family + 1 friend
  - 1 self + 1 family + 1 friend
- Use `ApprovedTransfers`: 
  - owner + 1 self + 1 family
  - owner + 1 self + 1 friend
  - owner + 1 family + 1 friend

### 5. 1 family + 1 friend + 1 misc
- Start recovery:
  - 1 family + 1 friend
  - 1 family + 1 misc
  - 1 friend + 1 misc
- Cancel recovery:
  - owner + 1 family + 1 friend
  - owner + 1 family + 1 misc
  - owner + 1 friend + 1 misc
  - 1 family + 1 family + 1 misc
- Use `ApprovedTransfers`: 
  - owner + 1 family + 1 friend
  - owner + 1 family + 1 misc
  - owner + 1 friend + 1 misc

### 6. 1 self + 1 family + 1 friend + 1 misc
- Start recovery:
  - 1 self + 1 family + 1 friend
  - 1 self + 1 family + 1 misc
  - 1 self + 1 friend + 1 misc
  - 1 family + 1 friend + 1 misc
- Cancel recovery:
  - owner + 1 self + 1 family + 1 friend
  - owner + 1 self + 1 family + 1 misc
  - owner + 1 self + 1 friend + 1 misc
  - 1 family + 1 friend + 1 misc
- Use `ApprovedTransfers`: 
  - owner + 1 self + 1 family + 1 friend
  - owner + 1 self + 1 family + 1 misc
  - owner + 1 self + 1 friend + 1 misc

### 7. 3 self + X social guardians
(We only look at self recovery here, see above for other cases)
- Start recovery: 3 self
- Cancel recovery:
  - owner + 2 self
  - 3 self
- Use `ApprovedTransfers`: 
  - owner + 2 self

LuciferMorning11:
6

Brechtpd:
@dong77 What do you think of the above cases? Do the minimum number of signers in the cases make sense?

dong77:
@Brechtpd give me a couple of days to think about it :)

dong77:
Here are my ideas:

First, we can have the following concepts:

- Guardians. A guardian is an address that can sign a transaction.
- Guardian Groups or groups. Each group can have 0, 1, or more guardians. Guardians in a group can collectively approve a transaction. Each group can have a vote value (or weight).  When a transaction is approved by a group, the transaction obtains N votes from that group. Note that N is not the number of guardians who signed the transaction. Each group can define how many guardians are required to sign a transaction.

We then have the following groups:

- owner group: This is a virtual group that has the current wallet owner. It has 1 vote. Approval requires the owner to sign.
- social family. It has 1 vote if the group has at least one guardian. Approval requires at least `M/2+1` guardians to sign.
- social friends. It has 1 vote if the group has at least one guardian. Approval requires at least `M/2+1` guardians to sign. 
- social others. It has 1 vote if the group has at least one guardian. Approval requires at least `M/2+1` guardians to sign.
- self-controlled: It has 2 votes if the group has at least one guardian. Approval requires at least 1 guardian to sign.

Now, each transaction can specify the following parameters regarding approval:
1. ownerSigining: (allow values are `required`, `optional`, `prohibited`). If required, the tx must have the owner group's approval; if prohibited, the owner group approval will be ignored.
2. number of votes: the number of votes required to the transaction, INCLUDING the owner group's vote if ownerSigning is either optional or required.

We can express the above as, for example, `SecurityPolicy(Required, 3)`, for ownerSigining == Required && number of votes == 3. `SecurityPolicy(Required)` means `SecurityPolicy(Required, 1)` or *owner only*.

User transactions do not know anything about guardians, only groups. In other words, guardians are entirely hidden by groups.

## Scenarios
- Start recovery: `SecurityPolicy(Prohibited, 2)`
- Cancel recovery: `SecurityPolicy(Prohibited, 2)`
- Approved Transfers: `SecurityPolicy(Required, 2)`
- Add address to whitelist: `SecurityPolicy(Required)`
- Add address to whitelist immediately: `SecurityPolicy(Required, 2)`

## What if there are not enough non-empty groups?
If not enough non-empty groups are set up, the corresponding transactions can simply fail. The alternative is to add another parameter to indicate whether all non-empty groups' approvals will be considered having enough votes.

```solidity
uint numNonEmptyGroups =...;
uint numVotes =...;
uint numApprovedGroups =...;
returns (numVotes >= requiredVotes || numApprovedGroups == numNonEmptyGroups && numNonEmptyGroups != 0);
```

Brechtpd:
Alright, not that many changes with the current implementation. If I didn't comment on it than it already works that way. :)

> self-controlled: It has 2 votes if the group has at least one guardian. Approval requires at least 1 guardian to sign.

This is different from the PR implementation:
- I find it hard to justify giving this group 2 votes. If a guardian is secured by a social account that is protected by the app by his phone (password protected/2FA/... enabled) than I think that's a better guardian than a self controlled guardian from the user's meta-mask address with private key stored on his PC.
- Only needing a fixed amount of signatures in this group also has some strange consequences. By adding more self-controlled guardians this group actually gets less secure! If a user adds 5 self controlled guardians you would think the system gets more secure. But the user now has 5 different sets of keys/passwords/etc... that he cannot lose. This means more ways to get hacked and security is reduced to the weakest self controlled guardian. This is why this PR also uses the majority rule for self controlled guardians.

> Now, each transaction can specify the following parameters regarding approval:
> 2. number of votes: the number of votes required to the transaction, INCLUDING the owner group's vote if ownerSigning is either optional or required.

I wonder if this is useful. Either something is critical security wise or it is not. So the owner can do it on his own or he cannot do it on his own. From a user's point of view this also makes sense. I can do it or I need to get a majority of my guardians for help. I can't really think of when this would be used (a feature is only a little bit dangerous?)

> Cancel recovery: SecurityPolicy(Prohibited, 2)

This is Allowed in this PR which I still think makes sense. I don't think there's a good reason to exclude the wallet owner, after all it's the wallet owner! Why wouldn't he have a vote on this? Looking at my cases above it also seems to work okay in practice. In practice, whether we allow him or not likely doesn't make any real difference. 

> If not enough non-empty groups are set up, the corresponding transactions can simply fail. The alternative is to add another parameter to indicate whether all non-empty groups' approvals will be considered having enough votes.

Like above with the self controlled guardians, I don't think it makes much sense to use any fixed numbers for this so this PR also uses the majority rule on groups. I think people will add guardians to increase flexibility and increase security.
In the worst case, if you don't have enough guardians you would not be allowed to so certain things. I would not be okay with that. If that's how the user uses the app than that should be okay. Maybe he's not planning to use the wallet with a lot of funds. To work around that people would probably add whatever guardian they can find, which could even lower the security of the wallet (as the owner keypair will be pretty well protected already, the other guardians people would add to unlock certain capabilities are likely worse).

I think for pretty much most people they could really care less how this guardian system works. The less they have to think about it the better. The method in this PR is already pretty complicated, so I wrote the cases down to make sure it actually makes sense if you don't know the details of the system. By adding more guardians you always increase security and flexibility. I think that that's an important property we need to have because people will not care about the details and will not read them (and to be honest, that would include me as I would assume it works in a certain way: more is better).

dong77:
Regarding self-controlled, originally I think these addresses should be hardware addresses. If they are just normal non-hardware addressed owned by the users, I agree they should follow the majority rule.

dong77:
There is a good reason NOT to allow wallet owner to cancel recovery. If the wallet owner's private-key is stolen, then the hacker will try to stop any recovery attempts to maintain control of the wallet. The rightful owner should contract his/her guardians via email or telephone and request a recovery, the hacker shall not be able to cancel the recovery.

dong77:
Just want to confirm: a transaction is approved when a majority of non-empty guardian-groups signed the transaction. Right?

Brechtpd:
> Maybe we should have a hardware-guardians group whose approval requires only one guardian to sign. The wallet mobile app has to make sure the guardian is indeed from a hardware wallet, not smart contracts.

In that case we can do the 1 hardware guardian signature to approve the self controlled group. Making an extra group may make things more complicated than needed.

> There is a good reason NOT to allow wallet owner to cancel recovery. If the wallet owner's private-key is stolen, then the hacker will try to stop any recovery attempts to maintain control of the wallet. The rightful owner should contract his/her guardians via email or telephone and request a recovery, the hacker shall not be able to cancel the recovery.

There is no case where only the wallet owner can cancel the recovery, guardians are always needed to do it. So a hacker needs to hack the wallet's guardians as well. I mainly see it for a wallet owner to more easily cancel a recovery attempt because either it's not needed any more or to protect himself better against social guardians (in case the wallet only has few). In case 3 for example, if the social guardian gets access to a self controlled guardian (e.g. a written down recovery code for a hardware wallet) the wallet owner can still cancel the recovery, if he's not allowed he cannot stop this. I think it's less likely a hacker can get access to multiple devices (and certainly not paper based ones) but people that you know do have more access to certain things. Either way, points can be made for both sides so I'm fine with either.

> Just want to confirm: a transaction is approved when a majority of non-empty guardian-groups signed the transaction. Right?

Correct!

dong77:
Alright, now the only outstanding issue (at least for me) is how to support hardware guardian, shall we add a flag to a guardian address?

Brechtpd:
I added a special handling for hardware guardians like this:
```
totalNumVotes += total.numHardware;
numVotes += signed.numHardware;
```

I removed the special logic for self authorization because as wallet owner + self controlled guardians + hardware guardians users already have a lot of flexibility to make sure they can do just that.

So conceptually every hardware guardian gets a group. 
- Initially I was planning to give a weight of 2 to each hardware guardian. I decided not to do that because that makes the social guardians have very little power. In that case what's the point of even having them. If a user sees no benefit in social guardians he can simply don't add any.
- This way in the rare case somebody add more than 1 hardware guardian security/flexibility is increased. The logic is also reduced to a simple majority in number of guardians which is what is basically used in other smart wallets.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -93,46 +88,36 @@ library GuardianUtils
...
+        if (total.numSocialOther > 0) {
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numSocialOther, total.numSocialOther) ? 1 : 0;
         }
+        if (total.numSelfControlled > 0) {
+            totalNumVotes += 1;
+            numVotes += hasMajority(signed.numSelfControlled, total.numSelfControlled) ? 1 : 0;
+        }
+        if (total.numHardware > 0) {
+            totalNumVotes += total.numHardware;
```
dong77: how about 
```
totalNumVotes += 1;
numVotes += 1;
```

dong77:
Approved, please merge.

After this PR, I think we should enforce each guardian to have one type, and use `==` to check the type instead of ` & TYPE != 0`. The current code gives the impression that an address can have multiple types, which doesn't make sense to me anymore and will make the approval logic more complicated.

PR #833, opened 2019-12-07
[hebao] improve PR from Steve

dong77:


letsgoustc:
Merged into previous patch

PR #837, opened 2019-12-08
[hebao] Add LRC Staking module

dong77:


packages/hebao_v1/contracts/modules/dapps/DAppModule.sol
```diff
@@ -0,0 +1,60 @@
...
+        returns (address[] memory signers)
+    {
+        bool found = false;
+        bytes4[] memory selectors = ownerMetaFunctions();
+
+        for (uint i = 0; i < selectors.length && !found; i++) {
+            if (method == selectors[i]) found = true;
+        }
+        require(found, "INVALID_METHOD");
+
```
Brechtpd: Do we need to do this? Just returning the owner as the signer seems good enough to me. The functions need to be guarded by `onlyFromMetaTx` or similar anyway.

packages/hebao_v1/contracts/iface/SubAccount.sol
```diff
@@ -24,12 +24,10 @@ contract SubAccount
     /// @dev Deposits Ether/token from the wallet to this sub-account.
     ///      The method must throw in case of error, or must emit a SubAccountTransfer event.
     /// @param wallet The wallt from which the Ether/token will be transfered out.
-    /// @param signers The list of meta-transaction signers, must be emptpy for normal transactions.
```
Brechtpd: I guess by removing signers here it could kind of limit the more riskier apps of making use of `SubAccount`. Though for ERC20 this could be solved quite nicely by not approving anything in the module itself and be counting on `approveToken` in `ApprovedTransfers` (but this doesn't work with ETH). Another possibility is by enforcing the quota. So there are definitely some options even in those (probably pretty rare) cases.

packages/hebao_v1/contracts/modules/dapps/LRCStakingModule.sol
```diff
@@ -17,15 +17,177 @@
...
+        require(
+            amount > 0 &&
+            amount <= tokenDepositable(wallet, token),
+            "INVALID_AMOUNT"
+        );
+
+        bytes memory txData = abi.encodeWithSelector(
+            ERC20(0).approve.selector,
+            address(stakingPool),
+            amount
```
Brechtpd: The safest thing to do here for dapps could be to use
```
uint amountToApprove = amount.add(ERC20(token).allowance(wallet, address(stakingPool));
```
or not setting the approval amount at all depending on the expected behaviour (use up the previously set approved amount or not). Current approvals (like the ones set in `approveToken` in `ApprovedTransfers`) would otherwise be overridden.

But it's unlikely that this would give any issues anyway so it probably doesn't matter.

packages/hebao_v1/contracts/iface/SubAccount.sol
dong77: The idea I have is that those dApps modules can only be accessed by the owners, not guardians or any other roles that a wallet may have. Therefore all meta-transactions must be signed by and only by owners.

packages/hebao_v1/contracts/modules/dapps/DAppModule.sol
dong77: You are right.

packages/hebao_v1/contracts/modules/dapps/LRCStakingModule.sol
dong77: I think it's OK.

packages/hebao_v1/contracts/modules/dapps/LRCStakingModule.sol
```diff
@@ -17,15 +17,177 @@
...
+        )
+        public
+        view
+        returns (uint depositable)
+    {
+        if (token != lrcTokenAddress) return 0;
+        else return super.tokenDepositable(wallet, token);
+    }
+
+    function ownerMetaFunctions()
```
Brechtpd: Can be removed now.

packages/hebao_v1/contracts/iface/SubAccount.sol
Brechtpd: The reason I think the signers would be helpful is for more security. There's a risk if the owner keypair gets stolen and the hacker can use it to extract all wallet funds without any limits just with the owner keypair. But probably all dapps we will support in the modules will have limited risks (and otherwise other solution can probably be found).

dong77: I'll think about it. Requiring other signers means great user experience problems.

PR #838, opened 2019-12-08
[hebao] PoolTogether module

Brechtpd:


packages/hebao_v1/contracts/modules/dapps/PoolTogether/IPool.sol
```diff
@@ -0,0 +1,39 @@
+// Created from the pooltogether codebase: https://github.com/pooltogether/pooltogether-contracts/blob/v2.x/contracts/BasePool.sol
```
dong77: move to `thirdparty`?

Brechtpd:
@dong77  Do I modify this PR just to add the interfaces needed to communicate with PoolTogether from a JS/... library? As the functionality can easily be implemented with `GenericDAppModule`?

PR #840, opened 2019-12-09
[hebao][WIP]Update compound to support borrow/repayBorrow

yueawang:

### DIFF
--- a/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
@@ -94,7 +94,8 @@ contract CompoundModule is BaseSubAccount, SecurityModule
 
         // mint new.
         if (mintAmount > 0) {
-            // TODO: check if we need exit market here.
+            // repay is done, exit market to get savings.
+            exitMarketsIfNeeded(wallet, cToken);
             mint(wallet, cToken, token, mintAmount);
             trackDeposit(wallet, token, mintAmount);
         }
@@ -354,7 +355,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         for (uint i = 0; i < tokens.length; ++i) {
             address cToken = compoundRegistry.getCToken(tokens[i]);
             if (cToken != address(0) && amounts[i] != 0) {
-                enterMarketIfNeeded(wallet, cToken);
+                enterMarketsIfNeeded(wallet, cToken);
                 mint(wallet, cToken, tokens[i], amounts[i]);
                 emit CollateralAdded(wallet, tokens[i], amounts[i]);
             }
@@ -381,7 +382,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
             address cToken = compoundRegistry.getCToken(tokens[i]);
             if (cToken != address(0)) {
                 redeemUnderlying(wallet, cToken, amounts[i]);
-                exitMarketIfNeeded(wallet, cToken);
+                exitMarketsIfNeeded(wallet, cToken);
             }
         }
 
@@ -465,7 +466,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         internal
     {
         // CErc20 and CEther have same function signature
-        enterMarketIfNeeded(_wallet, _cToken);
+        enterMarketsIfNeeded(_wallet, _cToken);
         transactCall(_wallet, _cToken, 0, abi.encodeWithSignature("borrow(uint256)", _amount));
     }
 
@@ -487,7 +488,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
     /// @dev Enters a cToken market if it was not entered before.
     /// @param wallet The target wallet.
     /// @param cToken The cToken contract.
-    function enterMarketIfNeeded(address wallet, address cToken) internal {
+    function enterMarketsIfNeeded(address wallet, address cToken) internal {
         if(!isCollaterlCToken(wallet, cToken)) {
             address[] memory market = new address[](1);
             market[0] = cToken;
@@ -498,7 +499,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
     /// @dev Exits a cToken market if there is no more collateral and debt.
     /// @param wallet The target wallet.
     /// @param cToken The cToken contract.
-    function exitMarketIfNeeded(address wallet, address cToken) internal {
+    function exitMarketsIfNeeded(address wallet, address cToken) internal {
         uint collateral = CToken(cToken).balanceOf(wallet);
         uint debt = CToken(cToken).borrowBalanceStored(wallet);
         if(collateral == 0 && debt == 0) {
--- a/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
@@ -455,6 +455,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         )
         internal
     {
+        // CErc20 and CEther have same function signature
         transactCall(_wallet, _cToken, 0, abi.encodeWithSignature("redeemUnderlying(uint256)", _amount));
     }
 
@@ -481,7 +482,14 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         transactCall(_wallet, _cToken, 0, abi.encodeWithSignature("repayBorrow(uint256)", _amount));
     }
 
-    function isCollaterlCToken(address wallet, address cToken) internal view returns (bool) {
+    function isCollaterlCToken(
+        address wallet,
+        address cToken
+        )
+        internal
+        view
+        returns (bool)
+    {
         return Comptroller(comptroller).checkMembership(wallet, CToken(cToken));
     }
 
--- a/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
@@ -30,8 +30,8 @@ import "../security/SecurityModule.sol";
 /// @title CompoundModule
 contract CompoundModule is BaseSubAccount, SecurityModule
 {
-    event CollateralAdded(address indexed _wallet, address _collateral, uint _collateralAmount);
-    event CollateralRemoved(address indexed _wallet, address _collateral, uint _collateralAmount);
+    event CollateralCTokenAdded(address indexed _wallet, address _collateral);
+    event CollateralCTokenRemoved(address indexed _wallet, address _collateral);
     event SubAccountBorrow(address indexed _wallet, address _debtToken, int _debtAmount);
 
     CompoundRegistry internal compoundRegistry;
@@ -50,9 +50,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         comptroller = _comptroller;
     }
 
-    /// @dev Fund Compound for earn interests and automatically enters market
-    ///      so the funds will be used as collateral; or return borrowed assets
-    ///      back to Compound.
+    /// @dev Fund Compound for earn interests if no borrowings or repay borrowings otherwise.
     function deposit(
         address            wallet,
         address[] calldata signers,
@@ -142,7 +140,6 @@ contract CompoundModule is BaseSubAccount, SecurityModule
                 borrow(wallet, cToken, borrowable);
                 trackBorrow(wallet, token, borrowable);
             }
-            return;
         }
 
         // Non-collateral token, user exchange current cToken and borrow more according to liquidity.
@@ -173,7 +170,8 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         }
     }
 
-    /// @dev tokenBalance is just the mapping from cToken to token, regardless of the collateral part.
+    /// @dev tokenBalance return positive number which is mapping from cToken to token if no loan, regardless of the collateral part,
+    ///      otherwise, return a minus token amount to indicate the loan.
     function tokenBalance (
         address wallet,
         address token
@@ -187,9 +185,14 @@ contract CompoundModule is BaseSubAccount, SecurityModule
             return 0;
         }
 
-        (uint err, uint cTokenSupply, , uint exchangeRateMantissa) = CToken(cToken).getAccountSnapshot(wallet);
+        (uint err, uint cTokenSupply, uint tokenBorrow, uint exchangeRateMantissa) = CToken(cToken).getAccountSnapshot(wallet);
         if (err != 0) {
             return 0;
+        }
+
+        // Return - loan if exist
+        if (tokenBorrow != 0) {
+            return - int(tokenBorrow);
         }
 
         int tokenSupply = int(cTokenSupply.mul(exchangeRateMantissa)) / (10 ** 18);
@@ -354,10 +357,11 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         // address[] memory cTokens = new address[](tokens.length);
         for (uint i = 0; i < tokens.length; ++i) {
             address cToken = compoundRegistry.getCToken(tokens[i]);
-            if (cToken != address(0) && amounts[i] != 0) {
+            if (cToken != address(0)) {
                 enterMarketsIfNeeded(wallet, cToken);
-                mint(wallet, cToken, tokens[i], amounts[i]);
-                emit CollateralAdded(wallet, tokens[i], amounts[i]);
+                if (amounts[i] != 0) {
+                    mint(wallet, cToken, tokens[i], amounts[i]);
+                }
             }
         }
 
@@ -501,6 +505,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
             address[] memory market = new address[](1);
             market[0] = cToken;
             transactCall(wallet, address(comptroller), 0, abi.encodeWithSignature("enterMarkets(address[])", market));
+            emit CollateralCTokenAdded(wallet, cToken);
         }
     }
 
@@ -512,6 +517,7 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         uint debt = CToken(cToken).borrowBalanceStored(wallet);
         if(collateral == 0 && debt == 0) {
             transactCall(wallet, address(comptroller), 0, abi.encodeWithSignature("exitMarket(address)", cToken));
+            emit CollateralCTokenRemoved(wallet, cToken);
         }
     }
 }
--- a/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
@@ -140,33 +140,33 @@ contract CompoundModule is BaseSubAccount, SecurityModule
                 borrow(wallet, cToken, borrowable);
                 trackBorrow(wallet, token, borrowable);
             }
-        }
-
-        // Non-collateral token, user exchange current cToken and borrow more according to liquidity.
-        // 2 cases:
-        //  a. amount <= balance                                  ==> redeem();
-        //  b. amount > balance && amount - balance <= borrowable ==> redeem() && borrow();
-        uint balance = CToken(cToken).balanceOfUnderlying(wallet);
-        uint borrowable = tokenBorrowable(wallet, token); // other collaterals tokens as this one is not.
-        require(amount <= balance + borrowable, "INVALID_WITHDRAW_AMOUNT");
-
-        uint withdrawAmount;
-        uint borrowAmount;
-        if (amount <= balance) {
-            withdrawAmount = amount;
         } else {
-            withdrawAmount = balance;
-            borrowAmount = amount - withdrawAmount;
-        }
+            // Non-collateral token, user exchange current cToken and borrow more according to liquidity.
+            // 2 cases:
+            //  a. amount <= balance                                  ==> redeem();
+            //  b. amount > balance && amount - balance <= borrowable ==> redeem() && borrow();
+            uint balance = CToken(cToken).balanceOfUnderlying(wallet);
+            uint borrowable = tokenBorrowable(wallet, token); // other collaterals tokens as this one is not.
+            require(amount <= balance + borrowable, "INVALID_WITHDRAW_AMOUNT");
 
-        if (withdrawAmount > 0) {
-            redeemUnderlying(wallet, cToken, withdrawAmount);
-            trackWithdrawal(wallet, token, amount);
-        }
+            uint withdrawAmount;
+            uint borrowAmount;
+            if (amount <= balance) {
+                withdrawAmount = amount;
+            } else {
+                withdrawAmount = balance;
+                borrowAmount = amount - withdrawAmount;
+            }
 
-        if (borrowAmount > 0) {
-            borrow(wallet, cToken, borrowAmount);
-            trackBorrow(wallet, token, borrowAmount);
+            if (withdrawAmount > 0) {
+                redeemUnderlying(wallet, cToken, withdrawAmount);
+                trackWithdrawal(wallet, token, amount);
+            }
+
+            if (borrowAmount > 0) {
+                borrow(wallet, cToken, borrowAmount);
+                trackBorrow(wallet, token, borrowAmount);
+            }
         }
     }
 
--- a/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
@@ -38,9 +38,6 @@ contract CompoundModule is BaseSubAccount, SecurityModule
     Comptroller      internal comptroller;
     address constant internal ETH_TOKEN_ADDRESS = address(0);
 
-    // wallet to ctoken borrow
-    mapping(address => address) internal borrowedCToken;
-
     constructor(
         Controller       _controller,
         CompoundRegistry _compoundRegistry,
@@ -79,27 +76,31 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         //  a. borrowed > 0 && amount <= borrowed ==> repayBorrow();
         //  b. borrowed > 0 && amount > borrowed  ==> repayBorrow() && deposit();
         //  c. borrowed == 0                      ==> deposit();
-        if (hasLoan(wallet)) {
-            if (hasCTokenLoan(wallet, cToken)) {
-                uint borrowed = CToken(cToken).borrowBalanceCurrent(wallet);
-                if (amount <= borrowed) {
-                    repayBorrow(wallet, token, amount);
-                    trackRepayBorrow(wallet, token, amount);
-                } else {
-                    repayBorrow(wallet, token, borrowed);
-                    trackRepayBorrow(wallet, token, borrowed);
-                    mint(wallet, cToken, token, amount - borrowed);
-                    trackDeposit(wallet, token, amount - borrowed);
-                }
-                return;
-            }
+        uint repayAmount;
+        uint mintAmount;
+        uint borrowed = CToken(cToken).borrowBalanceCurrent(wallet);
+        if (borrowed > amount) {
+            repayAmount = amount;
+        } else {
+            repayAmount = borrowed;
+            mintAmount = amount - borrowed;
         }
 
-        mint(wallet, cToken, token, amount);
-        trackDeposit(wallet, token, amount);
+        // repay borrowed amount if need
+        if (repayAmount > 0 ) {
+            repayBorrow(wallet, token, borrowed);
+            trackRepayBorrow(wallet, token, borrowed);
+        }
+
+        // mint new.
+        if (mintAmount > 0) {
+            // TODO: check if we need exit market here.
+            mint(wallet, cToken, token, mintAmount);
+            trackDeposit(wallet, token, mintAmount);
+        }
     }
 
-    /// @dev Redeem fund from Compound.
+    /// @dev Redeem or borrow fund from Compound.
     function withdraw(
         address            wallet,
         address[] calldata signers,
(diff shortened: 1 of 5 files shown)


Brechtpd:
I wonder if it wouldn't be a better idea to keep the normal depositing/withdrawing split from the borrowing/paying back the loan. I understand doing everything in `deposit`/`withdraw`, but that makes things more complex and more difficult to understand (and potentially dangerous because you could borrow just by withdrawing, which is unexpected). The UI will show different buttons for both operations as well so I think it makes sense to keep it separate in the smart contract:

```
contract ILending {
    function borrow(address wallet, address token, uint amount);
    function repayBorrow(address wallet, address token, uint amount);
}
````

To allow deposited funds to be used for borrowing as collateral (this doesn't seem to be needed for dxdy for example) we can use something like: 

```
contract ICollateral {
    function approveCollateral(address wallet, address token, uint amount);
}
````

For compound, `amount` would always need to be 0 or MAX_AMOUNT, because the approval is done on the token level (so for `amount == MAX_AMOUNT` `enterMarkers` is called, for `amount == 0` `exitMarket` is called). This PR mints/redeems tokens in `addCollaterals`/`removeCollaterals` which doesn't make a lot of sense to me (you may want to disable a token to be used as collateral, but that doesn't mean you want to redeem it). Either all funds deposited/token can be used or not as collateral. If a user wants to deposit + enable it as collateral both can be done in the same meta tx by calling `deposit` and `approveCollateral` as they are completely separate operations. A helper function on the smart contract could also be made, but this would make the interface less standard.

I don't think we need to make `SubAccount` support all defi functionality. Additional interfaces make a lot of sense. Not all `SubAccount` modules support every operation, having empty/default implementations doesn't always make it easier to work with. We can easily check if a module implements the `ILending` interface or not. If it does implement it we know the module supports lending, otherwise it doesn't. If a module has default functions for everything we cannot easily detect what functionality can actually be used or not. If a module implements `ILending` but not `ICollateral` we know that all deposits can automatically used for borrowing etc... Things like exchanges won't earn interest for users, so having a separate interface for `tokenInterestRate` makes sense as well (having a rate of 0 at some specific time does not mean no interest can be earned).

So basically I think most of the design used by Argent for defi apps makes a lot of sense. :)

**Comment by dong77** [comment]
The current design presents most DiFi modules as subaccounts to even further simplify data display and user interactions for users than Argent. When I designed SubAccount, I did not read any DiFi code just to make sure the SubAccount interface can provide a general enough interface for most of these dapps. 

Take `repayBorrow` as an example, if the current borrow balance is 1000, calling `deposit(token, 2000)` using the SubAccount design will automatically repay the borrowing of 1000 and add another 1000 to the pool; calling `withdraw(token, 400)` will borrow 400 more tokens. The UI will also have enough data to display proper information to the user before or after the transaction.

We can certainly support ILending/ICollateral/IInterestRate on SubAccount Modules to expose even more transactional functions, but I think even in these cases, the transfer/deposit function in the SubAccount design should be supported regardless.  From the user's perspective, most DeFi dapps are truly just sub accounts to earn more tokens and the main interactions users take are transferring assets to/from the dapps (unless the dapp is like gaming where the main interactions are playing the game inside the dapp).

**Comment by Brechtpd** [comment]
> The current design presents most DiFi modules as subaccounts to even further simplify data display and user interactions for users than Argent. When I designed SubAccount, I did not read any DiFi code just to make sure the SubAccount interface can provide a general enough interface for most of these dapps.

I think the smart contract design has a limited impact on the user experience because pretty much all interactions will be done using a UI anyway. Though of course important for the UI apps that need to interact with it.

> Take `repayBorrow` as an example, if the current borrow balance is 1000, calling `deposit(token, 2000)` using the SubAccount design will automatically repay the borrowing of 1000 and add another 1000 to the pool; calling `withdraw(token, 400)` will borrow 400 more tokens. The UI will also have enough data to display proper information to the user before or after the transaction.

The quite complex logic to this doesn't need to be implemented on the smart contract level to achieve this. The UI will need to know that the borrowed balance is 1000, so by depositing 2000 it needs to know that 1000 is used for repaying the debt and 1000 is used to deposit to the pool. So why not just do that and keep the smart contract much simpler (gas costs will be the same or even a bit lower because of the decreased complexity)?

> We can certainly support ILending/ICollateral/IInterestRate on SubAccount Modules to expose even more transactional functions, but I think even in these cases, the transfer/deposit function in the SubAccount design should be supported regardless. From the user's perspective, most DeFi dapps are truly just sub accounts to earn more tokens and the main interactions users take are transferring assets to/from the dapps (unless the dapp is like gaming where the main interactions are playing the game inside the dapp).

I agree, `deposit`/`withdraw` makes sense for all these kind of apps.

> For the Compound module, because we target users who don't really know much about finance and technology, we'd better hide the approveCollateral function entirely and let them know all their deposits are approved for collateral automatically.

I would actually argue the opposite because of the same reasons. :) I would not approve any token to be used for collateral as this is much too complex for normal users that likely just want to use compound to earn interest (especially with the current design! If the user would want to withdraw directly on the module contract he could accidentally start borrowing simply by withdrawing more than the actual token balance!) If you do have enough understanding and do want to borrow than I think being able to just set certain tokens as collateral would be nice to have (and not having only a ENABLE BORROWING button). You don't want to have a token you're saving for the long term to be used for liquidation.

I think the above also kind of shows my point. The module smart contract can be very powerful while still being very simple if we don't write all the logic we think we may need in the smart contract. It's not because we have an `approveCollateral` users need to know about it. If the UI makes it clear everything will be used for collateral than we simply call `approveCollateral` each time the user deposits a new token that hasn't been enabled yet. There's no need for each user action to map to a single smart contract function. I think in general most of the code in modules should simply wrap the external dapp calls in the standard interface functions and emit standard events.

**Comment by dong77** [comment]
@Brechtpd feel free to optimize the overall SubAccount/Dapp module design :)

**Comment by Brechtpd** [comment]
@dong77 I think the current design and use in other modules is very good, it's just the borrowing in this PR I think that doesn't fit in nicely with the existing interface. I will modify this PR tomorrow and see what comes out of it. :)

**Comment by dong77** [comment]
@brecht, There is another way to support third-party dapps, I'll draft a PR for it. Maybe you'll like it.

**Comment by Brechtpd** [comment]
Any third party dapp is actually supported because we can do `callContract` and call any contract we like. If the dapp contract is also whitelisted we're very close to these specific dapp modules. This takes the complete integration off-chain which actually would be very nice I think. If we maintain a whitelist of safe dapp addresses that the wallet owner can whitelist himself we're even closer to the modules.

File: packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol [hunk 1]
```diff
@@ -158,6 +379,72 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         internal
     {
         // CErc20 and CEther have same function signature
+        transactCall(_wallet, _cToken, 0, abi.encodeWithSignature("redeem(uint256)", _amount));
```
dong77: if possible, we should always use `encodeWithSelector` instead of `encodeWithSignature`.
Here you can use 
```solidity
encodeWithSelector(Comptroller(0).redeem.selector, _amount)
```

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -158,6 +379,72 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+    function isCollaterlCToken(address wallet, address cToken) internal view returns (bool) {
+        return Comptroller(comptroller).checkMembership(wallet, CToken(cToken));
+    }
+
+    /// @dev Enters a cToken market if it was not entered before.
+    /// @param wallet The target wallet.
+    /// @param cToken The cToken contract.
+    function enterMarketIfNeeded(address wallet, address cToken) internal {
+        if(!isCollaterlCToken(wallet, cToken)) {
+            address[] memory market = new address[](1);
```
dong77: market => markets, even though it only has one element

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
dong77: Also, when a function has more than 3 inputs, always break the inputs into multiple lines.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -64,6 +75,26 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+                } else {
+                    repayBorrow(wallet, token, borrowed);
+                    trackRepayBorrow(wallet, token, borrowed);
+                    mint(wallet, cToken, token, amount - borrowed);
+                    trackDeposit(wallet, token, amount - borrowed);
+                }
+                return;
+            }
+        }
+
```
dong77: I'm not sure the last `mint` call should happen in all cases.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -22,24 +22,35 @@ import "../../base/BaseSubAccount.sol";
...
+    event CollateralAdded(address indexed _wallet, address _collateral, uint _collateralAmount);
+    event CollateralRemoved(address indexed _wallet, address _collateral, uint _collateralAmount);
+    event SubAccountBorrow(address indexed _wallet, address _debtToken, int _debtAmount);
+
     CompoundRegistry internal compoundRegistry;
+    Comptroller      internal comptroller;
     address constant internal ETH_TOKEN_ADDRESS = address(0);
 
+    // wallet to ctoken borrow
+    mapping(address => address) internal borrowedCToken;
```
dong77: We should really remove this `borrowedCToken`  and relay on Compound for all states.
And the current design is also not right to me , how can a wallet address only has one CToken address?

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -64,6 +75,26 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         address cToken = compoundRegistry.getCToken(token);
         require(cToken != address(0), "NO_MARKET");
 
+        // 3 cases:
+        //  a. borrowed > 0 && amount <= borrowed ==> repayBorrow();
+        //  b. borrowed > 0 && amount > borrowed  ==> repayBorrow() && deposit();
+        //  c. borrowed == 0                      ==> deposit();
+        if (hasLoan(wallet)) {
+            if (hasCTokenLoan(wallet, cToken)) {
+                uint borrowed = CToken(cToken).borrowBalanceCurrent(wallet);
```
dong77: Here, we should remove the two if-statements and do the following:

``` solidity
require(amount <= tokenDepositable(token))
enterMarketAndEnableCollateralIfNecessary(token)
trackDeposit(amount)

uint repayAmount;
uint mintAmount;
uint borrowed = CToken(cToken).borrowBalanceCurrent(wallet);
if (borrowed > amount) {
    repalyAmount = amount;
 } else {
    repayAmount = borrowed;
    mintAmount = amount - borrowed;
}
if (repayAmount >0 ) repay(repayAmount);
if (mintAmount > 0) mint(mintAmount);
```

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -85,13 +116,49 @@ contract CompoundModule is BaseSubAccount, SecurityModule
         address cToken = compoundRegistry.getCToken(token);
         require(cToken != address(0), "NO_MARKET");
 
-        uint balance = uint(tokenBalance(wallet, token));
-        require(amount > 0 && amount <= balance, "INVALID_WITHDRAW_AMOUNT");
+        require(amount > 0, "ZERO_WITHDRAW_AMOUNT");
 
-        redeem(wallet, cToken, amount);
-        trackWithdrawal(wallet, token, amount);
+        if (hasLoan(wallet)) {
```
dong77:
```solidity
enterMarketAndEnableCollateralIfNecessary(token)
trackWithdrawal(amount)

uint withdrawAmount;
uint borrowAmount;
uint withdrawable = tokenWithdrawable(token)
if (withdrawable >= amount) {
    withdrawAmount = amount;
} else {
    withdrawAmount = withdrawable;
    borrowAmount = amount - withdrawable;
}
if (withdrawAmount > 0) claim(...)
if (borrowAmount > 0) borrow(...)

File: packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol [hunk 7]
```diff
@@ -105,10 +172,22 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
 
-        uint amount = CToken(cToken).balanceOf(wallet);
-        uint exchangeRateMantissa = CToken(cToken).exchangeRateStored();
-        uint tokenValue = amount.mul(exchangeRateMantissa) / (10 ** 18);
-        return int(tokenValue);
+        if (hasLoan(wallet)) {
+            if (hasCTokenLoan(wallet, cToken)) {
+                return - int(CToken(cToken).borrowBalanceStored(wallet));
+            } else if (isCollaterlCToken(wallet, cToken)) {
+                // if cToken is in collateral market of compound, no withdraw allowed.
```
dong77: tokenBalance()返回的是余额，不是可提现金额。可提现金额在tokenWithdrawalable返回。

dong77:
@yueawang Questions regarding compound (@Brechtpd please also comment)

1. If an address has some LRC used as collateral (for borrowing some other token), can LRC be withdrawn at all? If so, how to calculate the amount of LRC that can be withdrawn?

2. Is it possible that an address has some LRC being used as collateral and in the meanwhile has some borrowed LRC (negative balance)?

packages/hebao_v1/contracts/thirdparty/compound/ComptrollerStorage.sol
```diff
@@ -0,0 +1,74 @@
+// From Compound code base - https://github.com/compound-finance/compound-protocol/blob/master/contracts/ComptrollerStorage.sol
+// with minor modificaiton.
```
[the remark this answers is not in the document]
yueawang: 因为不是全部的ComptrollerStorage，去掉了一些无用的定义，api什么的，如果不改，那所有的import都要拿过来

packages/hebao_v1/contracts/thirdparty/compound/ComptrollerPriceOracle.sol
```diff
@@ -0,0 +1,22 @@
+// From Compound code base - https://github.com/compound-finance/compound-protocol/blob/master/contracts/PriceOracle.sol
+// with minor modificaiton.
```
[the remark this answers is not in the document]
yueawang: 文件名改了，因为有同名文件存在
另外把solidity版本改了

packages/hebao_v1/contracts/thirdparty/compound/Comptroller.sol
```diff
@@ -0,0 +1,60 @@
+// From Compound code base - https://github.com/compound-finance/compound-protocol/blob/master/contracts/Comptroller.sol
+// with minor modificaiton.
```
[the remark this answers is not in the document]
yueawang: 去掉了无用的函数和引用，变量，事件定义等等

packages/hebao_v1/contracts/thirdparty/compound/CToken.sol
```diff
@@ -3,15 +3,102 @@
...
-    function underlying() external view returns (address);
-    function symbol() external view returns (string memory);
-    function supplyRatePerBlock() external view returns (uint);
+
+/**
+ * @title Compound's CToken Contract
+ * @notice Abstract base for CTokens
+ * @author Compound
+ */
+contract CToken {
```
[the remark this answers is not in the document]
yueawang: 之前的master是基于2.2release的，当时里面还没有这个文件

packages/hebao_v1/contracts/thirdparty/compound/Comptroller.sol
```diff
@@ -0,0 +1,60 @@
...
+
+import "./CToken.sol";
+import "./ComptrollerStorage.sol";
+
+
+/**
+ * @title Compound's Comptroller Contract
+ * @author Compound
+ */
+contract Comptroller is ComptrollerV2Storage {
```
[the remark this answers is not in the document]
yueawang: 目前不行，因为ComptrollerInterface.sol没有计算抵押贷款额度的函数getAccountLiquidity，如果在hebao合约内部计算额度必须通过这个函数。

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: 这里是为了确定那个是借贷的token，只能有一个。对于Compound来说，一个token对应一个cToken，可以抵押不同的cToken，但不支持同时多个Token借贷，就是比如抵押cLRC和cDAI借ETH可以，但还完ETH之前不能借其他的。
BTW：argent注释里提到Compound不支持同时多币种借贷。但是他那边类似的处理应该是在app里面。

yueawang:
Based on my understanding.
1. If an address has some LRC used as collateral (for borrowing some other token), can LRC be withdrawn at all? If so, how to calculate the amount of LRC that can be withdrawn?
In Compound, NO, what you really have is cLRC rather than LRC since you used LRC as collateral. You may try transfer a part of cLRC to some others and Compound has internal logic to check if your liquidation is still valid, and if not, the transfer does not happen. 
If you want to calculate exact amount of LRC you can get, first, you query your cLRC Liquidity in Compound, and use Compound's price oracle to get the price of both cLRC and token you borrow(say ETH here), then, use same loan calculation to get the amount you can withdraw.
For example, say 1ETH == 100 LRC, 1 LRC == 100 cLRC, and Compound's collterals factor is 0.5.
you have 200 LRC, give Compound and get 20000 cLRC as collateral, so now you can borrow 1ETH = 100 LRC == 20000 cLRC * 0.5. Now If you just borrow 0.5 ETH i.e. use 10000 cLRC as collterals, in theory, you can withdraw the rest 10000 cLRC back. Here we assume price is stable, otherwise, such withdraw may lead to liquidation 

2. Is it possible that an address has some LRC being used as collateral and in the meanwhile has some borrowed LRC (negative balance)?
In theory it can, but the balance is normally not negative because of the collterals factor normally < 1.
say you get 0.5~0.7LRC with 1LRC as collaterals(actually it is mint to 100 cLRC)

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
Brechtpd: I tested it out and you can supply multiple tokens to increase collateral, and you can also borrow multiple tokens at once. I think the argent you mentioned is the one that mentions you can't open multiple loans. I think that means you can't open multiple loans for the **same** token, but you can open a loan for each token.

Brechtpd:
Are we still going to implement this logic in a smart contract? I thought we were just going to use `GenericDAppModule` for this (without any changes) because that's all we need on-chain. That's why I didn't give any further comments on this PR.

We still have to do most of the logic done in this smart contract off-chain of course, but we're not constrained by what we can do in a smart contract.

dong77:
I think we should implement SubAccount for all DiFi dApps, I think for non-DiFi dApps we should simply use GenericDAppModule.

As a user, I want to use Compound, but I don't want to know any details. I just want to move funds around.

Implementing Compound as a SubAccount module will also make mobile app development much easier.

dong77:
It seems we cannot implement SubAccount as Compound doesn't provide functions to query the amount of tokens that can be redeemed from a market - it only provides methods to check if a certain amount of redemption is allowed or not.

Brechtpd:
> As a user, I want to use Compound, but I don't want to know any details. I just want to move funds around.

I don't see why this makes a difference for the UI. The UI will be the same, it will just work slightly different behind the scenes. It's not like users will go to Etherscan to use the module directly.
The UI that already exist for these apps just use the interface directly. If some extra logic is done in JS or solidity doesn't seem to make a difference for the user.

> Implementing Compound as a SubAccount module will also make mobile app development much easier.

Using solidity as the universal language to make the logic platform independently is perhaps a bit unorthodox, but could certainly be a good reason. :)

But to make this work we need to make sure the interface exposes all data necessary for the UI to show to users. Otherwise we're still in the case we're doing the logic on-chain and off-chain. The borrowing makes things more complicated with things like
- How much can I borrow?
- How much collateral do I have (and in a which tokens)?
- When will I get liquidated?
- What's the collateral factor (as in, if I deposit X amount of a token as collateral, how much more will I be able to borrow)?

These are also important to know for the user and the UI (like only allowing sensible max amounts for users).

Brechtpd:
Because cTokens can freely be transferred between accounts `tokenReturn` and `tokenReturnAmount` can be incorrect. These use the tracked deposits and withdrawal amounts (in the underlying token), but a user can transfer cTokens in and out from/to other accounts (you can even buy/sell cDAI on uniswap) even when these cTokens are used as collateral (but you cannot transfer an amount that would liquidate yourself immediately). This also means that for the collateral/token we have to check the balance of the cToken directly and see if the user has entered the market, we cannot depend on only the events in the module to help keep track of the funds the user has in the Compound system.

> Implementing Compound as a SubAccount module will also make mobile app development much easier.

Maybe an interesting idea: We can do the integration with defi apps in a smart contract but not in the module itself, just in a helper smart contract with only view methods.  `SubAccount` is mainly view methods that don't depend on any data in the module so it's not really necessary to do these in the module itself. We could even add view methods that convert high level function calls (like `deposit`) to data that can be used directly in GenericDAppModule. This way the users have a module that is future proof and we can update the helper smart contract anytime we like and do seamless updates which allows us to gradually roll features out/update the integration without the user having to pay/approve module updates (and losing any data stored in the module itself. In any case, it seems best to avoid storing data in the module itself in general to avoid problems like this).

yueawang:
> But to make this work we need to make sure the interface exposes all data necessary for the UI to show to users. Otherwise we're still in the case we're doing the logic on-chain and off-chain. The borrowing makes things more complicated with things like
How much can I borrow?
How much collateral do I have (and in a which tokens)?
When will I get liquidated?
What's the collateral factor (as in, if I deposit X amount of a token as collateral, how much more will I be able to borrow)?
These are also important to know for the user and the UI (like only allowing sensible max amounts for users).

Exactly! Compound did not provide a good enough abstraction of the account model as ours, thus what we need to do is in time mapping from all mint(for deposit) and/or collaterals(for loan) ctokens back to token amount. In user/UI's point of view, it sometimes confuses as it changes either by time (price oracle changes) or by withdrawing of other tokens (account liquidity changes). And also because we need to report the amount of token balance/withdrawable, we actually re-implemented(copy for the accuracy) the same internal liquidity calculation of Compound.

BTW: Argent does not calculate the token amount in the contract, for invest(deposit) part, it uses the fraction so no exact amount is needed, for loan part, I guess it calculates the amount in its app since a few view APIs can give out an approximate value.

yueawang:
Talked with @letsgoustc and took Argent and InstaApp as the reference.
I will update this PR soon with the following assumptions:
1. Only ETH and DAI can be borrowed, and a user can only borrow 1 of them.
2. No collateral token loan, i.e. deposit ETH and borrow ETH, that's meaningless.
3. for cToken come from mint(deposit) or other users(transfer), tokenBalance is only the mapping from cToken to Token with the latest price, regardless of how many are collaterals. On the other hand, tokenBalance should be the loan and would be recorded in CompoundSubAccount if borrowing happens, and App decides how to display that later.
4. The withdrawable amount of a token is as much as possible until liquidation happens. therefore, each withdrawing changes the user's liquidity and influences the withdrawable amount of all other tokens.
5. App watches the cToken balance change, especially the token transfer, to avoid liquidation, App can achieve that by comparing the transfer amount and withdrawable.

Any suggestion? @dong77 @Brechtpd @letsgoustc

letsgoustc:
> Talked with @letsgoustc and took Argent and InstaApp as the reference.
> I will update this PR soon with the following assumptions:
> 
> 1. Only ETH and DAI can be borrowed, and a user can only borrow 1 of them.
User may borrow ETH and DAI at the same time. Normally user borrow ETH for short, while borrow DAI for long.
> 2. No collateral token loan, i.e. deposit ETH and borrow ETH, that's meaningless.
> 3. for cToken come from mint(deposit) or other users(transfer), tokenBalance is only the mapping from cToken to Token with the latest price, regardless of how many are collaterals. On the other hand, tokenBalance should be the loan and would be recorded in CompoundSubAccount if borrowing happens, and App decides how to display that later.
> 4. The withdrawable amount of a token is as much as possible until liquidation happens. therefore, each withdrawing changes the user's liquidity and influences the withdrawable amount of all other tokens.
> 5. App watches the cToken balance change, especially the token transfer, to avoid liquidation, App can achieve that by comparing the transfer amount and withdrawable.
> 
> Any suggestion? @dong77 @Brechtpd @letsgoustc

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: Yes, Compound does not limit the loan token, I thought by mistake it was Argent's design.
I talked with Steve to see if we need to limit our loan token to ETH & DAI only, any opinion on that?

yueawang:
I also think under current framework we can split tokens into 2 categories: collateral and non-collateral.
The main difference is in withdraw, for those collateral tokens, withdraw is done according to user liquidity calculation i.e. borrow only, regardless of the cToken amount. On the other hand, non-collateral tokens go to redeem path and can be borrowed as the collateral token if the user asks more. Then tokenBalance() and tokenWithdrawable() is more clear based on this assumption.

Brechtpd:
> BTW: Argent does not calculate the token amount in the contract, for invest(deposit) part, it uses the fraction so no exact amount is needed, for loan part, I guess it calculates the amount in its app since a few view APIs can give out an approximate value.

Argent almost doesn't do any logic in their modules, they just wrap the basic building blocks needed on-chain (which is important for them because they can only do a single function call/meta tx, we don't have that limitation). An integration with Compound is certainly not trivial and it doesn't seem to be designed to do everything on-chain (which also makes sense, if you can keep things off-chain it's almost always better). For example, for the app to know a user has been liquidated I think it can only listen to the `LiquidateBorrow` event, I don't think we can find that out on-chain in a different way (and this is another way the cToken balance of the user can be changed). So some Compound specific logic off-chain seems inevitable at some point.

> I also think under current framework we can split tokens into 2 categories: collateral and non-collateral.

Are all tokens enabled as collateral by default? Or do users need to enable which tokens can be used as collateral?

> 1. Only ETH and DAI can be borrowed

Is this just what will be possible in the UI? Or will this be enforced in the module? If enforced in the module, is there a reason to handle these tokens differently?

> 5. App watches the cToken balance change, especially the token transfer, to avoid liquidation, App can achieve that by comparing the transfer amount and withdrawable.

Compound will not allow the transfer if the user would liquidate himself, so luckily I think the app doesn't need to check the transfers done by the user (as this would be hard to do).

yueawang:
> Are all tokens enabled as collateral by default? Or do users need to enable which tokens can be used as collateral? Or are just ETH and DAI always enabled by the module (which would explain the point below)?

No, Compound needs the user to enter/exit some markets for their collateral tokens, so do we. There are no default collateral tokens, users need to select their own. 
For points 1 & 2 My initial thought was to forbid people to borrow ETH while holding cETH which was mint by previously deposited ETH as collaterals, I thought that is meaningless, But later I thought that assumption is impossible, as you can not forbid others to transfer a few cETH to you which potentially increases your liquidity. That's why I think split tokens to collateral & non-collaterals is better, as you can easily calculate the withdrawable amount of them respectively, i.e, for collaterals, based on liquidity, for non-collaterals, based on cToken balance. One more thing that needs to be dealt with is to switch collateral and non-collaterals while user deposit & withdraw.

> Is this just what will be possible in the UI? Or will this be enforced in the module? If enforced in the module, is there a reason to handle these tokens differently? 
Now, I think no limitation is applied here anymore as I mentioned above. Maybe it can be one of our product design if necessary.

Brechtpd:
> No, Compound needs the user to enter/exit some markets for their collateral tokens, so do we. There are no default collateral tokens, users need to select their own.

>  One more thing that needs to be dealt with is to switch collateral and non-collaterals while user deposit & withdraw.

I still don't fully understand how this will work. If I understand Compound correctly if you enter a market you enable the token to be used as collateral AND enable lending (and you can't decide how much can be used for collateral, it's either all you cToken or none). So will there be a separate function to enable the token to be used as collateral/enable borrowing it (enter market) or not? If a user deposits some tokens you can't automatically now if he wants to use these for collateral or not, with withdrawing I understand that you could automatically enter the market to enable borrowing (and exit the market when the the loan is repaid, but maybe the user wants to keep the token enabled for collateral?). So it will be something like this?

> For points 1 & 2 My initial thought was to forbid people to borrow ETH while holding cETH which was mint by previously deposited ETH as collaterals, I thought that is meaningless, But later I thought that assumption is impossible, as you can not forbid others to transfer a few cETH to you which potentially increases your liquidity.

Oh right, I didn't think of this yet. Do you know if you redeem the cETH does it automatically repay your debt or not? I'm guessing it does not, but I don't know for sure.

> That's why I think split tokens to collateral & non-collaterals is better, as you can easily calculate the withdrawable amount of them respectively, i.e, for collaterals, based on liquidity, for non-collaterals, based on cToken balance.

What do you mean by split tokens? Just checking to see if the token is in a market onchain and then deciding how to do the calculations to check the withdrawable amount? But whether it's in the market or not is left to the user right? If so than that makes sense.

In your cETH example, if you entered the market for ETH, the extra cETH transferred in will also automatically be used as collateral. If you aren't entered in the market for cETH your balance will just have increased (same as when entered in the market). You can't have cETH that is used for collateral and cETH that is not used for collateral as far as I know. Either your complete cETH balance is used or not. I think that's also what you're saying, just making sure. :)

yueawang:
> I still don't fully understand how this will work. If I understand Compound correctly if you enter a market you enable the token to be used as collateral AND enable lending (and you can't decide how much can be used for collateral, it's either all you cToken or none). So will there be a separate function to enable the token to be used as collateral/enable borrowing it (enter market) or not?

You are right. The user decides what tokens are collateral on his demand, and Compound provided such enter/exit API pairs, which hebao/Argent wrapped as add/removeCollaterals functions. 
That thing I mentioned is a special case in our hebao implementation, because deposit sometimes involve repay and withdraw includes borrow. Take withdraw as an example, when a user has cToken (say cLRC), he will meet following options when he withdraws
1. user gets LRC redeemed from cLRC, if cLRC is non-collateral, this is the easiest case.
2. user gets LRC redeemed from cLRC, as long as cLRC is non-collateral, and then App(maybe) changes cLRC to collateral type (because the user needs to enter LRC market) and borrow a few more to meet the user's request.
3. user borrows LRC, if that cLRC is collateral, and the users liquidity still has space.
Otherwise, there is borrowing only, as user does not have cLRC but may have other collaterals.

So, I think in option 2, App needs to help (popup a hint?) the user to make the cToken collateral.
In deposit function, we have similar cases, and App needs to change collateral token back to non-collateral when possible.

> Do you know if you redeem the cETH does it automatically repay your debt or not? I'm guessing it does not, but I don't know for sure.

No, it does not. Redeeming cETH gives user account ETH, and if a user wants to repay borrowing, he needs to call repay() to send those ETH back to Compound. That's 2 different paths, I think others who integrated Compound inherit these 2 paths. Here our hebao is different because it must combine these 2 together to fit account abstraction. I think use 2 paths is easier here, and leave App to organize the view as SubAccount.

> What do you mean by split tokens? Just checking to see if the token is in a market onchain and then deciding how to do the calculations to check the withdrawable amount? But whether it's in the market or not is left to the user right? If so than that makes sense.

Yes, I mean treat them like 2 groups. The user decides what are collaterals.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: Here CErc20 and CETH has the same signature. If using encodeWithSelector, we may add an if here, is that OK?

```
if (_token == ETH_TOKEN_ADDRESS) {
    transactCall(_wallet, _cToken, 0, abi.encodeWithSelector(CEther(0).redeem.selector, _amount));
} else {
    transactCall(_wallet, _cToken, 0, abi. encodeWithSelector(CErc20(0).redeem.selector, _amount));
}
```

packages/hebao_v1/contracts/thirdparty/compound/ComptrollerPriceOracle.sol
```diff
@@ -0,0 +1,22 @@
...
+pragma solidity ^0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "./CToken.sol";
+
+
+interface ComptrollerPriceOracle {
```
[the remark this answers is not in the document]
yueawang: We already have a PriceOracle.sol in our project, I think they can not coexist ?

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -133,7 +218,209 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+        nonReentrant
+        onlyFromMetaTxOrWalletOwner(wallet)
+        returns (uint)
+    {
+        require(tokens.length > 0, "EMPTY_TOKEN_ARRAY");
+
+        // address[] memory cTokens = new address[](tokens.length);
+        for (uint i = 0; i < tokens.length; ++i) {
+            address cToken = compoundRegistry.getCToken(tokens[i]);
+            if (cToken != address(0) && amounts[i] != 0) {
```
Brechtpd: I think we should allow `amounts[i] == 0` so that `enterMarketsIfNeeded` is called for the token. Otherwise if the user first used `deposit` to get cTokens there would otherwise be no way to use that existing balance as collateral without redeeming and calling `addCollaterals` again with the redeemed tokens, which is much more inefficient.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -22,24 +22,32 @@ import "../../base/BaseSubAccount.sol";
...
+        Comptroller      _comptroller
         )
         public
         SecurityModule(_controller)
     {
         compoundRegistry = _compoundRegistry;
+        comptroller = _comptroller;
     }
 
     /// @dev Fund Compound for earn interests and automatically enters market
```
Brechtpd: I think this comment is wrong as the market is only exited when there is no loan for the token and balanceOf(cToken) != 0. The market is never entered.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -85,13 +118,62 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+                trackWithdrawal(wallet, token, redeemable);
+            }
 
-        redeem(wallet, cToken, amount);
-        trackWithdrawal(wallet, token, amount);
+            if (borrowable > 0) {
+                borrow(wallet, cToken, borrowable);
+                trackBorrow(wallet, token, borrowable);
+            }
+            return;
```
Brechtpd: It's a bit confusing there's a `return` here. Can you make it an if/else instead?

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -85,13 +118,62 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+            trackWithdrawal(wallet, token, amount);
+        }
+
+        if (borrowAmount > 0) {
+            borrow(wallet, cToken, borrowAmount);
+            trackBorrow(wallet, token, borrowAmount);
+        }
     }
 
+    /// @dev tokenBalance is just the mapping from cToken to token, regardless of the collateral part.
```
Brechtpd: Shouldn't this return a negative number when the user has a loan open?

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: Fixed

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: you right, fixed.

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
yueawang: fixed

packages/hebao_v1/contracts/modules/dapps/CompoundModule.sol
```diff
@@ -85,13 +116,62 @@ contract CompoundModule is BaseSubAccount, SecurityModule
...
+                trackWithdrawal(wallet, token, redeemable);
+            }
 
-        redeem(wallet, cToken, amount);
-        trackWithdrawal(wallet, token, amount);
+            if (borrowable > 0) {
+                borrow(wallet, cToken, borrowable);
+                trackBorrow(wallet, token, borrowable);
+            }
+        }
```
Brechtpd: You removed the `return` here, but didn't put the code below in `else`. :)

yueawang: Oh, I misunderstood the comment, I thought I added a redundant return :(
Fix soon, Thanks!

PR #848, opened 2019-12-11
[hebao] add GenericDAppModule - one dapp per module

dong77:


dong77:
One thing I'm not sure is whether we should support many dapps in one single module or one module per dapp. The first approach requires an owner which is not as secure; the second approach is secure but needs extra deployment and user transactions.

packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
```diff
@@ -0,0 +1,124 @@
...
+        address       _dapp,
+        string memory _name
+        )
+        public
+        SecurityModule(_controller)
+    {
+        require(
+            _dapp != address(0) &&
+            !controller.moduleRegistry().isModuleRegistered(_dapp) &&
+            !controller.walletRegistry().isWalletRegistered(_dapp),
```
Brechtpd: Any particular reason to check for these? It doesn't make sense to use those addresses for sure, but these addresses should not be more dangerous than other addresses.

dong77: I think we can remove these checks :)

PR #849, opened 2019-12-11
[hebao] add GenericDAppModule - one module multiple dapps

dong77:


Brechtpd:
I would it like this. We could have an additional security parameter in this module/wallet. 

```
enum SecurityType {
    OWNER
    MAJORITY,
    TIME_DELAYED (not sure if this would be very helpful)
}
mapping (address => SecurityType) security;
mapping (address => mapping(address => bool)) approvedContracts;
```

The default would probably be `OWNER`. Wallets with security set to `SecurityType.MAJORITY` would need a majority for enabling a dapp (so `approvedContracts` needs to be true AND the contract needs to be whitelisted). If the security is set to `OWNER` the contract address only needs to be whitelisted (no need for `approvedContracts` to be true because the owner would be able to change it immediately anyway).

LuciferMorning11:
?

PR #850, opened 2019-12-11
[hebao] Small changes GenericDAppModule - one module multiple dapps

Brechtpd:
- Instead of using the fallback function I added `callDApp` because creating the correct data for the fallback function I think isn't easier because of the non-default data needed in the call. Also, by using `msg.value` it doesn't really play nice with our meta transactions.
- Added an `approveERC20` which is needed to approve ERC20 tokens for the dapp. This could also be done with `delegateCall` on the wallet but that would be more difficult to do.
- Added an optional extra security mode before the wallet owner can start using the dapp

### DIFF
--- a/packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
@@ -151,6 +151,36 @@ contract GenericDAppModule is SecurityModule
         onlyWhenWalletUnlocked(wallet)
         onlyApprovedDapp(wallet, dapp)
     {
+        approveERC20Internal(wallet, dapp, token, amount);
+    }
+
+    function approveAndCallDApp(
+        address          wallet,
+        address          dapp,
+        address          token,
+        uint             approvedAmount,
+        uint             value,
+        bytes   calldata data
+        )
+        external
+        nonReentrant
+        onlyFromMetaTxOrWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+        onlyApprovedDapp(wallet, dapp)
+    {
+        approveERC20Internal(wallet, dapp, token, approvedAmount);
+        transactCall(wallet, dapp, value, data);
+    }
+
+    function approveERC20Internal(
+        address wallet,
+        address dapp,
+        address token,
+        uint    amount
+        )
+        internal
+        nonReentrant
+    {
         bytes memory txData = abi.encodeWithSelector(
             ERC20(token).approve.selector,
             dapp,
--- a/packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
+++ b/packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
@@ -17,28 +17,21 @@
 pragma solidity ^0.5.13;
 pragma experimental ABIEncoderV2;
 
-import "../../../lib/AddressSet.sol";
-import "../../../lib/Claimable.sol";
-import "../../security/GuardianUtils.sol";
-
 import "../../security/SecurityModule.sol";
+
+import "../../security/GuardianUtils.sol";
 
 
 /// @title GenericDAppModule
 /// @dev GenericDAppModule allows wallet owners to transact directly or through meta
-///      transactions on any whitelisted dApps. The transaction data must be appended
-///      with the wallet address and then the dApp's address.
-contract GenericDAppModule is Claimable, AddressSet, SecurityModule
+///      transactions on any whitelisted dApps.
+contract GenericDAppModule is SecurityModule
 {
     enum SecuritySetting
     {
         Owner,
         Majority
     }
-
-    bytes32 internal constant DAPPS = keccak256("__DAPP__");
-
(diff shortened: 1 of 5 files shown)


packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
```diff
@@ -17,111 +17,163 @@
...
-            !controller.moduleRegistry().isModuleRegistered(dapp),
-            "MODULE_NOT_SUPPORTED"
-        );
-        require(
-            !controller.walletRegistry().isWalletRegistered(dapp),
-            "WALLET_NOT_SUPPORTED"
-        );
-        addAddressToSet(DAPPS, dapp, true);
-        emit DAppEnabled(dapp, true);
+        securitySettings[wallet] = SecuritySetting(_securitySetting);
```
dong77: when will an owner needs to use `setSecurity` and `setDAppApproved ` ?

dong77:
Maybe we should add a flag to Dapp registry to indicate `OWNER` or `MAJORITY`, otherwise, if the owner is fully authorized to invoke a dapp, he/she will not have the incentive to yield control to guardians. In other words, we should check signatures based on OWERN/MAJORITY when the wallet needs to transact on a dapp.

packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
```diff
@@ -17,111 +17,163 @@
...
+        onlyFromMetaTxOrWalletOwner(wallet)
+        onlyWhenWalletUnlocked(wallet)
+        onlyApprovedDapp(wallet, dapp)
     {
-        return addressesInSet(DAPPS);
+        transactCall(wallet, dapp, value, data);
     }
 
-    function()
+    function approveERC20(
```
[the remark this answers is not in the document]
Brechtpd: Because both will be often be called together it could certainly be helpful to have as an option to save a bit of gas.

packages/hebao_v1/contracts/modules/dapps/base/GenericDAppModule.sol
Brechtpd: If the default remains `SecuritySetting.Owner` than never. But the UI could expose an extra security setting to change this optionally for users that want extra security for dapps.

LuciferMorning11:
thats authentic .

dong77:
> Maybe we should add a flag to Dapp registry to indicate `OWNER` or `MAJORITY`, otherwise, if the owner is fully authorized to invoke a dapp, he/she will not have the incentive to yield control to guardians. In other words, we should check signatures based on OWERN/MAJORITY when the wallet needs to transact on a dapp.

Maybe we should allow the owner to request signatures from guardians to change a module from MAJORITY to OWNER. Then after that transaction, all future txs will only the owner's signature. But changing from OWNER to Majority doesn't make sense to me if I'm an owner.

Maybe we should split MAJORITY to `MAJORITY_GUARDIANS` and `MAJORITY_GUARDIANS_AND_OWNER`.

Brechtpd:
> Maybe we should add a flag to Dapp registry to indicate `OWNER` or `MAJORITY`, otherwise, if the owner is fully authorized to invoke a dapp, he/she will not have the incentive to yield control to guardians. In other words, we should check signatures based on OWERN/MAJORITY when the wallet needs to transact on a dapp.

That would indeed be an improvement. Though we still have to have a personal setting I think because users should not trust us to set the correct security value for each dapp. So perhaps the default value comes from the registry, but the user can also set the security value. The security used for the dapp is max(personalSecurity, registrySecurity). This way dapps we deem not secure enough just with the wallet owner cannot have lower security.

> Maybe we should allow the owner to request signatures from guardians to change a module from MAJORITY to OWNER. Then after that transaction, all future txs will only the owner's signature.

Don't we already kind of have this functionality by whitelisting addresses (used in `QuotaTransfers`)?

> But changing from OWNER to Majority doesn't make sense to me if I'm an owner.

It is if you're concerned with security.

> Maybe we should split MAJORITY to `MAJORITY_GUARDIANS` and `MAJORITY_GUARDIANS_AND_OWNER`.

I think we already do this in `requireSufficientSigners` with the `SigRequirement` parameter.

Brechtpd:
As `ApprovedTransfers` can be used with a majority to call any contract, and `QuotaTransfers` can be used within quota allowed, with whitelisted addresses (after a delay/dapp), or pending transactions (after a delay/transaction), the only advantage of `GenericDAppModule` seems to be that dapps we whitelist would be usable immediately without any limit.

dong77:
> As `ApprovedTransfers` can be used with a majority to call any contract, and `QuotaTransfers` can be used to call any contract within quota, with whitelisted addresses (after a delay/dapp), or pending transactions (after a delay/transaction), the only advantage of `GenericDAppModule` seems to be that dapps we whitelist would be usable immediately without any limit with only the authorization of the wallet owner. This is still an important advantage I think, but if we take this advantage away then basically `GenericDAppModule` is not really necessary and what we want can be done in existing modules (with maybe some small changes).

I agree. For most dapps, we should just enable wallet owners to transact. I still prefer https://github.com/Loopring/protocols/pull/848 which is simple and secure.

PR #851, opened 2019-12-12
[hebao] add a collectTokens to MetaTxModule

dong77:
Modules are not supposed to hold any assets.

### DIFF
--- a/packages/hebao_v1/contracts/base/ControllerImpl.sol
+++ b/packages/hebao_v1/contracts/base/ControllerImpl.sol
@@ -30,12 +30,13 @@ contract ControllerImpl is Claimable, Controller
     constructor() public Claimable() {}
     bool private initialized;
 
-    event ContractChanged(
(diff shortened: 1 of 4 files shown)


packages/hebao_v1/contracts/base/ControllerImpl.sol
```diff
@@ -65,19 +69,27 @@ contract ControllerImpl is Claimable, Controller
         quotaManager = _quotaManager;           // modifiable
     }
 
+    function setCollectTo(address _collectTo)
+        external
+        onlyOwner
+    {
+        collectTo = _collectTo;
```
Brechtpd: `collectTo` could be set to `address(0)`. Not necessarily wrong, but `init` does have the check. Either is fine for me.

dong77: fixed

PR #858, opened 2019-12-13
[hebao] Some small quota implementation changes

Brechtpd:
- Changed to a rolling daily limit instead of a hard day limit. The daily rolling limit is time zone independent and has some better worst case properties (with a hard day limit the daily limit be spent twice very shortly when done at 23.59 and 00.00, this is never possible with the rolling limit as the amount that can be spent gradually increases). The worst case (when hacked) is the same for both methods in that in 24 hours the daily limit can be spent twice.
- Allow the time it takes to change the daily quota to be set by the module. This allows for more flexible modules, which is used here to change the daily limit immediately with a majority.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/QuotaModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/QuotaModule.sol
@@ -20,16 +20,33 @@ pragma experimental ABIEncoderV2;
 import "../../iface/Wallet.sol";
 
 import "./SecurityModule.sol";
+import "./GuardianUtils.sol";
 
 
 /// @title QuotaModule
 /// @dev Manages transfer quota.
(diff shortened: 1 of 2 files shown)


packages/hebao_v1/contracts/modules/security/QuotaModule.sol
```diff
@@ -20,39 +20,58 @@ pragma experimental ABIEncoderV2;
...
 
 
 /// @title QuotaModule
 /// @dev Manages transfer quota.
 contract QuotaModule is SecurityModule, QuotaManager
 {
-    constructor(Controller _controller)
+    uint public delayPeriod;
+
+    modifier onlySufficientSigners(address wallet, address[] memory signers) {
```
dong77: should we move this method to SecurityModule?

PR #859, opened 2019-12-14
[hebao] Guardian refactoring

Brechtpd:
Instead of using enums with specific names for guardian groups, which we don't use, I just changed it so that it's possible to put a guardian in a certain group (nameless). The group decides which rules to follow:
- Group 0: No grouping (1 vote/guardian)
- Groups [1, 5]: Single guardian needed per group
- Groups [6, 10]: Half the guardians needed per group
- Groups [11, 15]: A majority of guardians needed per group

The reason for this fixed group ordering is that it makes counting the number of votes very easy and there's no need for an unlimited number of groups.

This way we don't try to make decisions on-chain what to do with certain types of guardians because this ended up being harder than expected (I don't believe there are universal rules that always work best for this). The user/UI can now decide how to group guardians and with which rules. Of course, the existing approach is compatible with the code in this PR.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -58,10 +58,6 @@ library GuardianUtils
                 signingGuardians[numGuardians++] = securityStore.getGuardian(wallet, signers[i]);
             }
         }
-        // Update the signingGuardians array with the actual number of guardians that have signed
-        // (could be 1 less than the length if the owner signed as well)
-        assembly { mstore(signingGuardians, numGuardians) }
-        uint[16] memory signed = countGuardians(signingGuardians);
 
         // Check owner requirements
         if (requirement == SigRequirement.OwnerRequired) {
@@ -69,6 +65,11 @@ library GuardianUtils
         } else if (requirement == SigRequirement.OwnerNotAllowed) {
             require(!walletOwnerSigned, "WALLET_OWNER_SIGNATURE_NOT_ALLOWED");
         }
+
+        // Update the signingGuardians array with the actual number of guardians that have signed
+        // (could be 1 less than the length if the owner signed as well)
+        assembly { mstore(signingGuardians, numGuardians) }
+        uint[16] memory signed = countGuardians(signingGuardians);
 
         // Count the number of votes
         uint totalNumVotes = 0;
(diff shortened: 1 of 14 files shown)


packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -61,12 +43,12 @@ library GuardianUtils
...
         // Calculate total group sizes
-        SecurityStore.Guardian[] memory allGuardians = securityStore.guardians(wallet);
-        GuardianGroupSizes memory total = countGuardians(allGuardians);
+        Data.Guardian[] memory allGuardians = securityStore.guardians(wallet);
+        uint[16] memory total = countGuardians(allGuardians);
 
         // Calculate how many signers are in each group
         bool walletOwnerSigned = false;
-        SecurityStore.Guardian[] memory signingGuardians = new SecurityStore.Guardian[](signers.length);
+        Data.Guardian[] memory signingGuardians = new Data.Guardian[](signers.length);
```
dong77: in the for-loop below, should we enforce `signers[i] > lastSigner` to deal with dupliciate signers?

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -79,7 +61,7 @@ library GuardianUtils
         // Update the signingGuardians array with the actual number of guardians that have signed
         // (could be 1 less than the length if the owner signed as well)
         assembly { mstore(signingGuardians, numGuardians) }
-        GuardianGroupSizes memory signed = countGuardians(signingGuardians);
+        uint[16] memory signed = countGuardians(signingGuardians);
```
dong77: maybe we should move the `SigRequirement.OwnerRequired` check to above, right after the for-loop.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
Brechtpd: The assumption is that the signatures are already checked at this point (as we don't check them here at all), so duplicate checking is also already done. Because of that I think we don't need to re-check those things here.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
Brechtpd: Will do to keep things closer apart, though it shouldn't matter because relayers/UI will check if these are valid before anything is submitted on-chain anyway, but a little bit of gas can indeed be saved in case things go wrong with this.

PR #871, opened 2019-12-18
[hebao] prevent certain actions when in recovery

dong77:
## Summary
Disable confirmGuardianAddition/confirmGuardianRemoval/addToWhitelist when a wallet is in recovery.

## Why
We have the following time-related parameters:

- `GuardianModule:pendingPeriod`
- `GuardianModule:confirmPeriod`
- `LockModule:lockPeriod`
- `QuotaModule:delayPeriod`
- `RecoveryModule:recoveryPeriod`
- `WhitelistModule:delayPeriod`

Assuming a wallet owner's private key is leaked, we need to make sure the wallet's lost is limited. 

The hacker will try to remove all guardians ASAP. The rightful owner will try to start recovery ASAP to set a new owning address. Therefore, removing guardians should take longer than the recovery process, i.e.,  we need to enforce  `GuardianModule:pendingPeriod > RecoveryModule:recoveryPeriod`.

The hacker may also want to add a new address to the whitelist to drain all assets. Therefore we also need to enforce `WhitelistModule:delayPeriod > RecoveryModule:recoveryPeriod`.

**The other approach** is to disallow guardian/whielist modificaiton while the recovery is in process. This PR implements this alternative approach.

## Discussion
We may also want to apply
```solidity
        onlyWhenWalletUnlocked(wallet)
        onlyWhenWalletNotRecovering(wallet)
```
to replace all ` onlyWhenWalletUnlocked(wallet)` occurances. But I'm not sure if we should do it.

### DIFF
--- a/packages/hebao_v1/contracts/modules/core/WalletFactoryModule.sol
+++ b/packages/hebao_v1/contracts/modules/core/WalletFactoryModule.sol
@@ -97,6 +97,16 @@ contract WalletFactoryModule is WalletFactory, MetaTxModule
         }
     }
 
+    function extractWalletAddress(bytes memory data)
+        internal
+        view
+        returns (address wallet)
+    {
+        require(extractMethod(data) == this.createWallet.selector, "INVALID_METHOD");
+        address owner = extractAddressFromCallData(data, 0);
+        wallet = computeWalletAddress(owner);
+    }
+
     function areAuthorizedMetaTxSigners(
         address   wallet,
         bytes     memory /*data*/,
--- a/packages/hebao_v1/contracts/modules/core/WalletFactoryModule.sol
+++ b/packages/hebao_v1/contracts/modules/core/WalletFactoryModule.sol
@@ -50,17 +50,14 @@ contract WalletFactoryModule is WalletFactory, MetaTxModule
         public
         MetaTxModule(_controller)
     {
-
     }
 
     /// @dev Create a new wallet by deploying a proxy.
-    /// @param _expectedWallet The expected smart contract wallet address that will be created.
     /// @param _owner The wallet's owner.
     /// @param _modules The wallet's modules.
     /// @param _subdomain The ENS subdomain to register, use "" to skip.
     /// @return _wallet The newly created wallet's address.
     function createWallet(
-        address            _expectedWallet,
         address            _owner,
         string    calldata _subdomain,
         address[] calldata _modules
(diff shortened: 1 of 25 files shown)


Brechtpd:
Argent does something clever where only the module that did the lock can do the unlock. That's actually pretty cool because a module can decide on the locking rules. Perhaps we can do the same.

Another thing I'm wondering if `cancelRecovery` and `completeRecovery` are even needed. `startRecovery` already checks the majority (though without the owner). What if we instead do the recovery immediately in a `recover` function that checks the majority with the owner included. This should work the same as the current system but without the extra recovery delay which simplifies the other delays.

dong77:
I prefer that lock and unlock are on the wallet level, not the module level. From the user's perspective, it will be really confusing to tell his/her guardian "please unlock/lock my wallet using module ABC".  The wallet lock should be global to all modules and independent of modules.

dong77:
I will change to enable immediate recovery.

Brechtpd:
> I prefer that lock and unlock are on the wallet level, not the module level. From the user's perspective, it will be really confusing to tell his/her guardian "please unlock/lock my wallet using module ABC". The wallet lock should be global to all modules and independent of modules.

The locking would be handled on a higher level. Guardians do not need to know the details. In the case of the 2 step recovery it would work like this:
- `startRecovery` which locks the wallet. This lock is on the wallet level and `onlyWhenWalletUnlocked` works as it does now (so still wallet wide, not just the recovery module)
- Now the only way to unlock the wallet is by calling `cancelRecovery`. But this is completely transparent to users/guardians, they don't even need to know that the locking/unlocking happens. It just automatically makes it so that any other operation is impossible (similar to what this PR does with `onlyWhenWalletNotRecovering`).

So it's more of an operation specific lock which can be used in these special cases. I still thinks it's pretty neat because future modules may need to have something similar. We still haven't deployed anything so we can make still easily modify the store contracts, but in the future we'll have to remain compatible with existing modules so having a general system for this would probably be nice to have.

dong77:
So your idea is to have a `lockInternal(yesOrNo, who)` function that can be called by any module, and only that module can unlock with his own  public/external function by calling `lockInternal(false, address(this))`. And the LockModule is just a special module that provides default public lock/unlock functions? Just want to confirm the design.

dong77:
I'll create another PR for the lock/unlock change. Please take another look at this PR without worrying about locking.

packages/hebao_v1/contracts/modules/core/WalletFactoryModule.sol
```diff
@@ -50,39 +50,34 @@ contract WalletFactoryModule is WalletFactory, MetaTxModule
...
     }
 
     /// @dev Create a new wallet by deploying a proxy.
-    /// @param _expectedWallet The expected smart contract wallet address that will be created.
     /// @param _owner The wallet's owner.
     /// @param _modules The wallet's modules.
     /// @param _subdomain The ENS subdomain to register, use "" to skip.
     /// @return _wallet The newly created wallet's address.
     function createWallet(
-        address            _expectedWallet,
```
Brechtpd: I added this here to make it compatible with meta transactions without doing anything as these currently always get the wallet address from the first function parameter. But this can also be done quite easily by overriding `extractWalletAddress` in `WalletFactoryModule`:

```
function extractWalletAddress(bytes memory data)
    internal
    view
    returns (address wallet)
{
    address owner = extractAddressFromCallData(data, 0);
    wallet = computeWalletAddress(owner);
}
```

Brechtpd:
> So your idea is to have a `lockInternal(yesOrNo, who)` function that can be called by any module, and only that module can unlock with his own public/external function by calling `lockInternal(false, address(this))`. And the LockModule is just a special module that provides default public lock/unlock functions? Just want to confirm the design.

Yes! Otherwise any wallet lock we would ever do in a module would be able to be unlocked by a single guardian, which would be quite limiting in what would be possible with locking.

Issue #873, opened 2019-12-18
[hebao] security issue - SecurityModule can be removed


dong77:
If a wallet owner's private key is hacked, the hacker can call BaseModule.removeModule to remove the Security module to prevent ownership recovery by the rightful owner.

```solidity
    /// @dev Removes a module from a wallet. Callable only by the wallet owner.
    ///      Note that the module must have been added to the wallet.
    function removeModule(
        address wallet,
        address module
        )
        external
        nonReentrant
        onlyFromMetaTxOrWalletOwner(wallet)
    {
        require(module != address(this), "SELF_REMOVE_PROHIBITED");
        Module(module).deactivate(wallet);
        Wallet(wallet).removeModule(module);
    }
```

PR #874, opened 2019-12-18
[hebao] Fix remove module security issue

dong77:
Try to solve issue https://github.com/Loopring/protocols/issues/873

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseModule.sol
+++ b/packages/hebao_v1/contracts/base/BaseModule.sol
@@ -46,12 +46,6 @@ contract BaseModule is ReentrancyGuard, Module
         _;
     }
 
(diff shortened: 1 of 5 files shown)


Brechtpd:
If we do it like this again we'll have to re-add the check in `transactInternal` that a module cannot be called, otherwise the wallet owner can do a call using e.g. `callContract` to `activate`/`deactivate` to the module. I tried to avoid giving the wallet address any special permissions in the modules so checks like this weren't needed, but on second thought I don't think this is necessarily bad as the check is limited to a single location (well, it's possible the function can also be called in the fallback function, but then the module would have to bind it which won't be the case as only registered modules can be used bu the user).

dong77:
I committed another code change to disable Wallet calling transact on Modules.

PR #876, opened 2019-12-19
[protocol3] Some small improvements to DelayedOwner

Brechtpd:
Some small changes I did for testing chainlink.

### DIFF
--- a/packages/loopring_v3/contracts/lib/DelayedTransaction.sol
+++ b/packages/loopring_v3/contracts/lib/DelayedTransaction.sol
@@ -1,0 +1,321 @@
+/*
+
+  Copyright 2017 Loopring Project Ltd (Loopring Foundation).
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


packages/loopring_v3/contracts/lib/DelayedTransaction.sol
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    function exectuteTransaction(
+        Transaction memory transaction
+        )
+        internal
+        returns (bool success, bytes memory returnData)
+    {
+        // solium-disable-next-line security/no-call-value
+        (success, returnData) = transaction.to.call.value(transaction.value)(transaction.data);
```
dong77: should we make sure `to != address(this)`?

packages/loopring_v3/contracts/lib/DelayedTransaction.sol
```diff
@@ -0,0 +1,321 @@
...
+import "./AddressUtil.sol";
+import "./BytesUtil.sol";
+import "./MathUint.sol";
+
+
+/// @title DelayedOwner
+/// @author Brecht Devos - <brecht@loopring.org>
+/// @dev Base class for an Owner contract where certain functions have
+///      a mandatory delay for security purposes.
+contract DelayedTransaction is IDelayedTransaction
```
dong77: How about we use the AddressSet's design to create a BytesSet, then each function will have its own bytes set for encoded DelayedFunction. I think this will simplify storage-related code in this contract.

Brechtpd: See #888

packages/loopring_v3/contracts/lib/DelayedTransaction.sol
Brechtpd: I actually use this in #877 to delay the setting of the USD values in the owner contract itself. So I would definitely keep this.

PR #877, opened 2019-12-19
[protocol3] Initial chainlink integration

Brechtpd:


packages/loopring_v3/contracts/impl/owners/LoopringV3Owner.sol
```diff
@@ -19,19 +19,137 @@ pragma solidity ^0.5.11;
...
 
+import "./ChainlinkTokenPriceOracle.sol";
+
 
 /// @title LoopringV3Owner
 /// @author Brecht Devos - <brecht@loopring.org>
 contract LoopringV3Owner is DelayedOwner
 {
+    uint public constant NUM_MOVING_AVERAGE_DATA_POINTS = 7;
+    uint public constant MOVING_AVERAGE_TIME_PERIOD = 1 days;
```
dong77: Brecht, can we have a separate contract that provides the usdToLRC function using the 7-day moving average so other contracts (such as DowntimeCostCalculator) can also use?

packages/loopring_v3/contracts/impl/owners/ChainlinkTokenPriceOracle.sol
```diff
@@ -0,0 +1,55 @@
...
+    /// @param usd The amount of USD (1 wei == 1 USD)
+    /// @return The amount of LRC
+    function usd2lrc(uint usd)
+        external
+        view
+        returns (uint)
+    {
+        uint EthUsd = uint(Eth2Usd.currentAnswer());
+        uint LrcEth = uint(Lrc2Eth.latestAnswer());
+        return usd.mul(100000000 * 1 ether * 1 ether) / LrcEth.mul(EthUsd);
```
dong77: can you explain this line to me, having a hard time to understand it.

dong77: Maybe add as a comment.

dong77:
Brecht, if we have a MovingAveragePriceOracle or something like this, we can greatly simplify the logic in the OwnerContract.

PR #883, opened 2019-12-20
[hebao] refactor modules for upgradability

dong77:
Merge WalletUpgraderModule to UpgraderModule.


packages/hebao_v1/contracts/modules/core/UpgraderModule.sol
```diff
@@ -4,32 +4,49 @@ import "../../iface/Controller.sol";
...
+        implementation = _implementation;
         modulesToAdd = _modulesToAdd;
+        modulesToRemove = _modulesToRemove;
     }
 
     function activate()
         external
     {
+        if (implementation != OwnedUpgradabilityProxy(msg.sender).implementation() &&
+            controller.implementationRegistry().isImplementationRegistered(implementation)) {
```
Brechtpd: Because the implementation is set in the module constructor, and only modules that are registered can be used by users, maybe we can just remove the `implementationRegistry` now because it's not really necessary any more?

Brechtpd:
I just noticed we don't have a `addModule` any more that can add modules using a meta transaction. This isn't good because users would need ETH to do it. Maybe we can expose `addModule` here (for any module, not just those given in the contructor) as well from a meta transaction so all upgrade features are bundled in `UpgraderModule`. Otherwise adding it back in `BaseModule` would also work.

dong77:
> I just noticed we don't have a `addModule` any more that can add modules using a meta transaction. This isn't good because users would need ETH to do it. Maybe we can expose `addModule` here (for any module, not just those given in the contructor) as well from a meta transaction so all upgrade features are bundled in `UpgraderModule`. Otherwise adding it back in `BaseModule` would also work.

I created an issue for it:)

packages/hebao_v1/contracts/modules/core/UpgraderModule.sol
dong77: done

dong77: done

dong77: done

PR #884, opened 2019-12-20
[hebao] better lock and unlock

dong77:
- unlock is now only possible by the same module.
- UI level lock is only possible through LockModule, but all SecurityModules can lock a wallet using its own internal logic.


packages/hebao_v1/contracts/iface/Controller.sol
```diff
@@ -37,6 +37,7 @@ contract Controller
     // The address to which all colletable tokens/ether in modules will
     // be sent to.
     address public collectTo;
+    uint    public lockPeriod;
```
Brechtpd: I think this will be module depend, but it's not like the module is forced to use this value so I guess it can't hurt to have a default value.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -128,23 +143,32 @@ contract SecurityStore is DataStore
...
+    function setLock(
+        address wallet,
+        uint    lock
+        )
         public
         onlyManager
     {
-        require(lock == 0 || lock > now, "INVALID_LOCK_TIME");
+        if (isLocked(wallet)) {
+            require(
```
Brechtpd: Maybe we should keep this check in the module. I'm thinking of a scenario like with the recovery in multiple steps. As a first step the wallet owner may lock the wallet using a single guardian because it's quick. After he has gathered enough signatures he would start the recovery which locks the wallet again, but with a much stricter unlock criteria. With the current logic this would be impossible because the user would have to unlock the wallet first just so he can start the recovery that locks it again (and this stricter lock would be really needed, so skipping the lock is not an option when the wallet is already locked). So I think we should make it possible for modules to overwrite locks in some cases (but not in all cases, otherwise after recovery is started with a majority the wallet is locked with a majority, we don't want to allow the case where the lock is overwritten by a single guardian in `LockModule` so it can also be unlocked again by a single guardian).

dong77: Agreed.

packages/hebao_v1/contracts/iface/Controller.sol
dong77: that's right.  I changed it to `defaultLockPeriod` and add another `lockWallet` function with a second parameter called `lockPeriod`

PR #888, opened 2019-12-20
[protocol3] Use Map for DelayedTransactions

Brechtpd:
Introduces a reusable Map and Set with bytes so they could be used for any data type (see [here](https://github.com/Loopring/protocols/pull/876#discussion_r360206847)). Having a Map directly instead of making use of a Set makes more sense in this case (as we need map functionality).

It has its pro's and con's. The code is simpler, but there's also a decent amount of boilerplate code to encode/decode structs (and expose the structs without relying on `ABIEncoderV2` for now). It would have been very nice of solidity would expose encoding/decoding structs, but [that's not the design they want to make possible unfortunately](https://github.com/ethereum/solidity/issues/7134) (the design they want to implement is not coming soon in any case, while encoding/decoding structs is already possible internally...).

### DIFF
--- a/packages/loopring_v3/contracts/iface/IDelayedTransaction.sol
+++ b/packages/loopring_v3/contracts/iface/IDelayedTransaction.sol
@@ -66,7 +66,6 @@ contract IDelayedTransaction
     {
         address to;
         bytes4  functionSelector;
-        uint    delay;
     }
 
     // The maximum amount of time (in seconds) a pending transaction can be executed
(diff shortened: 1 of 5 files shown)


dong77:
@Brechtpd what's your preference given the pros and cons?

dong77:
@Brechtpd what's your preference given the pros and cons?

Brechtpd:
I would probably not do this for the time being. I have tried improving on this but have not found anything really satisfactory. We can keep this method in mind for further implementations. Currently this would only be used for the owner contracts which are easily replaced by new implementations. Probably best not to rush into actually using it here as we'll deploy this soon for the chainlink integration.

dong77:
I see. I guessed this is not what you want. Sorry for wasting some of your time :)

dong77:
I'll close it for now but keep the branch.

PR #906, opened 2019-12-25
[Protocol3] fix `npm run genabi` issue

dong77:


Brechtpd:
I also fixed this in #877 differently by allowing interfaces to be imported in libs, see [here](https://github.com/Loopring/protocols/blob/4055ce6d4dfc354936efe4888f9f4ebddeaf03ca/packages/loopring_v3/genAbi.sh). I don't know what would be best (or if there it matters).

PR #920, opened 2019-12-28
[Protocol3] add batch operator contract

dong77:


packages/loopring_v3/contracts/impl/operators/BatchOperator.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    function batchCall(
+        bytes[] calldata transactions
+        )
+        external
+        payable
+        onlyOwner
+    {
+        require(transactions.length > 0, "EMPTY_DATA");
+        require(msg.value == 0, "INVALID_MSG_VALUE");
```
Brechtpd: If this is a requirement not making the function `payable` makes more sense.

packages/loopring_v3/contracts/impl/operators/BatchOperator.sol
```diff
@@ -0,0 +1,137 @@
...
+
+        for (uint i = 0; i < transactions.length; i++) {
+            /* solium-disable-next-line */
+            (bool success, ) = exchange.call(transactions[i]);
+            require(success, "FAILURE");
+        }
+    }
+
+    function batchCall(
```
Brechtpd: Seems like it would be better to just to have this `batchcall` so we don't have 2 similar functions (with one that can just do less)?

Brechtpd:
Some other things to note when using this:
- The DEX browser lib will need to be familiar with the transaction data for operators like this so the data can be successfully extracted. Extracting data from the internal call to the exchange would be possible (some websites can do it), but is harder to do (I'm pretty sure you can't do it with just web3).
- Bigger blocks are of course much more efficient than smaller blocks. Batching multiple smaller blocks for `commitBlock` of the same type doesn't make that much sense because
  - Doing the work of the smaller blocks in a single bigger block would be more efficient
  - By batching the blocks the economic finality of blocks is postponed (so e.g. creating 2 blocks of 128 rings/block and batching them makes no sense at all as it would effectively be exactly the same as 1 block of 256 rings/block, and the 256 rings/block would be much more efficient).
- Batching the commit and the verify of the same block would save some gas (around 5000), but 
  - It doesn't necessarily save gas in all scenario's because the batch verification, only possible for blocks with the exact same block parameters type and size, would save a lot more gas. Depending on the scenario and how the blocks are batched this could be slightly more efficient or substantially less efficient. 
  - No use is made of the faster economic finality we have with the commit/verify scheme.

So the main gas savings for using this is to batch multiple Ethereum transaction into a single one, saving the fixed 21,000 gas cost/Ethereum transaction. Using the exchange with the biggest blocks possible and efficient batch verification is much more important, but the batching done at this level can also save a bit of gas.

packages/loopring_v3/contracts/impl/operators/BatchOperator.sol
dong77: Removed

dong77:
> Some other things to note when using this:
> 
> * The DEX browser lib will need to be familiar with the transaction data for operators like this so the data can be successfully extracted. Extracting data from the internal call to the exchange would be possible (some websites can do it), but is harder to do (I'm pretty sure you can't do it with just web3).
> * Bigger blocks are of course much more efficient than smaller blocks. Batching multiple smaller blocks for `commitBlock` of the same type doesn't make that much sense because
>   
>   * Doing the work of the smaller blocks in a single bigger block would be more efficient
>   * By batching the blocks the economic finality of blocks is postponed (so e.g. creating 2 blocks of 128 rings/block and batching them makes no sense at all as it would effectively be exactly the same as 1 block of 256 rings/block, and the 256 rings/block would be much more efficient).
> * Batching the commit and the verify of the same block would save some gas (around 5000), but
>   
>   * It doesn't necessarily save gas in all scenario's because the batch verification, only possible for blocks with the exact same block parameters type and size, would save a lot more gas. Depending on the scenario and how the blocks are batched this could be slightly more efficient or substantially less efficient.
>   * No use is made of the faster economic finality we have with the commit/verify scheme.
> 
> So the main gas savings for using this is to batch multiple Ethereum transaction into a single one, saving the fixed 21,000 gas cost/Ethereum transaction. Using the exchange with the biggest blocks possible and efficient batch verification is much more important, but the batching done at this level can also save a bit of gas.

I totally agree.

PR #922, opened 2019-12-29
[hebao] Add tests for Hebao smart contracts

Brechtpd:
Still trying some new things out and working around solidity/truffle bugs...


Brechtpd:
Some changes in this PR:
- `registerWallet` is called in the module instead of in the wallet, because it can only be called by authorized addresses (which is now just the `walletFactory`)
- The return data of the call in a meta tx is now logged. This allows, for example, to decode the revert message when the call fails.
- A pending tx in TransferModule is now only created when the owner wants to transfer more than his current quota.

Some open questions:
- I think `QuotaManager` can be removed to keep things simple. I added an `updateQuota` function that can be used in the modules which seems sufficient.
- A meta tx cannot be used to remove itself when the user pays for it. This is also a problem for creating a wallet using a meta tx.

Brechtpd:
Not 100% tested yet, but it should be pretty good already.

dong77:
@letsgoustc can you help review this PR?

dong77:
https://github.com/Loopring/team/issues/67#issuecomment-583118559

可以参考新的费用设置。我们对maker收取的交易费为0.

packages/hebao_v1/contracts/base/ERC1271Module.sol
```diff
@@ -42,35 +43,23 @@ abstract contract ERC1271Module is ERC1271, BaseModule
...
+        } else {
+            hash = keccak256(_data);
         }
-
-        bytes32 signHash = keccak256(
-            abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(_data))
-        );
-
-        return signHash.recoverECDSASigner(_signature) == _address;
+        hash.verifySignature(Wallet(msg.sender).owner(), _signature);
```
dong77: should we use try-catch and make sure this method returns instead of throwing exception?

Brechtpd: Try/Catch can only be used for external calls unfortunately.

But I do agree that it would be cleaner if the function just returns true/false instead of throwing, but that needs quite some changes in `verifySignature`, but it's probably worth it do do that.

dong77: Ok, lets add an TODO or issue then.

dong77:
@Brechtpd please wait for @letsgoustc 's review result, my review is not thorough.

packages/hebao_v1/contracts/base/ERC1271Module.sol
Brechtpd: Changed it so it returns a boolean. The function can still revert when the data is wrongly formatted, but I think that is acceptable.

packages/hebao_v1/test/helpers/Signature.ts
```diff
@@ -0,0 +1,146 @@
...
+  const type = data.extractUint8(data.length() - 1);
+  if (type === SignatureType.ETH_SIGN) {
+    const personalMessage = ethUtil.hashPersonalMessage(message);
+    assert(
+      verifyECDSA(personalMessage, data, signer),
+      "invalid ETH_SIGN signature"
+    );
+  } else if (type === SignatureType.EIP_712) {
+    assert(verifyECDSA(message, data, signer), "invalid EIP_712 signature");
+  }
```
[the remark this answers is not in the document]
Brechtpd: Added the WALLET type, though these typescript functions aren't really used in the tests, they are mainly there for debugging/sanity checking. 

The WALLET signature can be anything, but when it's compatible with our `ERC1271Module` module it will mostly contains data like this:
```
[actual signature data][signature type X][signature type WALLET]
```
So it's kind of like a recursion thing. The first time the signature is processed in the meta tx we detect it's a wallet signature. We then strip off the wallet byte and pass the signature to the wallet contract. The wallet contract can do anything with it as it likes, but our `ERC1271Module` module checks the latest byte of the signature again, which can be any signature type (it could even be a WALLET signature again which is passed on to another contract). This way we can support any signature type even if the initial wallet type is WALLET.

PR #927, opened 2019-12-31
[Protocol3] withdraw tokens not owned by users to vault

dong77:


packages/loopring_v3/contracts/impl/libexchange/ExchangeAdmins.sol
```diff
@@ -224,7 +226,10 @@ library ExchangeAdmins
         amount = totalBalance - userBalance;
 
         if (amount > 0) {
-            token.safeTransferAndVerify(recipient, amount);
+            if (token == address(0))
+                recipient.sendETHAndVerify(amount, gasleft());
```
Brechtpd: We can't do ETH here because ETH is also used for paying on-chain deposit/withdrawal fees (we don't track these in `tokenBalances`).

dong77: I see.

Issue #291, opened 2019-07-22
[Protocol3] Test the optimized libsnark implementation from SECBIT


kongliangzhong:
SECBIC optimized [libsnark](https://github.com/huyuguang/libsnark) with a new paring library, we can try to use their implementation and measure there is any performance boost (~20% expected)

Brechtpd:
I don't think we can use this (yet). It seems like we need to use MCL_BN128 which I think is not the same as ALT_BN128 (different parameters/calculations). We need all parameters (not just the curve parameters `a` and `b`) to be exactly the same as described in [EIP-196](http://eips.ethereum.org/EIPS/eip-196) and [EIP-197](http://eips.ethereum.org/EIPS/eip-197), otherwise we can't use the precompile. For example, we can't use BN128 because it differs from ALT_BN128 by just a generator.

However, maybe some optimizations can also be done on ALT_BN128, I don't know, but it doesn't look like we can just switch to MCL_BN128 as it is.

I have some problems compiling it with ethsnarks so I can't be sure of this, but I'd like to hear the thoughts of SECBIT on this.

With Istanbul we may be able to use any curve (still very uncertain), but if we can use any curve we'll probably want to switch to a curve that allows for recursive SNARKs (MNT).

huyuguang:
> I don't think we can use this (yet). It seems like we need to use MCL_BN128 for the better performance, but this curve setup doesn't look the same as ALT_BN128 (different parameters/calculations). We need all parameters (not just the curve parameters `a` and `b`) to be exactly the same as described in [EIP-196](http://eips.ethereum.org/EIPS/eip-196) and [EIP-197](http://eips.ethereum.org/EIPS/eip-197), otherwise we can't use the precompile (if we can't use the precompile the gas cost is too high to verify the proof). For example, we can't use BN128 because it differs from ALT_BN128 by just a generator.
> 
> However, maybe some optimizations can also be done on ALT_BN128, I don't know, but it doesn't look like we can just switch to MCL_BN128 as it is.
> 
> I have some problems compiling it with ethsnarks when using MCL_BN128 so I'm not sure of this yet, but I'd like to hear the thoughts of SECBIT on this.
> 
> With Istanbul we may be able to use any curve (still very uncertain, and maybe just some extra curves), but if we can use any curve we'll probably want to switch to a curve that allows for recursive SNARKs (MNT).

It because of the mcl_bn128 use the curve param same with bn128 but not alt_bn128.
[...]

PR #990, opened 2020-02-05
[Protocol3] Prover optimizations

Brechtpd:
This PR (together with the circuits PR, all on the `mcl` branch) contains the following optimizations:
- **Switched to the [mcl](https://github.com/herumi/mcl) library for curve operations.** This is also what was basically discussed in #291, but I do not use any branch of SECBIT because those branches have some issues (large diffs because of formatting changes, introduces some bugs even for standard alt_bn254). The main integration code is the same to let libsnark make use of the lib (which was mainly written by herumi himself, the author of mcl), but that's it. Other changes are minimal. The result is 2x-3x faster multi exponentiations, resulting in ~50% faster proving times (but we have some overhead so this was much less, see below). The only thing we have to do to make use of this is convert our proving keys to a slightly different binary format. I have built the necessary conversion into our circuits binary, it can be used as 
`-pk_alt2mcl <block.json> <pk_alt.raw> <pk_mlc.raw>: Converts the proving key from the alt format to the mcl format `. By default the new format is used now (as there is no reason not to), the conversion is only necessary for our trusted setup proving keys.
- **Introduced a server mode in the circuits binary**. This was necessary to make efficient use of hardware with a lot of cores. Some of the operations needed to start proving a block are slow and single-threaded:
  - Creating the circuits (creating all the constraints etc...)
  - Loading the proving key (large files + needs processing)
But these operations can be reused every time the same circuit needs to be used to prove a block! And because we'll do trade blocks ~>90% of the time with some block size we can easily have some machines set to prove those kind of blocks as efficiently as possible.
  I implemented a simple http server for instructing the prover server which works surprisingly well and is very easy to use and integrate with as any langure will have libraries to communicate with an http server. For example, a block can be proven without any extra scripting with `curl "http://localhost:1234/prove?block_filename=blocks/block_3_3.json" > proof.json`. Multiple blocks can be queued so the hardware can be used 100% efficiently without any downtime. Of course if some other mechanism is needed that can be implemented as well, but this seems to work pretty well already.
- **Parallelized the witness generation**: Some extra data needs to be generated before a block can be proven (mostly intermediate values for the constraints). This was not multi-threaded and could take some time.
- **Parallelized the block validation**: Blocks don't need to be validated before they are proven, but I guess this was re-enabled to make sure no useless proofs were generated for invalid blocks. Should only be done as long as we are unsure the relayer can create invalid blocks, otherwise this should be skipped as it is still quite expensive.
- **Enabled more compiler optimizations** when the circuits binary is compiled, which were some small but easy gains.

With these changes we can make 100% use of expensive hardware with a lot of cores without wasting time doing work on a single thread while lots of cores are just idle. When using the server mode all main tasks are now done multi-threaded and no unnecessary work is done between proving different blocks.

As a result, at least on my computers, proofs can now be generated around **70% faster** so in theory the prover costs can be reduced to ~1/3 of what they currently are. Of course, results may vary between different computers (depending on number of cores used, IO speed, ....) but without a doubt the hardware is used much more efficiently.

I think there are still some areas in the prover that can be improved, but I think I got most of the low hanging fruit in this PR. 

Almost all changes in the respective libraries:
- https://github.com/Loopring/libff/tree/mcl
- https://github.com/Loopring/libsnark/tree/mcl
- https://github.com/Loopring/ethsnarks/tree/mcl

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "packages/loopring_v3/ethsnarks"]
 	path = packages/loopring_v3/ethsnarks
-	url = https://github.com/LoopringSecondary/ethsnarks
+	url = https://github.com/Loopring/ethsnarks
 	branch = mcl
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "packages/loopring_v3/ethsnarks"]
 	path = packages/loopring_v3/ethsnarks
 	url = https://github.com/LoopringSecondary/ethsnarks
-	branch = mpc
+	branch = mcl
--- a/packages/loopring_v3/Makefile
+++ b/packages/loopring_v3/Makefile
@@ -2,7 +2,7 @@ ROOT_DIR := $(shell dirname $(realpath $(MAKEFILE_LIST)))
 PYTHON=python3
 
 # CMAKE_TYPE := cmake-release
-CMAKE_TYPE := cmake-openmp-release
+CMAKE_TYPE := cmake-openmp-performance
 
 ifeq ($(OS),Windows_NT)
 	detected_OS := Windows
--- a/packages/loopring_v3/circuit/CMakeLists.txt
+++ b/packages/loopring_v3/circuit/CMakeLists.txt
@@ -1,15 +1,28 @@
 cmake_minimum_required(VERSION 3.9)
 project(circuit)
 
+set(
+  CURVE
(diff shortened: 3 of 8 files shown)


dong77:
@Brechtpd   "Parallelized the witness generation" and "Parallelized the block validation"  are both done to make more use of multiple cores, correct? (compared to network-level parallelization)

PR #1006, opened 2020-02-14
add docker files

kongliangzhong:


Brechtpd:
This PR seems to roll back the ethsnarks submodule to before #990, which I assume is unintentional?

kongliangzhong:
> This PR seems to roll back the ethsnarks submodule to before #990, which I assume is unintentional?

You are right. Sorry, I will correct it soon.

PR #1021, opened 2020-02-27
[protocol3] Prover optimizations (part 2)

Brechtpd:
The next batch of optimizations building on #990. Compared to that version the prover should be about **5x faster on a 16 core CPU and uses 5x less memory** (less than 1GB per 1M constraints). Because of that we can now use a **c5.9xlarge** "compute optimized" instance at **$1.5/hour** to prove a 1024 trade block (currently our biggest circuit with 64M constraints). This instance type even has 18 CPU cores (though currently we can't make really good use of those 2 extra cores, but it already helps a bit).

**A 1024 trades block now only takes 105 seconds to prove on a c5.9xlarge.** Note that because a certain optimization that changes the circuits themselves currently cannot be used without redoing our trusted setup it's actually 5% slower for us now (110 seconds). So not too bad, but something to keep in mind for next trusted setups to not forget to do this optimization.

## Practical info

- **The proving key needs to be updated again**. The conversion needs to be done on top of the previous `mcl` format:
```
-pk_mcl2nozk <pk_mlc.raw> <pk_nozk.raw>: Converts the proving key from the mcl format to the nozk format
```
If the prover keys are still in their original format they first need to be converted to the `mcl` format first:

```
-pk_alt2mcl <pk_alt.raw> <pk_mlc.raw>: Converts the proving key from the alt format to the mcl format
```

- For the best performance on different hardware the prover can be tweaked using a config file. **This config file is currently required because the optimization `swapAB` needs to be turned off to be compatible with our trusted setup circuits**. This config file needs to be named `config.json` and needs to be placed in the working directory. A good config file for running on an c5.9xlarge instance is:
```
{
	"num_threads": 32,
	"smt": true,
	"fft": "recursive",
	"multi_exp_c": 16,
	"swapAB": false,
	"prefetch_stride": 16,
	"multi_exp_prefetch_locality": 1,
	"multi_exp_look_ahead": 2
}
```

You can verify the config file is correctly picked up as the config is the first thing that's being printed in the output:
```
Config: num_threads: 32, smt: 1, fft: recursive, radixes: [4,4], exp_c: 16, pre_stride: 16, exp_preloc: 1, exp_lookahead: 2
```

- **Compile with clang!** I always compile with clang but by chance I compiled with gcc and gcc creates an executable that is 10% slower than the clang executable (for some reason my FFT is 30% slower when compiled with gcc). On linux, clang can be installed and set as the default compiler like this:

```
sudo apt-get install clang
sudo update-alternatives --config c++
``` 
Now enter the number corresponding with clang and press enter. Delete the build files (if they exist) and run `make` again.

## Optimizations

A bit more in depth info about the optimizations. I will provide some more info in a medium article or you can ask me here for more information:
- New FFT implementation. Recursion based and with precomputed twiddle factors.
- Modified the multiexp algorithm so it works more like bellman. But with additional tweaks that have a large impact.
- Removed almost all memory allocations in the prover itself. Memory is allocated a single time when the prover starts up and it is reused for all proofs from then on.
- Removed some calculations that are only needed for the zero-knowledge part of groth16. As we only use snarks for scalability we do not need it.
- Memory use was reduced a lot by storing the circuit more efficiently in memory. There is still a flat buffer of all constraints currently but I have moved partially  away from that. There is still some work to completely move away from it, there could still be some decent memory savings into doing the complete migration.
  - There were many duplicated coefficients in our constraints. Coefficient are field elements so they are 32 bytes, which is quite large. These are now saved in a single pool and the constraint itself contains an index to the coefficient
  - Constraints themselves are duplicated a lot as most constraints are from Poseidon and SHA256, which mostly do the same thing each time data is hashed, just on different variables. Constraints for these are only stored once and they are modified on the fly for a particular instance.
- The circuits are now much faster to create in memory.
- The proving key is now substantially smaller and loads faster. The circuit was serialized into the proving key. We don't need that as we can regenerate the circuit from code which has some additional benefits. The A query is now stored and serialized in a sparse vector. Both save on storage and on memory.
- There's now a benchmark mode that allows easy testing of different parameters to find the optimal configuration on a certain system.

## Link to repos with the actual changes
- https://github.com/Loopring/libfqfft/tree/opt
- https://github.com/Loopring/libff/tree/opt
- https://github.com/Loopring/libsnark/tree/opt
- https://github.com/Loopring/ethsnarks/tree/opt
- https://github.com/Loopring/protocol3-circuits/tree/opt

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "packages/loopring_v3/ethsnarks"]
 	path = packages/loopring_v3/ethsnarks
 	url = https://github.com/Loopring/ethsnarks
-	branch = mcl
+	branch = opt
--- a/packages/loopring_v3/README.md
+++ b/packages/loopring_v3/README.md
@@ -13,7 +13,7 @@ To understand the overall design for Loopring 3.0, including Ethereum smart cont
 |                           | Loopring 2.x | Loopring 3.0 <br> (w/ Data Availability) | Loopring 3.0 <br> (w/o Data Availability) |
 | :------------------------ | :----------: | :--------------------------------------: | :---------------------------------------: |
 | Trades per Ethereum Block |      26      |                  26300                   |                  216000                   |
-| Trades per Second         |      ~2      |                   2025                   |                   16400                    |
+| Trades per Second         |      ~2      |                   2025                   |                   16400                   |
 | Cost per Trade            | ~300,000 gas |                 375 gas                  |                  47 gas                   |
 
 - _Cost in USD per Trade_ in the table does not cover off-chain proof generation.
@@ -48,7 +48,7 @@ Please use node v10.
 
 ### Circuits
 
-The code of our circuits is currently not open source. If you have access to the private repo `protocol3-circuits` please clone it and update the `circuit_src_folder` variable in `circuit/CMakeLists.txt` so it points to the correct folder.
+Clone the `protocol3-circuits` repo and update the `circuit_src_folder` variable in `circuit/CMakeLists.txt` so it points to the correct folder.
 
 ```
 make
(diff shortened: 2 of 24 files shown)


kongliangzhong:
@Brechtpd , I got the following error when trying to update submodules:
ubuntu@ip-172-31-3-125:~/codebase/protocols/packages/loopring_v3$ git submodule update --init --recursive --remote
fatal: Needed a single revision
Unable to find current origin/opt revision in submodule path 'depends/libfqfft'
Failed to recurse into submodule path 'ethsnarks'

Brechtpd:
Please use a fresh checkout to not run into problems with the submodules.

It's also possible to manually fix this problem by removing the module in the `.git` folder and the repo, and to run `git submodule sync` in the correct repo, but that's making things more difficult than necessary. Unfortunately changes to submodules URLs/branches don't always work smoothly.

PR #1035, opened 2020-03-13
[HEBAO_V1] add unit test for LoopringModule

kongliangzhong:


packages/hebao_v1/contracts/test/DummyExchange.sol
```diff
@@ -0,0 +1,182 @@
...
+
+    function deposit(
+        address tokenAddress,
+        uint96  amount
+        )
+        external
+        payable
+        override
+    {
+
```
letsgoustc: Looks we should emit Deposit event.

packages/hebao_v1/contracts/test/DummyExchange.sol
```diff
@@ -0,0 +1,182 @@
...
+    function withdraw(
+        address tokenAddress,
+        uint96  amount
+        )
+        external
+        payable
+        override
+    {
+
```
letsgoustc: Looks we should emit Withdraw event.

packages/hebao_v1/test/testLoopringModule.ts
```diff
@@ -0,0 +1,132 @@
...
+            []
+          ),
+          ctx,
+          true,
+          wallet,
+          [owner],
+          { from: owner }
+        );
+
+        await assertEventEmitted(loopringModule, "AccountUpdated");
```
adamchen0510: We should check the msg.sender in AccountCreated(emit in exchange) is wallet

packages/hebao_v1/contracts/test/DummyExchange.sol
kongliangzhong: All the events I checked in test/testLoopringModule.ts were emitted in LoopringModule.sol.
So I think I will just remove all the events in DummyExchange.sol

packages/hebao_v1/test/testLoopringModule.ts
kongliangzhong: Updated. please take another look.

Issue #1086, opened 2020-04-14
[Hebao]ConfirmGuardian discussion


adamchen0510:
In order of user experience, the confirmGuardianAddition and confirmGuardianRemoval need to be called immediately after user add or remove guardian, but would fail in contract as too early.
Relay can store user behavior and confirm after pendingPeriod passed for the wallet owner, but this would be failed as the signer must be the wallet owner but can not be anyone else.
![image](https://user-images.githubusercontent.com/13066091/79228069-710a1580-7e93-11ea-8934-39bd0bef286a.png)
We need support confirmGuardianAddition and confirmGuardianRemoval called by anyone(especially the relayer), but need to consider if there are some security issues.

Brechtpd:
I think we can allow `confirmGuardianAddition` and  `confirmGuardianRemoval` to be called by anyone. I don't see any security issue with it as I don't think it increases security at all (if the wallet owner keys are compromised than it doesn't matter anyway, and that's the only way an unwanted guardian can be added).

Another option would be to let the wallet owner also immediately sign the confirm transactions so the relayer can do the confirmation when it is possible, but I don't see how this is any better than just allowing anyone to confirm.

adamchen0510:
Yes, just allow confirmGuardianAddition and confirmGuardianRemoval to be called by anyone is better. 
If relay store the signed transaction it mostly will be fail as nonce issue (
if use the nonce increase way, the nonce would be increased if wallet owner do some operation before pendingPeriod passed).

PR #1088, opened 2020-04-15
[HEBAO_v1] fix issue #1086

kongliangzhong:
issue #1086

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
@@ -93,7 +93,6 @@ contract GuardianModule is SecurityModule
         external
         nonReentrant
         onlyWhenWalletUnlocked(wallet)
-        onlyFromMetaTxOrWalletOwner(wallet)
     {
         uint confirmStart = pendingAdditions[wallet][guardian][group];
         require(confirmStart != 0, "NOT_PENDING");
@@ -142,7 +141,6 @@ contract GuardianModule is SecurityModule
         external
         nonReentrant
         onlyWhenWalletUnlocked(wallet)
-        onlyFromMetaTxOrWalletOwner(wallet)
     {
         uint confirmStart = pendingRemovals[wallet][guardian];
         require(confirmStart != 0, "NOT_PENDING");
--- a/packages/hebao_v1/test/testGuardians.ts
+++ b/packages/hebao_v1/test/testGuardians.ts
@@ -452,5 +452,175 @@ contract("GuardiansModule", (accounts: string[]) => {
         );
       }
     );
+
+    it(
+      description(
+        "owner should not be able to confirm guardian removal after expired"
+      ),
+      async () => {
+        useMetaTx = metaTx;
+        const owner = ctx.owners[0];
+        const { wallet } = await createWallet(ctx, owner);
+        const group = 0;
+
+        // The first guardian is added immediately (so cannot be cancelled)
+        await addGuardianChecked(owner, wallet, ctx.guardians[0], group);
+
+        // Now remove the guardian
+        await executeTransaction(
+          ctx.guardianModule.contract.methods.removeGuardian(
+            wallet,
+            ctx.guardians[0]
+          ),
+          ctx,
+          useMetaTx,
+          wallet,
+          [owner],
+          { from: owner }
+        );
+
+        // Skip forward `pendingPeriod` + `confirmPeriod` seconds
+        await advanceTimeAndBlockAsync(pendingPeriod + confirmPeriod);
+
+        // Try to confirm
+        await expectThrow(
+          executeTransaction(
+            ctx.guardianModule.contract.methods.confirmGuardianRemoval(
+              wallet,
+              ctx.guardians[0]
+            ),
+            ctx,
+            useMetaTx,
+            wallet,
+            [owner],
+            { from: owner }
+          ),
+          "TOO_EARLY_OR_EXPIRED"
+        );
+
+        // Now cancel
+        await executeTransaction(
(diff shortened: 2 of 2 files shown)


packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+            ctx.guardians[0]
+          ),
+          ctx,
+          useMetaTx,
+          wallet,
+          [owner],
+          { from: owner }
+        );
+
+        // Make sure the expired guardian is not a guardian
```
adamchen0510, line 514: comment issue, it's still guardian

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+          group
+        ),
+        ctx,
+        useMetaTx,
+        wallet,
+        [owner],
+        { from: owner }
+      );
+
+      // Skip forward `pendingPeriod` + `confirmPeriod` seconds
```
adamchen0510, line 552: comment issue

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+          ctx.guardians[0]
+        ),
+        ctx,
+        useMetaTx,
+        wallet,
+        [owner],
+        { from: owner }
+      );
+
+      // Skip forward `pendingPeriod` + `confirmPeriod` seconds
```
adamchen0510, line 604: comment issue

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+          ctx.guardians[0]
+        ),
+        ctx,
+        useMetaTx,
+        wallet,
+        [user],
+        { from: user }
+      );
+
+      // Make sure the expired guardian is not a guardian
```
adamchen0510, line 621: comment issue

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+        { from: owner }
+      );
+
+      // Skip forward `pendingPeriod` + 1 seconds
+      await advanceTimeAndBlockAsync(pendingPeriod + 1);
+
+      const user = accounts[10];
+
+      // Try to confirm
+      await executeTransaction(
```
Brechtpd, line 558: This will use a meta tx signed by the wallet owner to confirm the guardian, so this isn't really a test that anyone can confirm the guardian addition. I think a better way to test is is to call `confirmGuardianAddition` immediately on the module instead of via a meta tx.

A meta tx NEEDS to be signed by the wallet owner or one of his guardians in all cases. The reason is that otherwise anyone would be able to completely use up the daily limit of the wallet by creating dummy meta txs on the wallet and paying themselves fees from the wallet. So the relayer can't just create a meta tx by himself for the wallet.

The downside is that the relayer will end up paying the gas costs for the confirmation. I think this is fine as the relayer can request a higher than necessary fee payment for the initial guardian addition. You can think of other solutions but I think this would add unnecessary complexity.

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -452,5 +452,178 @@ contract("GuardiansModule", (accounts: string[]) => {
...
+        [owner],
+        { from: owner }
+      );
+
+      // Skip forward `pendingPeriod` + 1 seconds
+      await advanceTimeAndBlockAsync(pendingPeriod + 1);
+      const user = accounts[10];
+
+      // Try to confirm
+      await executeTransaction(
```
Brechtpd, line 609: Same as above with `confirmGuardianAddition`.

packages/hebao_v1/test/testGuardians.ts
adamchen0510: Kongliang and I have discussed about this, confirmGuardianAddition and confirmGuardianRemoval now just use useMetaTx=false case, metaTx of others can not pass in current contract as areMetaTxSignersAuthorized check and verifySignatures check(only for wallet owner now).

By keep current design, relay use itself key sign the tx directly call confirmGuardianAddition and confirmGuardianRemoval is a better way, and relay would pay for the gas.

Another way is store the data from owner by using nonce=0 case, but need handle the next add same guardian caused the same hash issue. so the previous way is better.

Brechtpd: Ah I didn't notice the `useMetaTx = false`. But it does seem like `from: owner` still needs to be changed to a different address (the "relayer").

> but need handle the next add same guardian caused the same hash issue. so the previous way is better.

Yeah, I think we should add a `validUntil` value to `MetaTransaction` to solve this (even if not really needed in this case). This value can be used to limit the validity of the transaction in time, but can also be used to make it unique so the hash is different even if the actual transaction is the same. (I think you could do that as well now by slightly changing some of the other data like `gasLimit` but that would be hacky).

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -93,7 +93,6 @@ contract GuardianModule is SecurityModule
         external
         nonReentrant
         onlyWhenWalletUnlocked(wallet)
-        onlyFromMetaTxOrWalletOwner(wallet)
```
dong77, line 96: 为什么要去掉这个？没有安全问题？

PR #1090, opened 2020-04-15
Update DESIGN - Efficiently using orderIDs

Brechtpd:
Could be useful info for people using the API.


Brechtpd:
How does the `/api/v2/orderId` API work (`Get next order ID`)? Does it simply increment the orderID each time it is called? It would be great if it would internally use method 1. (which would also be more future proof for how things work in 3.5).

mmmmmagina:
> How does the `/api/v2/orderId` API work (`Get next order ID`)? Does it simply increment the orderID each time it is called? It would be great if it would internally use method 1. (which would also be more future proof for how things work in 3.5).

We use method 1 roughly, but there are some details:
  1. For orderId> 2 ^ 14, it is necessary to determine whether orderId-n * 2 ^ 14 (n = 1,2,3 ...) has been completed (complete filled or cancelled), if there is no completion, the new order cannot be submitted
  2. The case where orderId can not be reused is also based on the complexity of logical judgment after orderId exceeds 2 ^ 14. At the same time, when there are a large number of orders, it will be more time-consuming and complicated to retrieve the next order that can be reused in the database. , Will lead to longer order submitting time, and now when our market maker account's orderId is close to 2 ^ 20, will change to another accountId

Method 2 looks more complicated in engineering, so we don’t consider implementing it for now.

PR #1093, opened 2020-04-16
[HEBAO_V1] revert PR #789

kongliangzhong:


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -50,6 +50,8 @@ abstract contract MetaTxModule is BaseModule
     using SignatureUtil for bytes32;
     using BytesUtil     for bytes;
 
+    uint   constant public   BLOCK_BOUND = 100;
```
Brechtpd, line 53: This is pretty low, no? If the current block number at time of signing is put into the nonce than the transaction is only valid for around 20 minutes. If Ethereum is congested for even ~15 minutes people would have to re-sign their transactions.

dong77: will 1000 be a reasonable value?

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -472,7 +474,8 @@ abstract contract MetaTxModule is BaseModule
             require(!wallets[wallet].metaTxHash[metaTxHash], "INVALID_HASH");
             wallets[wallet].metaTxHash[metaTxHash] = true;
         } else {
-            require(nonce == wallets[wallet].nonce + 1, "INVALID_NONCE");
+            require(nonce > wallets[wallet].nonce, "NONCE_TOO_SMALL");
+            require((nonce >> 128) <= (block.number + BLOCK_BOUND), "NONCE_TOO_LARGE");
```
Brechtpd, line 478: This time limit is only usable for transactions using nonces, not in the case when `nonce = 0`. If we have a time limit for transaction with nonces I think it probably makes sense to also have a time limit for transactions without nonces as well especially because they are valid for an unlimited time? (transactions with nonces can be cancelled in certain ways while transactions without nonces cannot)

I don't think this is really important, just pointing out that there is now a strange difference between the 2 types of transactions.

packages/hebao_v1/contracts/base/MetaTxModule.sol
Brechtpd: I think that should be okay. Though on black Thursday Ethereum was congested (gas price > 50 Gwei) for almost a day I think? So in exceptional cases transactions could still expire I guess, but that may not be a bad thing.

adamchen0510: BLOCK_BOUND used for nonce not too large, limit the client don't fill the value too large.
It can not happen re-sign case.

packages/hebao_v1/contracts/base/MetaTxModule.sol
adamchen0510: I can not understand your words, maybe your misunderstand the BLOCK_BOUND using here? It's not time limit just make sure the nonce not too large and and reach the max value.

kongliangzhong: @Brechtpd , Do we support transactions without nonces? As far as I understand, all meta transactions must have a nonce.
And also, what is the use of BLOCK_BOUND, Maybe just `require((nonce >> 128) <= (block.number), "NONCE_TOO_LARGE")` will be okay?

Brechtpd: > Do we support transactions without nonces? As far as I understand, all meta transactions must have a nonce.

All meta transactions have a nonce value, but when `nonce == 0` we don't do anything with the nonce onchain because the hash is used to prevent replay attacks. It's just above the code here (line 473). Transactions with `nonce == 0` can be executed in any order and the nonce value of previous meta transactions don't matter.

> And also, what is the use of BLOCK_BOUND, Maybe just require((nonce >> 128) <= (block.number), "NONCE_TOO_LARGE") will be okay?

I think the idea behind this code originated from argent so I can only guess at what the intention is. But I think it's basically just a `validUntil `value (in blocks instead of in seconds) + `nonce` value packed inside a single `uint` value. Why is it in blocks instead of in seconds? Why pack it into a single value? I'm not 100% sure and I don't really like it (which is why I proposed to change it :). I guess the benefit of packing it into a single value like this is that in most cases you don't have to keep track of the nonce because you can just set the current block number when the transactions is signed (with the actual nonce part 0) and `nonce > wallets[wallet].nonce` will still be true because the block number part is stored in the upper bits.

>  It's not time limit just make sure the nonce not too large to reach the max value.

If this would be the case than the code would not check the left part against `block.number + BLOCK_BOUND` but indeed something like `(nonce >> 128) <= (block.number)` like @kongliangzhong said. The fact that ` + BLOCK_BOUND` is done indicates that it's used for something time related.

But we don't really have to try to find out what the previous code based on argent did haha. :) We just need to have a mechanism that is workable for the client and makes sense. What about something like this:

```
require(metaTransaction.validUntil <= now, "TX_EXPIRED");
if (nonce == 0) {
    require(!wallets[wallet].metaTxHash[metaTxHash], "INVALID_HASH");
    wallets[wallet].metaTxHash[metaTxHash] = true;
} else {
    // Option 1 (current code):
    require(nonce == wallets[wallet].nonce + 1, "INVALID_NONCE");
    // Option 2 (previously proposed to keep things simpler):
    require(nonce > wallets[wallet].nonce, "INVALID_NONCE");

    wallets[wallet].nonce = nonce;
}
```

To decide which meta transaction type to use: Is the order the transactions are executed in important?
- No -> Use `nonce == 0`
- Yes -> Use `nonce (>)= wallets[wallet].nonce + 1`

kongliangzhong: Yes, I agree with you, have a validUntil field would be better.

kongliangzhong: @Brechtpd , I'd added the validUntil field to MetaTransaction, please take another look.

kongliangzhong: @Brechtpd , I think the BLOCK_BOUND variable has nothing to do with validUntil or timestamp, so I will revert the validUntil changes. If validUntil is neccessary, I can add it in another PR.

PR #1110, opened 2020-04-21
Sync EIP712.sol with with Loopring3.5's same named library

dong77:


Brechtpd:
Added a commit that fixes the tests because of the different data that needs to be signed.

dong77:
Feel free to merge the PR.

Issue #1099, opened 2020-04-17
[Hebao]why need confirmGuardianAddition and confirmGuardianRemoval


adamchen0510:
Just like addToWhitelist, the guardian will be effective after pendingPeriod, don't need confirmGuardianAddition.
And confirmGuardianRemoval is the same.

Brechtpd:
If I remember correctly it's mostly because of efficiency and simplicity. Because we have to run over all guardians sometimes and the list would contain guardians in different states. So even to just know how many guardians a wallet has you'd have to run over all guardians and check if the guardian is actually valid at the time or not. For example, we currently do this:

```
uint count = controller.securityStore().numGuardians(wallet);
require(count < MAX_GUARDIANS, "TOO_MANY_GUARDIANS");
```

This will have to check how many guardians are now active or could be active at some point of time in the future. (Hmm actually the current code isn't safe either, but can easily fixed. Will create a PR).

For whitelisted addresses we don't have that problem because we only need to check if a single address is whitelisted or not, we don't care about how many whitelisted addresses there are etc...

But sure, if you think it's worth it to not need any confirmations at the cost of some extra complexity than I'm okay with that.

adamchen0510:
I see, by add confirmGuardianAddition, it would be efficient and simple for contract.

If just for count numGuardians, we can count guardians group num and pendingGuardian group num(which can be treat as guardian too) when addGuardian.
I know numGuardians is just an example, if we decide to remove confirm step, we need go through the codes.

For confirmGuardianAddition, it decided by our design, whether guardian can reject the addGuardian request, if can, it still need confirmGuardianAddition(else need every one can call cancelGuardianAddition to cancel the guardian addition).
But the first guardian don't need guardian confirm, so if need guardian accept this is a problem.

For confirmGuardianRemoval, it absolutely don't need guardian confirm.

By these two confirmation, relayer need create task for each wallet to confirm for the wallet. we need balance the effort.
It's just a discussion, and it's not a small task if decide to remove these two api and need more consideration.

kongliangzhong:
resolved.

PR #1115, opened 2020-04-24
[HEBAO-V1] enable auto comfirmation for guardian addition and removal

kongliangzhong:
Fix issue #1099 
All tests passed.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
@@ -61,6 +61,9 @@ contract GuardianModule is SecurityModule
     {
         require(guardian != address(0), "ZERO_ADDRESS");
         require(group < GuardianUtils.MAX_NUM_GROUPS(), "INVALID_GROUP");
+        uint numGuardians = controller.securityStore().numGuardiansWithPending(wallet);
+        require(numGuardians < MAX_GUARDIANS, "TOO_MANY_GUARDIANS");
+
         uint effectiveTime = now;
         if (controller.securityStore().numGuardians(wallet) > 0) {
             effectiveTime = now + pendingPeriod;
--- a/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianModule.sol
@@ -31,30 +31,21 @@ contract GuardianModule is SecurityModule
 {
     uint constant public MAX_GUARDIANS = 20;
     uint public pendingPeriod;
-    uint public confirmPeriod;
 
-    mapping (address => mapping(address => mapping(uint => uint))) public pendingAdditions;
-    mapping (address => mapping(address => uint)) public pendingRemovals;
-
-    event GuardianAdditionPending   (address indexed wallet, address indexed guardian, uint group, uint confirmAfter);
-    event GuardianAdded             (address indexed wallet, address indexed guardian, uint group);
+    event GuardianAdded             (address indexed wallet, address indexed guardian, uint group, uint effectiveTime);
     event GuardianAdditionCancelled (address indexed wallet, address indexed guardian);
-
-    event GuardianRemovalPending    (address indexed wallet, address indexed guardian, uint confirmAfter);
-    event GuardianRemoved           (address indexed wallet, address indexed guardian);
+    event GuardianRemoved           (address indexed wallet, address indexed guardian, uint removalEffectiveTime);
     event GuardianRemovalCancelled  (address indexed wallet, address indexed guardian);
 
     constructor(
         Controller _controller,
-        uint       _pendingPeriod,
-        uint       _confirmPeriod
+        uint       _pendingPeriod
         )
         public
         SecurityModule(_controller)
     {
-        require(_pendingPeriod > 0 && _confirmPeriod > 0, "INVALID_DELAY");
+        require(_pendingPeriod > 0, "INVALID_DELAY");
         pendingPeriod = _pendingPeriod;
-        confirmPeriod = _confirmPeriod;
     }
 
     function addGuardian(
@@ -70,53 +61,24 @@ contract GuardianModule is SecurityModule
     {
         require(guardian != address(0), "ZERO_ADDRESS");
         require(group < GuardianUtils.MAX_NUM_GROUPS(), "INVALID_GROUP");
-
-        if (controller.securityStore().numGuardians(wallet) == 0) {
-            controller.securityStore().addOrUpdateGuardian(wallet, guardian, group);
-            emit GuardianAdded(wallet, guardian, group);
-        } else {
-            uint confirmStart = pendingAdditions[wallet][guardian][group];
-            require(confirmStart == 0 || now > confirmStart + confirmPeriod, "ALREADY_PENDING");
-            pendingAdditions[wallet][guardian][group] = now + pendingPeriod;
-            emit GuardianAdditionPending(wallet, guardian, group, now + pendingPeriod);
+        uint effectiveTime = now;
+        if (controller.securityStore().numGuardians(wallet) > 0) {
+            effectiveTime = now + pendingPeriod;
         }
-    }
-
-    function confirmGuardianAddition(
-        address wallet,
-        address guardian,
-        uint    group
-        )
-        external
-        nonReentrant
-        onlyWhenWalletUnlocked(wallet)
-    {
-        uint confirmStart = pendingAdditions[wallet][guardian][group];
-        require(confirmStart != 0, "NOT_PENDING");
-        require(now >= confirmStart && now < confirmStart + confirmPeriod, "TOO_EARLY_OR_EXPIRED");
-        controller.securityStore().addOrUpdateGuardian(wallet, guardian, group);
-
-        // Now check if we don't have too many guardians active at once
-        uint count = controller.securityStore().numGuardians(wallet);
-        require(count <= MAX_GUARDIANS, "TOO_MANY_GUARDIANS");
-
-        delete pendingAdditions[wallet][guardian][group];
-        emit GuardianAdded(wallet, guardian, group);
+        controller.securityStore().addOrUpdateGuardian(wallet, guardian, group, effectiveTime);
+        emit GuardianAdded(wallet, guardian, group, now);
     }
 
     function cancelGuardianAddition(
         address wallet,
-        address guardian,
-        uint    group
+        address guardian
         )
         external
         nonReentrant
         onlyWhenWalletUnlocked(wallet)
         onlyFromMetaTxOrWalletOwner(wallet)
     {
-        uint confirmStart = pendingAdditions[wallet][guardian][group];
-        require(confirmStart > 0, "INVALID_GUARDIAN");
-        delete pendingAdditions[wallet][guardian][group];
+        controller.securityStore().cancelGuardianAddition(wallet, guardian);
         emit GuardianAdditionCancelled(wallet, guardian);
     }
 
@@ -130,26 +92,8 @@ contract GuardianModule is SecurityModule
         onlyWalletGuardian(wallet, guardian)
         onlyFromMetaTxOrWalletOwner(wallet)
     {
-        uint confirmStart = pendingRemovals[wallet][guardian];
-        require(confirmStart == 0 || now > confirmStart + confirmPeriod, "ALREADY_PENDING");
-        pendingRemovals[wallet][guardian] = now + pendingPeriod;
-        emit GuardianRemovalPending(wallet, guardian, now + pendingPeriod);
-    }
-
-    function confirmGuardianRemoval(
-        address wallet,
-        address guardian
-        )
-        external
-        nonReentrant
-        onlyWhenWalletUnlocked(wallet)
-    {
-        uint confirmStart = pendingRemovals[wallet][guardian];
-        require(confirmStart != 0, "NOT_PENDING");
-        require(now >= confirmStart && now < confirmStart + confirmPeriod, "TOO_EARLY_OR_EXPIRED");
-        controller.securityStore().removeGuardian(wallet, guardian);
-        delete pendingRemovals[wallet][guardian];
-        emit GuardianRemoved(wallet, guardian);
+        controller.securityStore().removeGuardian(wallet, guardian, now + pendingPeriod);
+        emit GuardianRemoved(wallet, guardian, now + pendingPeriod);
     }
 
     function cancelGuardianRemoval(
@@ -161,10 +105,21 @@ contract GuardianModule is SecurityModule
         onlyWhenWalletUnlocked(wallet)
         onlyFromMetaTxOrWalletOwner(wallet)
     {
-        uint confirmStart = pendingRemovals[wallet][guardian];
-        require(confirmStart > 0, "INVALID_GUARDIAN");
-        delete pendingRemovals[wallet][guardian];
+        controller.securityStore().cancelGuardianRemoval(wallet, guardian);
         emit GuardianRemovalCancelled(wallet, guardian);
+    }
+
+    function cleanRemovedGuardians(
+        address wallet,
+        uint from,
+        uint to
+        )
+        external
+        nonReentrant
+        onlyWhenWalletUnlocked(wallet)
+        onlyFromMetaTxOrWalletOwner(wallet)
+    {
+        controller.securityStore().cleanRemovedGuardians(wallet, from, to);
     }
 
     function extractMetaTxSigners(
@@ -180,9 +135,7 @@ contract GuardianModule is SecurityModule
         if (method == this.addGuardian.selector ||
             method == this.removeGuardian.selector ||
             method == this.cancelGuardianAddition.selector ||
-            method == this.cancelGuardianRemoval.selector ||
-            method == this.confirmGuardianAddition.selector ||
-            method == this.confirmGuardianRemoval.selector) {
+            method == this.cancelGuardianRemoval.selector) {
             signers = new address[](1);
             signers[0] = Wallet(wallet).owner();
         } else {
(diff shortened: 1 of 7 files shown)


packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -70,53 +61,24 @@ contract GuardianModule is SecurityModule
...
-        controller.securityStore().addOrUpdateGuardian(wallet, guardian, group);
-
-        // Now check if we don't have too many guardians active at once
-        uint count = controller.securityStore().numGuardians(wallet);
-        require(count <= MAX_GUARDIANS, "TOO_MANY_GUARDIANS");
-
-        delete pendingAdditions[wallet][guardian][group];
-        emit GuardianAdded(wallet, guardian, group);
+        controller.securityStore().addOrUpdateGuardian(wallet, guardian, group, effectiveTime);
+        emit GuardianAdded(wallet, guardian, group, now);
```
adamchen0510, line 69: now -> effectiveTime

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -53,46 +55,93 @@ contract SecurityStore is DataStore
...
+        returns (Data.Guardian memory)
     {
-        uint index = wallets[wallet].guardianIdx[_guardian];
+        Wallet storage w = wallets[wallet];
+        uint index = w.guardianIdx[guardian];
         require(index > 0, "NOT_A_GUARDIAN");
-        guardian = wallets[wallet].guardians[index-1];
+        require(
+            w.guardianEffectiveTime[guardian] >= now &&
+            w.guardianRemovalEffectiveTime[guardian] < now,
```
Brechtpd, line 76: Maybe best to create a `isGuardianInternal` that contains this logic as it is duplicated at a couple of places.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -53,46 +55,93 @@ contract SecurityStore is DataStore
...
+        effectiveGuardians = new Data.Guardian[](w.guardians.length);
+        uint index = 0;
+        for (uint i = 0; i < w.guardians.length; i++) {
+            Data.Guardian memory g = w.guardians[i];
+            if (w.guardianEffectiveTime[g.addr] >= now &&
+                w.guardianRemovalEffectiveTime[g.addr] < now) {
+                effectiveGuardians[index] = g;
+                index ++;
+            }
+        }
```
Brechtpd, line 97: You can add `assembly { mstore(effectiveGuardians, index) }` so the array that is returned has the correct length.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -53,46 +55,93 @@ contract SecurityStore is DataStore
...
+        Wallet storage w = wallets[wallet];
+        uint num = 0;
+        for (uint i = 0; i < w.guardians.length; i++) {
+            Data.Guardian memory g = w.guardians[i];
+            if (w.guardianEffectiveTime[g.addr] >= now &&
+                w.guardianRemovalEffectiveTime[g.addr] < now) {
+                num ++;
+            }
+        }
+        return num;
```
Brechtpd, line 114: Maybe this can just do `return guardians(wallet).length;` which won't be significantly more expensive.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -105,22 +154,40 @@ contract SecurityStore is DataStore
...
+            require(
+                 w.guardianRemovalEffectiveTime[guardian] == 0 ||
+                 w.guardianRemovalEffectiveTime[guardian] >= now,
+                 "PENDING_REMOVAL_GUARDIAN"
+            );
+
             // Update the guardian
             w.guardians[pos-1].group = group;
+            w.guardianEffectiveTime[guardian] = effectiveTime;
+            delete w.guardianRemovalEffectiveTime[guardian];
```
Brechtpd, line 170: Seeing the code like this it seems like we can't allow updating the guardian because of security reasons:
- The attacker would currently would be able to disable the guardian immediately (because `guardianEffectiveTime` is set in the future) which could be dangerous.
- Even if `guardianEffectiveTime` is left unchanged (the previous behaviour) an attacker would be able to update the group number of all guardians immediately which he could use to his advantage.

So it seems like we should disable allowing the guardian group to be updated, otherwise we'd need a separate time limit for the group number so it can't be changed immediately.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -182,4 +289,30 @@ contract SecurityStore is DataStore
         wallets[wallet].inheritor = who;
         wallets[wallet].lastActive = uint128(now);
     }
-}
\ No newline at end of file
+
+    function cleanRemovedGuardians(address wallet)
```
adamchen0510, line 293: Can add a method or add some codes in this function and named function as CleanUpGuardians and also add effective guardian to guardians array.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -130,6 +197,46 @@ contract SecurityStore is DataStore
         }
         w.guardians.pop();
         delete w.guardianIdx[guardian];
+        delete w.guardianRemovalEffectiveTime[guardian];
```
Brechtpd, line 200: `delete w.guardianEffectiveTime[guardian];`?

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -182,4 +289,30 @@ contract SecurityStore is DataStore
...
+
+    function cleanRemovedGuardians(address wallet)
+        private
+    {
+        Wallet storage w = wallets[wallet];
+
+        for (uint i = 0; i < w.guardians.length; i ++) {
+            if (i >= w.guardians.length) {
+                break;
+            }
```
Brechtpd, line 301: Shouldn't be needed.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -182,4 +289,30 @@ contract SecurityStore is DataStore
...
+            if (w.guardianRemovalEffectiveTime[guardian.addr] >= now) {
+                Data.Guardian memory lastGuardian = w.guardians[w.guardians.length - 1];
+
+                if (guardian.addr != lastGuardian.addr) {
+                    w.guardians[i] = lastGuardian;
+                    w.guardianIdx[lastGuardian.addr] = i + 1;
+                }
+                w.guardians.pop();
+                delete w.guardianIdx[guardian.addr];
+                delete w.guardianRemovalEffectiveTime[guardian.addr];
```
Brechtpd, line 312: Also `delete w.guardianEffectiveTime[guardian.addr];`?

packages/hebao_v1/contracts/stores/SecurityStore.sol
adamchen0510, line 293: I think we can keep pendingAdditions and pendingRemovals, addGuardian to pendingAdditions first(except the first one), removeGuardian to pendingRemovals first.
And then refine this function as CleanUpGuardians which called before each public function to move pendingRemovals to guardians array and delete pendingRemovals from guardians array, and also delete themselves.
By this, we dont need loop guardians each time to judge the effectivetime to calc the numGuaridan and only calc the guardians length plus pendingAdditions and pendingRemovals.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -53,46 +55,93 @@ contract SecurityStore is DataStore
         view
         returns (bool)
     {
-        return wallets[wallet].guardianIdx[guardian] > 0;
+        return wallets[wallet].guardianIdx[guardian] > 0 &&
+            wallets[wallet].guardianEffectiveTime[guardian] >= now &&
+            wallets[wallet].guardianRemovalEffectiveTime[guardian] < now;
```
Brechtpd, line 60: If a guardian is added in `addOrUpdateGuardian` then `guardianRemovalEffectiveTime` is not set, so it will be 0. So either the logic here needs to be updated or `guardianRemovalEffectiveTime` needs to be set to some large value.

Depending on the implementation maybe `guardianValidFrom` and `guardianValidUntil` may be better names.

kongliangzhong: w.guardianRemovalEffectiveTime[guardian] == 0 then it is < now, I can't see any problem here.
Yes, guardianValidFrom and guardianValidUntil are better names, I will update it.

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: 1. I think we should not allow a user to submit another removal when there is pending removal.
2. maybe we should not allow a user to change groups. all he can do is remove it and add a new one.

Brechtpd: 1. Any reason in particular you wouldn't allow multiple removals? It will still take `pendingPeriod` to remove them all so the owner still has time to recover from this and cancel all the removals in that time period.
2. Yes, I think that would be best.

packages/hebao_v1/contracts/stores/SecurityStore.sol
Brechtpd: Oh right sorry. For some reason I always get these time related things wrong. :)

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: Ok, then I will disable the guardian updating function.

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: OK

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -184,26 +151,24 @@ contract SecurityStore is DataStore
...
         Data.Guardian memory lastGuardian = w.guardians[w.guardians.length - 1];
-
         if (guardian != lastGuardian.addr) {
             w.guardians[idx - 1] = lastGuardian;
             w.guardianIdx[lastGuardian.addr] = idx;
         }
         w.guardians.pop();
         delete w.guardianIdx[guardian];
-        delete w.guardianRemovalEffectiveTime[guardian];
+        delete w.guardianValidUntil[guardian];
```
Brechtpd, line 165: I think this should be `delete w.guardianValidSince[guardian];`.

kongliangzhong: Yes, you are right.

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
letsgoustc: now --> effectiveTime?

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -130,6 +176,66 @@ contract SecurityStore is DataStore
...
+            Data.Guardian memory guardian = w.guardians[i];
+            Data.Guardian memory lastGuardian = w.guardians[w.guardians.length - 1];
+
+            if (guardian.addr != lastGuardian.addr) {
+                w.guardians[i] = lastGuardian;
+                w.guardianIdx[lastGuardian.addr] = i;
+            }
+            w.guardians.pop();
+            delete w.guardianIdx[guardian.addr];
+            delete w.guardianRemovalEffectiveTime[guardian.addr];
```
letsgoustc: Do we need to delete guardianEffectiveTime also?

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -130,6 +176,66 @@ contract SecurityStore is DataStore
         }
         w.guardians.pop();
         delete w.guardianIdx[guardian];
+        delete w.guardianRemovalEffectiveTime[guardian];
```
letsgoustc: Because here we just cancel add a guardian, so I think we should also delete it from guardianEffectiveTime.
delete w.guardianEffectiveTime[guardian];

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -70,53 +61,27 @@ contract GuardianModule is SecurityModule
...
-        if (controller.securityStore().numGuardians(wallet) == 0) {
-            controller.securityStore().addOrUpdateGuardian(wallet, guardian, group);
-            emit GuardianAdded(wallet, guardian, group);
-        } else {
-            uint confirmStart = pendingAdditions[wallet][guardian][group];
-            require(confirmStart == 0 || now > confirmStart + confirmPeriod, "ALREADY_PENDING");
-            pendingAdditions[wallet][guardian][group] = now + pendingPeriod;
-            emit GuardianAdditionPending(wallet, guardian, group, now + pendingPeriod);
+        uint effectiveTime = now;
+        if (controller.securityStore().numGuardians(wallet) > 0) {
```
Brechtpd, line 68: May as well just use `numGuardians` here which will work the same.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -182,4 +266,62 @@ contract SecurityStore is DataStore
...
+                if (g.addr != lastGuardian.addr) {
+                    w.guardians[i] = lastGuardian;
+                    w.guardianIdx[lastGuardian.addr] = i + 1;
+                }
+                w.guardians.pop();
+                delete w.guardianIdx[g.addr];
+                delete w.guardianValidSince[g.addr];
+                delete w.guardianValidUntil[g.addr];
+            }
+        }
```
Brechtpd, line 289: This only cleans up the guardians so it doesn't really matter, but this loop may skip cleaning up some guardians.

For example, if the guardians are like this: (valid, invalid, invalid) the loop will switch guardian[1] with guardian[2] and stop looping. But guardian[2] is also invalid so we'd have to check the guardian again at each index where there was a removal. You can do that by not incrementing `i` when a guardian was removed, but the more standard way to do this is to loop backwards over the list:
```
for (int i = int(w.guardians.length) - 1; i >= 0; i--) {
    Data.Guardian memory g = w.guardians[uint(i)];
    if (isGuardianRemoved(wallet, g.addr)) {
        Data.Guardian memory lastGuardian = w.guardians[w.guardians.length - 1];

        if (g.addr != lastGuardian.addr) {
            w.guardians[uint(i)] = lastGuardian;
            w.guardianIdx[lastGuardian.addr] = uint(i) + 1;
        }
        w.guardians.pop();
        delete w.guardianIdx[g.addr];
        delete w.guardianValidSince[g.addr];
        delete w.guardianValidUntil[g.addr];
    }
}
```

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -182,4 +266,62 @@ contract SecurityStore is DataStore
...
+    }
+
+    function isGuardianValid(address wallet, address guardian)
+        view
+        private
```
Brechtpd, line 294: I've got a formatting warning here that `private` should be declared before `view`. Also for the functions below this one.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -53,83 +55,165 @@ contract SecurityStore is DataStore
...
+        )
+        public
+        view
+        returns (uint, uint)
+    {
+        return (wallets[wallet].guardianValidSince[guardian],
+                wallets[wallet].guardianValidUntil[guardian]);
+    }
+
+
```
Brechtpd, line 74: If we want to expose the timestamps anyway, what about just adding the `validSince`and `validUntil` to `Data.Guardian`? This way all data for the guardian is nicely together is also directly available in the other getters we already have for the guardians.

packages/hebao_v1/test/testGuardians.ts
```diff
@@ -50,16 +49,14 @@ contract("GuardiansModule", (accounts: string[]) => {
...
-    ).toNumber();
+
   });
 
   beforeEach(async () => {
     ctx = await createContext(defaultCtx);
   });
 
-  [false, true].forEach(function(metaTx) {
+  [false/*, true*/].forEach(function(metaTx) {
```
Brechtpd, line 59: Seems like meta tx can be enabled here?

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
kongliangzhong: fixed

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: updated

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: fixed

packages/hebao_v1/contracts/stores/SecurityStore.sol
kongliangzhong: Yes, I agree. updated.

packages/hebao_v1/test/testGuardians.ts
kongliangzhong: Sorry, I disabled it for debugging. enabled.

kongliangzhong:
About the getGuardian function in SecurityStore.sol:
`
    function getGuardian(
        address wallet,
        address guardianAddr
        )
        public
        view
        returns (Data.Guardian memory)
    {
        uint index = wallets[wallet].guardianIdx[guardianAddr];
        if(index > 0) {
            return wallets[wallet].guardians[index-1];
        }
    }
`
I don't use require(index > 0, "...") because I don't want get a revert exception in a view function, so I just return default empty data if index = 0.

@Brechtpd

PR #1128, opened 2020-05-03
[Protocol3] double check reentrancy and add other small improvements

dong77:
Double-check reentrancy guards, and make sure function modifiers are orders as follows:

```
internal/private/public/external
virtual/override
view/nonReentrant
only***
```


packages/loopring_v3/contracts/impl/agents/FastWithdrawalAgent.sol
```diff
@@ -231,16 +231,12 @@ contract FastWithdrawalAgent is ReentrancyGuard
...
-            } else {
-                // ERC20 token
-                success = token.safeTransferFrom(from, to, amount);
-            }
-        }
+        if (amount == 0) return;
+       
+        bool success = token == address(0) ? 
+            to.sendETH(amount, gasleft()) :
+            token.safeTransferFrom(from, to, amount);
```
dong77, line 238: I think the original code has a bug when account == 0

packages/loopring_v3/contracts/iface/ILoopringV3.sol
```diff
@@ -141,13 +141,13 @@ abstract contract ILoopringV3 is ILoopring
     /// @param onchainDataAvailability True if the exchange has on-chain
     ///        data-availability, else false
     /// @return True if the exchange has staked enough, else false
-    function canExchangeCommitBlocks(
+    function canExchangeSubmitBlocks(
```
Brechtpd, line 144: Okay for me, but I kept it like this so that 3.5 exchanges are compatible with the existing Loopring contracts.

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -200,14 +199,12 @@ contract BasicDepositContract is IDepositContract, ReentrancyGuard
...
     {
-        if (amount > 0) {
-            if (token == address(0)) {
-                // ETH
-                to.sendETHAndVerify(amount, gasleft());
-            } else {
-                // ERC20 token
-                token.safeTransferAndVerify(to, amount);
-            }
+        if (amount == 0) return;
```
Brechtpd, line 202: To be consistent I would add brackets here, because nowhere else I think we do this (I always use brackets).

packages/loopring_v3/contracts/impl/agents/FastWithdrawalAgent.sol
Brechtpd: Yes you're right. Though `amount == 0` doesn't do much useful. It could actually just use the verify variants of the functions:
```
if (amount > 0) {
    if (token == address(0)) {
        // ETH
        to.sendETHAndVerify(amount, gasleft());
    } else {
        // ERC20 token
        token.safeTransferAndVerify(to, amount);
    }
}
```

Brechtpd:
Is there some style guide you followed for the order of the modifiers? I tried to follow the logic that the function declaration is the same as the function implementation + extra modifiers, so if you have 
```
function deposit()
    external
    view
    payable;
```

The implementation is the same modifiers + `override` and other implementation specific modifiers:
```
function deposit()
    external
    view
    payable
    override
    nonReentrant
{}
```

I think this is the same style as in the solidity docs, though there aren't that [many examples](https://solidity.readthedocs.io/en/v0.6.6/using-the-compiler.html?highlight=virtual#review-changes):
```
abstract contract Updateable {
    function run() public view virtual returns (bool);
    function update() public virtual;
}
```

So I'm not sure what the standard is, but I think there is some logic to having e.g. `view` before the `virtual`/`override`.

Brechtpd:
I'm not totally sure why the ethsnark module needs to be updated in this PR. It looks like it's an update to the correct code, but I don't know why it would be necessary as it should already be set to the correct branch.

packages/loopring_v3/contracts/iface/ILoopringV3.sol
dong77: I get it. Will revert this back.

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
dong77: done

dong77:
> Is there some style guide you followed for the order of the modifiers? I tried to follow the logic that the function declaration is the same as the function implementation + extra modifiers, so if you have
> 
> ```
> function deposit()
>     external
>     view
>     payable;
> ```
> 
> The implementation is the same modifiers + `override` and other implementation specific modifiers:
> 
> ```
> function deposit()
>     external
>     view
>     payable
>     override
>     nonReentrant
> {}
> ```
> 
> I think this is the same style as in the solidity docs, though there aren't that [many examples](https://solidity.readthedocs.io/en/v0.6.6/using-the-compiler.html?highlight=virtual#review-changes):
> 
> ```
> abstract contract Updateable {
>     function run() public view virtual returns (bool);
>     function update() public virtual;
> }
> ```
> 
> So I'm not sure what the standard is, but I think there is some logic to having e.g. `view` before the `virtual`/`override`.

The main reason for this PR is to make sure we either have `external view` or `external nonReentrant` to avoid reentrancy attach. So put view and nonReentrant below external and make them comparable is easier to identify potential missing cases.

kongliangzhong:
@dong77, I'd checkout ethsnarks from the protocol-35 branch.

PR #1140, opened 2020-05-07
[HEBAO_v1]migration scripts refine

kongliangzhong:


packages/hebao_v1/contracts/modules/dapps/LRCStakingModule.sol
```diff
@@ -37,12 +37,20 @@ contract LRCStakingModule is SubAccountDAppModule
...
+    {
+    }
+
+    function init(
+        address wallet,
+        IUserStakingPool _stakingPool,
+        address          _lrcTokenAddress
+        )
+        external
+        onlyFromWalletOwner(wallet)
```
Brechtpd, line 53: `onlyFromWalletOwner(wallet)` doesn't seem to make much sense?

kongliangzhong: Sorry, I had a misunderstanding about it. I had made this contract Ownable and use onlyOwner modifier for it.

PR #1142, opened 2020-05-07
[Hebao] Signers refactor

Brechtpd:
Likely still contains mistakes but I just tried a couple of things out.

2 commits:
- The first commit only sends the signers next to the meta tx for functions that need a majority. So these functions are kind of special which is a negative.
- The second commit (the diff of this PR) always sends the signers next to the meta tx and verifies if the list is as expected for the different functions. This seems to work pretty well and all things that need to be checked can be done in a single function `verifySigners` so there's no more separate `extractMetaTxSigners` and `areMetaTxSignersAuthorized` which could be error prone as they each could contain special cases for a single function.


packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -112,22 +112,22 @@ contract GuardianModule is SecurityModule
...
 
-    function extractMetaTxSigners(
-        address wallet,
-        bytes4  method,
-        bytes   memory
+    function verifySigners(
+        address   wallet,
+        bytes4    method,
+        bytes     memory /*data*/,
+        address[] memory signers
```
kongliangzhong, line 119: I think the `signers` parameter is unnecessary, instead, users can pass in a guardianGroupId and we can get all the signers from securityStore. And in this way, we don't need to validate signers.

Brechtpd: I don't think that's sufficient because there's too many possibilities. Signers can be from different groups, and each group can have different majority rules. And over all the different groups it is decided if there's a majority or not. Sometimes the owner is needed for a majority, sometimes it is disallowed, other times it is allowed but not required.

By passing the signers separately we have the full flexibility per function without any limitations on the actual function parameters (a limitation which we had until now) which I think is an overall benefit. Now there's not need to add function parameters just so we can extract the signers from it.

kongliangzhong: Okay

dong77:
Some metadata-transaction only methods have 'signers' as a parameter so these methods can have their local logic to check if these signers are sufficient. The current issue is that before we know who will sign, we don't know the signer list, then we cannot construct the `data` for all possible signers to sign in the first place.

Is it possible to zero the `signers` related bytes in the `data` (call it `data2`) and ask potential singers to sign  `data2` instead of `data`. In executeMetaTx, we do `data->data2` as well before verifying signatures, then when we do `address(this).call(data2)` now the actual method will know who actually signed.

We may want to leave enough 0 bytes so we can support many singers. Each method will also need to handle address(0) as signers and simply ignore these addresses.

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -421,41 +435,39 @@ abstract contract MetaTxModule is BaseModule
...
         )
         private
         view
-        returns (address[] memory signers)
+        returns (bool)
     {
         bytes4 method = extractMethod(data);
         if (method == this.executeTransactions.selector) {
-            return extractAddressesFromCallData(data, 1);
+            address[] memory addresses = extractAddressesFromCallData(data, 1);
```
dong77, line 449: why don't we remove `signers` from the data (or the method's ABI)?

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -421,41 +435,39 @@ abstract contract MetaTxModule is BaseModule
...
+            // Check if it matches the list of signers
+            if (addresses.length == signers.length) {
```
dong77, line 451: should we simply make sure addresses[i] == signers[i]?

packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -235,10 +236,12 @@ abstract contract MetaTxModule is BaseModule
     /// @param wallet The wallet used in all transactions.
     /// @param signers The signers needed for all transactions.
```
dong77, line 237: signers The signers needed for all transactions. =>
signers The signers needed for all transactions whose signatures have been verified.

packages/hebao_v1/contracts/base/MetaTxModule.sol
Brechtpd: So not check if the length of both lists are equal? If so, I'm not sure why not?

Brechtpd:
Apologies for some of the formatting changes in random files that are done automatically. I have no idea why the auto formatting done to the code can differ between different commits even though nothing has changed otherwise.

packages/hebao_v1/contracts/modules/dapps/PoolTogetherModule.sol
```diff
@@ -29,15 +29,16 @@ contract PoolTogetherModule is SecurityModule
...
+        address   /*wallet*/,
+        bytes4    /*method*/,
+        bytes     memory /*data*/,
+        address[] memory /*signers*/
         )
         internal
         view
         override
-        returns (address[] memory signers)
+        returns (bool)
```
letsgoustc, line 41: Do we need to return  a default value false?

packages/hebao_v1/contracts/modules/dapps/MakerModule.sol
```diff
@@ -29,15 +29,16 @@ contract MakerModule is SecurityModule
...
+        address   /*wallet*/,
+        bytes4    /*method*/,
+        bytes     memory /*data*/,
+        address[] memory /*signers*/
         )
         internal
         view
         override
-        returns (address[] memory signers)
+        returns (bool)
```
letsgoustc, line 41: Do we need to return  a default value false?

packages/hebao_v1/contracts/modules/dapps/DyDxModule.sol
```diff
@@ -29,15 +29,16 @@ contract DyDxModule is SecurityModule
...
+        address   /*wallet*/,
+        bytes4    /*method*/,
+        bytes     memory /*data*/,
+        address[] memory /*signers*/
         )
         internal
         view
         override
-        returns (address[] memory signers)
+        returns (bool)
```
letsgoustc, line 41: Do we need to return  a default value false?

packages/hebao_v1/contracts/modules/dapps/PoolTogetherModule.sol
Brechtpd: By not returning anything the return value will always be false I think.
But these are just placeholders and the implementation still needs to be done so I don't think it matters either way.

packages/hebao_v1/contracts/base/MetaTxModule.sol
dong77: Brecht, sorry this PR has been merged, if you feel we need to correct things, don't hesitate to create a new PR :)

PR #1145, opened 2020-05-09
[Hebao] Upgrade to solidity 0.6.6 and imporved reentrancy check

dong77:


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -154,12 +154,12 @@ abstract contract MetaTxModule is BaseModule
...
     ///      the 'Checks Effects Interactions' pattern! We do not use `nonReentrant`
     ///      because this function is used to call into the same contract.
+    ///
     /// @param data The raw transaction to be performed on this module.
     /// @param nonce The nonce of this meta transaction. When nonce is 0, this module will
     ///              make sure the transaction's metaTxHash is unique; otherwise, the module
     ///              requires the nonce is greater than the last nonce used by the same
     ///              wallet, but not by more than `block.number * 2^128`.
-    ///
     /// @param gasSetting A list that contains `gasToken` address, `gasPrice`, `gasLimit`,
```
dong77, line 163: this can be passed in as a struct?

Brechtpd: Yeah, sure. If I remember correctly web3 still had some problem with encoding `BN`s in structs pretty recently, but I'm pretty sure that's fixed now (otherwise we'd also have this problem in protocol 3.5).

packages/hebao_v1/contracts/base/BaseModule.sol
```diff
@@ -75,6 +75,11 @@ contract BaseModule is ReentrancyGuard, Module
         _;
     }
 
+    receive() external payable
+    {
+        revert("UNSUPPORTED");
```
Brechtpd, line 80: I think this is pretty much the same as not having the function.

packages/hebao_v1/contracts/base/MetaTxModule.sol
dong77: Not a big deal, I think.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -232,30 +223,55 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
...
-            // -- the 2300 gas limit for send/transfer may case out-of-gas
-            // issue after Istanbul.
-            return;
-        }
-
         address module = methodToModule[msg.sig];
         require(isAddressInSet(MODULE, module), "MODULE_UNAUTHORIZED");
 
-        (bool success, bytes memory returnData) = module.call(msg.data);
+        (bool success, bytes memory returnData) = nonReentrantCall(uint8(1), module, msg.value, msg.data);
```
Brechtpd, line 237: I don't think it's a good idea to make this `nonReentrant`. One of the reasons this functionality is here is so that a contract can call a function on `msg.sender` (which will be this wallet) which will also be the walllet so it is kind of made for reentrant by design. So if the transaction was initiated with `transact` this will fail. This can be a problem with ERC1271 or even token standards that call function on the receiver of tokens.
This function also does 0 state changes because it just forwards calls so I also don't think it's necessary. The functions called here are also only verified module functions which won't do much to begin with.

packages/hebao_v1/contracts/base/BaseModule.sol
dong77: Ok, removed.

packages/hebao_v1/contracts/base/BaseWallet.sol
dong77: OK

PR #1148, opened 2020-05-11
[hebao] callstack based reentrancy check

dong77:
@Brechtpd , I propose this more complex but useful change to reentrancy check. It will allow contracts to control which methods can have reentrancy based on its upstream method selector.

Many tests fail, and I think this also a reason we need to figure out exactly where those reentrancies are.

The code will be cleaner if we know how to get the selector of the current method.

Please let me know what you think.

### DIFF
--- a/packages/hebao_v1/contracts/base/BaseWallet.sol
+++ b/packages/hebao_v1/contracts/base/BaseWallet.sol
@@ -14,7 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
-pragma solidity ^0.6.0;
+pragma solidity ^0.6.6;
 
 import "../lib/ERC20.sol";
 import "../lib/AddressSet.sol";
@@ -122,6 +122,7 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
         override
         onlyOwnerOrModule
     {
+        require(!isAddressInSet(MODULE, _module), "MODULE_ADDED_ALREADY");
         addModuleInternal(_module);
     }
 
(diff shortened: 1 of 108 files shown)


packages/hebao_v1/contracts/lib/ReentrancyGuard.sol
```diff
@@ -26,13 +26,52 @@ pragma solidity ^0.6.6;
 contract ReentrancyGuard
 {
     //The default value must be 0 in order to work behind a proxy.
-    uint private _guardValue;
+    bytes4[] private _callstack;
 
-    modifier nonReentrant()
+    //TODO: It will great if we have a way to get the current selector
```
Brechtpd, line 31: If this is used in a `public`/`external` function it's just `msg.sig`.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -159,7 +159,10 @@ contract BaseWallet is ReentrancyGuard, AddressSet, Wallet
     function bindMethod(bytes4 _method, address _module)
         external
         override
-        nonReentrant
+        reentrantWhitelist(
+            this.bindMethod.selector,
+            this.setup.selector | this.addModule.selector | this.removeModule.selector
+        )
```
Brechtpd, line 165: Because the function is `onlyModule`, isn't this just basically trying to ensure the module works correctly? Does is really matter that the reentrancy happens through `setup`, `addModule` or `removeModule` at some point? It seems to me that `bindMethod` needs to be safe against reentrancy attacks regardless, how it got there isn't important.

Also, because it's `onlyModule` safety is actually guaranteed by the module, you could see `bindMethod` more as an internal method of a module, which doesn't really need guards against reentrancy because that's handled on a higher level at functions that are externally accessible.

Brechtpd:
I'm not sure if this adds a lot of security. Most times it's very easy:  make all `public`/`external` functions `nonReentrant`.

In the cases this isn't possible it seems to me that's mostly because we have some wallet <-> module interaction that is special, but pretty contained as they are fairly simple and don't do real external calls to unknown code themselves. (I tried to minimize this interaction in #788 but this was changed back afterwards). In these cases, we really need to think how this can be abused, but I don't know if reentrancy is the problem here. It really depends on exactly how the function works.

PR #1149, opened 2020-05-11
[protocol3] Updates for snarkjs 0.1.31

Brechtpd:
snarkjs was updated in #1109 but the update contains breaking changes.

### DIFF
--- a/packages/loopring_v3.js/src/babyjub.js
+++ b/packages/loopring_v3.js/src/babyjub.js
@@ -1,8 +1,10 @@
 // Taken and modified from
 // https://github.com/iden3/circomlib
 
-const bn128 = require("snarkjs").bn128;
-const bigInt = require("snarkjs").bigInt;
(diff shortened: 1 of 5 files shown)


kongliangzhong:
Does all tests past on your computer? some of them failed on my side.

Brechtpd:
Did you check out the latest circuits and removed old proving/verifying keys in the `keys` folder? I now added a change that when running `make` the cached keys are automatically removed.

2 tests are skipped, so seeing `2 pending` after running all tests is normal.

If the above isn't the problem please let me know which tests fail. I re-ran the tests and all tests pass for me.

kongliangzhong:
> Did you check out the latest circuits and removed old proving/verifying keys in the `keys` folder? I now added a change that when running `make` the cached keys are automatically removed.

Great, I guess I just forgot to remove the old keys

PR #1165, opened 2020-05-19
[hebao] optionally delete all guardians once inherited

dong77:
The following modifications are included:

1. After recovery, if the new owner used to be one of the guardians, it will be removed from the guardian list; if the wallet is locked, it will be unlocked forcefully.
1. After inheriting, the wallet will be unlocked forcefully, and all existing guardians are removed. otherwise, they may become obstacles for the new owner to execute transactions.

Something on my mind but I'm not sure whether to proceed:

**If wallet W1's inheritor (the new owner) is wallet W2, maybe we should copy the list of guardians of W2 to W1, and assign theW2's owner as W1's new owner. The benefit is that the our mobile app can simply support multiple wallets without complicating the design. Otherwise, the mobile app has to support a chain of ownership: W1 owned by W2 which is owned by an EOA, and the chain can be longer in theory.**

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
@@ -66,12 +66,12 @@ contract RecoveryModule is SecurityModule
         w.setOwner(newOwner);
 
         SecurityStore securityStore = controller.securityStore();
-        bool newOwnerIsGuardian = securityStore.isGuardian(wallet, newOwner);
-        if (newOwnerIsGuardian) {
+        bool removedAsGuardian = securityStore.isGuardianOrPendingAddition(wallet, newOwner);
+        if (removedAsGuardian) {
            securityStore.removeGuardian(wallet, newOwner, now);
         }
 
-        emit Recovered(wallet, oldOwner, newOwner, newOwnerIsGuardian);
+        emit Recovered(wallet, oldOwner, newOwner, removedAsGuardian);
     }
 
     function verifySigners(
--- a/packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
@@ -35,7 +35,8 @@ contract RecoveryModule is SecurityModule
     event Recovered(
         address indexed wallet,
         address indexed oldOwner,
-        address indexed newOwner
+        address indexed newOwner,
+        bool            removedAsGuardian
     );
 
     constructor(Controller _controller)
@@ -63,7 +64,14 @@ contract RecoveryModule is SecurityModule
         require(newOwner != address(0), "ZERO_ADDRESS");
 
         w.setOwner(newOwner);
-        emit Recovered(wallet, oldOwner, newOwner);
+
+        SecurityStore securityStore = controller.securityStore();
+        bool newOwnerIsGuardian = securityStore.isGuardian(wallet, newOwner);
+        if (newOwnerIsGuardian) {
+           securityStore.removeGuardian(wallet, newOwner, now);
+        }
+
+        emit Recovered(wallet, oldOwner, newOwner, newOwnerIsGuardian);
     }
 
     function verifySigners(
--- a/packages/hebao_v1/contracts/lib/OwnerManagable.sol
+++ b/packages/hebao_v1/contracts/lib/OwnerManagable.sol
@@ -51,6 +51,16 @@ contract OwnerManagable is Claimable, AddressSet
         return addressesInSet(MANAGER);
     }
 
+    /// @dev Gets the number of managers.
+    /// @return The numer of managers.
+    function numManagers()
+        public
+        view
+        returns (uint)
+    {
+        return numAddressesInSet(MANAGER);
+    }
+
     /// @dev Checks if an address is a manger.
     /// @param addr The address to check.
     /// @return True if the address is a manager, False otherwise.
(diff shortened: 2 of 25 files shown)


Brechtpd:
> 2. After inheriting, the wallet will be unlocked forcefully, and all existing guardians are removed. otherwise, they may become obstacles for the new owner to execute transactions.

If the guardians are against the inheritor they could have started the recovery procedure before the inheritor is even set as the wallet owner, so I'm not sure this is especially useful for that. There is also some risk in removing all guardians as only the inheritor has complete access to it and cannot recover it in any way if the inheritor keys are lost. Depending on the situation (guardians are social recovery or the user's own hardware wallet) this may or may not make sense. But the wallet owner should be able to remove the guardians afterwards if necessary. 

> **If wallet W1's inheritor (the new owner) is wallet W2, maybe we should copy the list of guardians of W2 to W1, and assign theW2's owner as W1's new owner. The benefit is that the our mobile app can simply support multiple wallets without complicating the design. Otherwise, the mobile app has to support a chain of ownership: W1 owned by W2 which is owned by an EOA, and the chain can be longer in theory.**

Seems like this potential long chain is inevitable. Guardians can be any address, and the owner can also be any address (so can be another wallet) so I'm not sure how inheritance here is a special case for this. But this doesn't mean the app needs to immediately support this of course.

I also think inheritance will mostly be used for wallets without any guardians. Otherwise the wallet would change owner sooner because the recovery process can be used.

dong77:
> > 1. After inheriting, the wallet will be unlocked forcefully, and all existing guardians are removed. otherwise, they may become obstacles for the new owner to execute transactions.
> 
> If the guardians are against the inheritor they could have started the recovery procedure before the inheritor is even set as the wallet owner, so I'm not sure this is especially useful for that. There is also some risk in removing all guardians as only the inheritor has complete access to it and cannot recover it in any way if the inheritor keys are lost. Depending on the situation (guardians are social recovery or the user's own hardware wallet) this may or may not make sense. But the wallet owner should be able to remove the guardians afterwards if necessary.
> 
> > **If wallet W1's inheritor (the new owner) is wallet W2, maybe we should copy the list of guardians of W2 to W1, and assign theW2's owner as W1's new owner. The benefit is that the our mobile app can simply support multiple wallets without complicating the design. Otherwise, the mobile app has to support a chain of ownership: W1 owned by W2 which is owned by an EOA, and the chain can be longer in theory.**
> 
> Seems like this potential long chain is inevitable. Guardians can be any address, and the owner can also be any address (so can be another wallet) so I'm not sure how inheritance here is a special case for this. But this doesn't mean the app needs to immediately support this of course.
> 
> I also think inheritance will mostly be used for wallets without any guardians. Otherwise the wallet would change owner sooner because the recovery process can be used.

I see what you mean , but let me digest and think about it. 

Do you agree with these changes then, Brecht:

- After recovery, if the new owner used to be one of the guardians, it will be removed from the guardian list; if the wallet is locked, it will be unlocked forcefully.
- After inheriting, the wallet will be unlocked forcefully.

Brechtpd:
> * After recovery, if the new owner used to be one of the guardians, it will be removed from the guardian list; if the wallet is locked, it will be unlocked forcefully.

This makes sense. Because a guardian could be removed It could be possible that the wallet can't be unlocked any more manually depending on the lock mechanism (but there is a time limit on a lock so it should be unlocked again automatically after some time in any case). Perhaps safe to unlock it to make sure,  this could depend on the lock mechanism.

> After inheriting, the wallet will be unlocked forcefully.

Seems unlikely it will be locked at this point (guardians would have to keep locking it for some time), but also makes sense I think.

> Or are you saying these are all not very necessary and the current implementation should allow ways for the new owner to take care of all the situations already.

I don't think they are very necessary, but I think both scenario's don't have a big impact on the defence mechanisms which need to work at all times. So if there is a problem in these situations, then I think it's also very likely the situation can happen in some other way.

dong77:
Brecht, the reason I think we should remove all guardians once a wallet is inherited is that the wallet's guardians are likely to act differently if the original owner passed away. This is quite normal in real life. If the original owner is alright and wants to give up control, he/she doesn't have to rely on the inheritance feature. Once this feature is used and takes effects, the owner is probably gone, and that will influence the existing guardian's decision making, especially when the total assets in the wallet are considerable.

Will you feel more comfortable with the optional removal of all guardians using another parameter?

Brechtpd:
Yes, I think that'd be nice. My main concern is to just do what people expect would happen so they don't get locked out of the wallet by accident.

kongliangzhong:
> If the guardians are against the inheritor they could have started the recovery procedure before the inheritor is even set as the wallet owner, so I'm not sure this is especially useful for that.

I think we can do this: if the inheritor is not 0x0, then it's signature is required to the recovery procedure.

dong77:
Then the inheritor has the power to say no to recovery, he can wait until he becomes the owner...

packages/hebao_v1/contracts/modules/security/RecoveryModule.sol
```diff
@@ -63,7 +64,14 @@ contract RecoveryModule is SecurityModule
         require(newOwner != address(0), "ZERO_ADDRESS");
 
         w.setOwner(newOwner);
-        emit Recovered(wallet, oldOwner, newOwner);
+
+        SecurityStore securityStore = controller.securityStore();
+        bool newOwnerIsGuardian = securityStore.isGuardian(wallet, newOwner);
```
kongliangzhong, line 69: newOwner isGuardian or isPendingAddition. should add a new method in securityStore: isGuardianOrPendingAddition.

PR #1187, opened 2020-05-23
[hebao][audit] Issue 7

dong77:
This PR has the following changes:

- The inheritor can optionally remove all existing guardians
- Once recovered or inherited, the wallet will be automatically unlocked
- Once recovered or inherited, if the wallet's new owner is one of the guardians or is pending to be added as a guardian, it will be removed from the guardian list immediately

We still allow adding an existing guardian as the inheritor.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
+++ b/packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
@@ -90,10 +90,12 @@ contract InheritanceModule is SecurityModule
         );
 
         SecurityStore securityStore = controller.securityStore();
+        securityStore.setInheritor(wallet, address(0));
+
         if (removeAllGuardians) {
             securityStore.removeAllGuardians(wallet);
         }
-        securityStore.setInheritor(wallet, address(0));
+        
         unlockWallet(wallet, true /*force*/);
         Wallet(wallet).setOwner(newOwner);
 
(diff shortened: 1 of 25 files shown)


packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
```diff
@@ -88,10 +90,20 @@ contract InheritanceModule is SecurityModule
...
         );
 
-        controller.securityStore().setInheritor(wallet, address(0));
+        SecurityStore securityStore = controller.securityStore();
+        bool removedAsGuardian = securityStore.isGuardianOrPendingAddition(wallet, newOwner);
+
+        if (removeAllGuardians) {
+            securityStore.removeAllGuardians(wallet);
+        } else if (removedAsGuardian) {
+            securityStore.removeGuardian(wallet, newOwner, now);
```
kongliangzhong, line 99: move        bool removedAsGuardian = securityStore.isGuardianOrPendingAddition(wallet, newOwner);
here

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -178,6 +190,17 @@ contract SecurityStore is DataStore
         w.guardians[idx - 1].validUntil = validUntil;
     }
 
+    function removeAllGuardians(address wallet)
+        public
+        onlyManager
+    {
+         Wallet storage w = wallets[wallet];
+         for (uint i = 0; i < w.guardians.length; i++) {
```
kongliangzhong, line 198: Is there any simpler way delete the whole mapping?

PR #1184, opened 2020-05-23
[hebao] refine transfer modules

kongliangzhong:
1. remove unnecessary functions in transfer modules
2. remove pending tx in QuotaTransfers module
 
tests not fixed yet.

PR #1189, opened 2020-05-23
[hebao] remove dapp modules and introduce DappTransfers

dong77:
This PR has the following changes:

1. Removed SubAccountModule
1. Removed most dapp modules
1. Improved existing transfer modules based on Kongliang's changes in https://github.com/Loopring/protocols/pull/1184.
1. Introduced a generic module `DappTransfers` to interact with dapps. Whitelisted apps will not consume user daily quota. This is a different design from Argent.
1. In TransferModule, always set allowance to 0 before setting it to another value.
1. In various place, when transfer/approve is called by modules using `transactCall`, we always verify the return value is true.


packages/hebao_v1/contracts/thirdparty/BytesUtil.sol
```diff
@@ -282,6 +283,18 @@ library BytesUtil {
         return tempBytes;
     }
 
+    // @dev This method was added by Daniel Wang.
+    function toBool(bytes memory _bytes, uint _start) internal pure returns (bool) {
```
Brechtpd, line 287: This seems to be used to decode the return value. This code likely works, but decoding the return value is ABI encoded so most of the time decoding it as harder. But there's a built in function that can be used for this:
```
abi.decode(data, (bool));
```

This is an easy way to decode whatever a function returns.

packages/hebao_v1/contracts/modules/transfers/DappTransfers.sol
```diff
@@ -0,0 +1,133 @@
...
+        returns (bool)
+    {
+        require (
+            method == this.transferToken.selector ||
+            method == this.approveToken.selector ||
+            method == this.callContract.selector ||
+            method == this.approveThenCallContract.selector,
+            "INVALID_METHOD"
+        );
+        return GuardianUtils.requireMajority(
```
Brechtpd, line 126: Does this need majority? This needs to be just owner no? Otherwise there is no difference with `ApprovedTransfers`?

packages/hebao_v1/contracts/modules/transfers/DappTransfers.sol
```diff
@@ -0,0 +1,133 @@
...
+pragma experimental ABIEncoderV2;
+
+import "../../lib/AddressSet.sol";
+import "../../lib/ERC20.sol";
+
+import "./TransferModule.sol";
+
+
+/// @title DappTransfers
+contract DappTransfers is TransferModule
```
Brechtpd, line 27: If `executeTransfers` is removed (I'm still not convinced! ;) then only being able to call a single function (+ optional approve) in a meta tx is probably a serious limitation.

dong77: If we truly want to support atomic batch calls to an external dapp, we can write a custom module to do that, it's probably easier and safer.

packages/hebao_v1/contracts/modules/transfers/DappTransfers.sol
dong77: You are right, this should be just the owner.

packages/hebao_v1/contracts/modules/transfers/DappTransfers.sol
```diff
@@ -0,0 +1,128 @@
...
+        address            token,
+        address            to,
+        uint               amount,
+        bytes     calldata logdata
+        )
+        external
+        nonReentrant
+        onlyWhitelistedDapp(to)
+        onlyWhenWalletUnlocked(wallet)
+        onlyFromMetaTx
```
kongliangzhong, line 53: onlyFromMetaTxOrWalletOwner(wallet)

packages/hebao_v1/contracts/thirdparty/ens/BaseENSManager.sol
```diff
@@ -60,7 +60,10 @@ contract BaseENSManager is IENSManager, OwnerManagable, ENSConsumer {
      * @param _rootName The root name (e.g. argentx.eth).
      * @param _rootNode The node of the root name (e.g. namehash(argentx.eth)).
      */
-    constructor(string memory _rootName, bytes32 _rootNode, address _ensRegistry, address _ensResolver) ENSConsumer(_ensRegistry) public {
+    constructor(string memory _rootName, bytes32 _rootNode, address _ensRegistry, address _ensResolver)
+        ENSConsumer(_ensRegistry)
+        OwnerManagable()
```
kongliangzhong, line 65: // will invoke automatically

PR #1193, opened 2020-05-23
[hebao][audit] issue 14

kongliangzhong:


packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
```diff
@@ -74,6 +74,7 @@ contract InheritanceModule is SecurityModule
         )
         external
         nonReentrant
+        onlyHaveEnoughGuardians(wallet)
```
Brechtpd, line 77: Isn't this a bad idea? Effectively the wallet is disabled if the owner is lost and there is only a single guardian.

I think the scenario of an owner (mobile wallet) + a single guardian (hardware wallet) is a very reasonable combination.

dong77:
In the mobile app we will strongly recommend and keep reminding users to add at least 2 guardians otherwise the app will disable certain features. The number of guardians required being greater than 1 is very important to create a network effect. 

The other reason is that we will offer users the Loopring Guardian service. What if this guardian is hacked (there will be a hot wallet to support signing)? We don’t want to see all wallets which Loopring Guardian as the only guardian get hacked.

Brechtpd:
> In the mobile app we will strongly recommend and keep reminding users to add at least 2 guardians otherwise the app will disable certain features. The number of guardians required being greater than 1 is very important to create a network effect.

For sure. But what is the benefit of enforcing this in the smart contract and potentially creating dangerous unexpected situations for users. There's a big difference between what the app recommends (and rightly so) and disabling something completely onchain.

Let's say the wallet has 2 guardians. The owner recovers his wallet and now 1 of the guardians is the new owner. In this new situation the wallet has now only 1 guardian and inheritance is disabled. How does this help? I don't really see the link between the number of guardians and the inheritance feature. They seem to be completely unrelated to each other and I think inheritance should work even without any guardians.

dong77:
Inheritance should not be disabled in this case. But we should disable recovery if there is only one guardian. Especially when the guardian is our official guardian operated by a hot wallet.

Brechtpd:
Alright, that makes sense in most cases. 

Maybe there's also a slight drawback of enabling the first guardian immediately but all other guardians are added with a delay. If the owner wants to add 2 friends as a guardian, and he wants to make it so both need to sign for recovery, with the current code 1 of those guardians will be added immediately, having full majority until the second one is enabled without the >= 2 guardians check. This likely wasn't what the owner expected/wants. Probably makes more sense to add both with a delay so they are enabled at the same time. Requiring 2 guardians is a workaround to this problem.

packages/hebao_v1/contracts/modules/security/LockModule.sol
```diff
@@ -66,6 +66,7 @@ contract LockModule is SecurityModule
         // onlyWhenWalletUnlocked(wallet)
         onlyFromMetaTxOr(guardian)
         onlyWalletGuardian(wallet, guardian)
+        onlyHaveEnoughGuardians(wallet)
```
dong77, line 69: recovery needs 2 guardian

packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
dong77: This is an incorrect modification, Kongliang should change the recovery module, not this inheritance module.

PR #1200, opened 2020-05-25
[hebao] special handling of erc20 transfers

dong77:
Add special handling of ERC20 transfers to support old tokens.

### DIFF
--- a/packages/hebao_v1/contracts/base/MetaTxModule.sol
+++ b/packages/hebao_v1/contracts/base/MetaTxModule.sol
@@ -388,13 +388,8 @@ abstract contract MetaTxModule is BaseModule
         if (gasSettings.token == address(0)) {
             transactCall(wallet, feeRecipient, gasCost, "");
         } else {
-            bytes memory txData = abi.encodeWithSelector(
-                ERC20(0).transfer.selector,
-                feeRecipient,
-                gasCost
-            );
             require(
-                abi.decode(transactCall(wallet, gasSettings.token, 0, txData), (bool)),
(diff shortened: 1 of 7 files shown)


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -388,13 +388,8 @@ abstract contract MetaTxModule is BaseModule
...
             transactCall(wallet, feeRecipient, gasCost, "");
         } else {
-            bytes memory txData = abi.encodeWithSelector(
-                ERC20(0).transfer.selector,
-                feeRecipient,
-                gasCost
-            );
             require(
-                abi.decode(transactCall(wallet, gasSettings.token, 0, txData), (bool)),
+                transactCallTokenTransfer(wallet, gasSettings.token, feeRecipient, gasCost),
```
Brechtpd, line 392: I forgot this is why we previously didn't check the return value here.

At this point the relayer will have spent the gas anyway, so even if the transfer actually fails for old tokens the cost has already happened for him. So I don't think it's super important to check this here.

But we can still do it quite easily without a special `transactCallTokenTransfer` like this (copied from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/SafeERC20.sol to be safe):
```
if (returndata.length > 0) { // Return data is optional
    require(abi.decode(returndata, (bool)), "SafeERC20: ERC20 operation did not succeed");
}
```

if `success == false` `BaseWallet.transact` will revert so we don't need to have access to it here.

dong77:
Great, Brecht, can you improve the PR using your approach. I won’t have time until Thursday, too late for Hebao’s app dev team.

PR #1201, opened 2020-05-25
[hebao] improve datastore security

dong77:
Only allow a wallet's module to change DataStore state.

### DIFF
--- a/packages/hebao_v1/contracts/base/DataStore.sol
+++ b/packages/hebao_v1/contracts/base/DataStore.sol
@@ -16,7 +16,7 @@
 */
 pragma solidity ^0.6.6;
 
-import "../lib/OwnerManagable.sol";
+import "../iface/Wallet.sol";
 
 
 /// @title DataStore
@@ -27,7 +27,11 @@ import "../lib/OwnerManagable.sol";
 ///
 /// The design of this contract is inspired by Argent's contract codebase:
 /// https://github.com/argentlabs/argent-contracts
-contract DataStore is OwnerManagable
+contract DataStore
 {
-    constructor() public OwnerManagable() {}
+  modifier onlyWalletModule(address wallet)
+    {
+        require(Wallet(wallet).hasModule(msg.sender), "UNAUTHORIZED");
+        _;
+    }
 }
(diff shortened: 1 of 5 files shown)


packages/hebao_v1/contracts/base/DataStore.sol
```diff
@@ -27,7 +27,11 @@ import "../lib/OwnerManagable.sol";
 ///
 /// The design of this contract is inspired by Argent's contract codebase:
 /// https://github.com/argentlabs/argent-contracts
-contract DataStore is OwnerManagable
+contract DataStore
 {
-    constructor() public OwnerManagable() {}
+  modifier onlyWalletModule(address wallet)
+    {
+        require(Wallet(wallet).hasModule(msg.sender), "UNAUTHORIZED");
```
Brechtpd, line 34: I think this may be exploitable like this:
- Deploy a malicious contract at the address a wallet will be created at with create2. Set all data storage to what the attacker wants (as the implementation can just always return true for `hasModule`)  
- Call self destruct on the contract so the wallet can be redeployed again at the same address using create2
- All settings set by the malicious wallet implementation will still be active for the actual wallet

Another change, though not exploitable at this point I think, is that any contract will be able to call functions on storage contracts just by implementing the `hasModule` interface.

dong77: "Deploy a malicious contract at the address a wallet will be created at with create2." is this even possible? I thought the code must be the same to end up the new contract in the same address.

dong77: "Another change is that any contract will be able to call functions on storage contracts just by implementing the hasModule interface. Perhaps if some of these would throw events it could spam a bit, but I don't think it will ever be able to do something really bad." -- this is indeed a good point.

Brechtpd: Oh right, my mistake. It also depends on the contract that deploys the contract which will also be different.

Brechtpd: > It means a DataStore can be modified by an arbitrage wallet, but that wallet will not happen to be one of our wallets, right? Our wallets won't be affected?

Yeah I think so, because the wallet is always used as the identifier and a wallet can only change the storage for its own wallet, so I don't think what will ever be a problem.

Issue #1168, opened 2020-05-19
[loopring4] allow permission-less account registration 


dong77:
If Alice would like to send some tokens to Bob on layer-2, if Bob does not have an Loopring account, Alice should be able to create Bob’s account by paying the fees. Bob’s initial EdDSA pubkey must be 0 for security considerations. Alice then cAn do the offchain transfers. 

Bob will have to reset the account key for the first time before he can get the control.

Brechtpd:
This is a good idea.

In 3.5 this is kind of already possible by using an Agent that would allow this. But in 3.5 only the account owner can authorize agents. So Bob would now have to authorize the Agent himself before the account can be created by that Agent, which doesn't make sense in this case. I thought about allowing exchanges to whitelist agents for all users, and I think this shows that this is indeed useful. 

Building this into the exchange contract is of course also possible, but I think it's also a good fit for Agents because this is the reason it was created (do things for a some user). The changes necessary for this are:
- Allow the Exchange owner or the Loopring protocol (owner) to whitelist agents
- Implement this behaviour in an Agent

Brechtpd:
On the other hand, this is so easy we may as well do this directly in the exchange contract.

dong77:
I would vote for the native support.

dong77:
https://github.com/Loopring/protocols/pull/1228

PR #1223, opened 2020-05-29
[Protocol3.5] [Draft] Permissionless account creation (not tested yet)

dong77:
Addressing https://github.com/Loopring/protocols/issues/1168
Not tested.

### DIFF
--- a/packages/loopring_v3/contracts/impl/ExchangeV3.sol
+++ b/packages/loopring_v3/contracts/impl/ExchangeV3.sol
@@ -233,13 +233,13 @@ contract ExchangeV3 is IExchangeV3
         }   
 
         return updateAccountAndDepositInternal(
-                owner,
-                pubKeyX,
-                pubKeyY,
-                address(0),
-                0,
-                permission
-            );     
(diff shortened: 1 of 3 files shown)


packages/loopring_v3/contracts/impl/ExchangeV3.sol
```diff
@@ -218,21 +218,28 @@ contract ExchangeV3 is IExchangeV3
...
             bool   isAccountNew,
             bool   isAccountUpdated
         )
     {
+        try state.getAccountID(owner) {
+            // Account exist, perform update
+            require(isAgent(owner, msg.sender), "UNAUTHORIZED");
+        } catch {
+             // Account not exist perform creation
+            require(owner == msg.sender || pubKeyX == 0 && pubKeyY == 0, "PUBKEY_MUST_BE_ZERO");
```
Brechtpd, line 232: Use `pubKeyX == 1` here for example. `pubKeyX == 0` is detected in the circuits as an unregistered account and nothing can be transferred to an account like that.

Issue #1218, opened 2020-05-27
[Protocol 4.0] Balance the Merkle tree


dong77:
> I think I proposed something similar before.

The account id is incremental by 1, this end up with a off-balance Merkel tree. The left tree is always *heavier* than that right one, as seen in the diagram below (in the tree on the left).

![WechatIMG2](https://user-images.githubusercontent.com/1710087/83002350-8b611280-a03f-11ea-926b-ec49a40aa4aa.jpeg)

As we increase the height of the tree, we hope the update to the Merkel tree can be parallelized using multiple machines. This off-balance property of the current Merkel tree makes some shards has no update to do and a few shards have all the work.

Can we place account nodes differently so the tree is more balanced, like the tree on the right?

dong77:
The proposed balancing approach can also be seen as a new way to generate the next account ID. Current we use `accountId += 1`, but we may now use two variables, one is the `accountIndex` to track the index, then the accounted (location the Merkle tree as:

```
accountIndex += 1;
accountId = bitReverse(accountIndex)
```

where bitReverse is a function to reverse every bit in the integer, for example:

1000 -> 0001,
1100 -> 0011,
1010 -> 0101

Brechtpd:
It is possible, but I just wonder what the problem is that this solves. It seems like this optimization depends on either the Merkle tree being largely empty or on a bit of luck that the updates that need to happen are spread out. What is the bottleneck? The hash calculations? IO?

Batch updates in a SMT are already very parallelizable for any leafs that need updating by calculating the new hashes per level (see https://youtu.be/cF6sG47pcws?t=1641). Doing the updates like this is very parallelizable, in all cases. There are only `O(tree_height)` sequential steps, but all the hash calculations in each step can be fully parallelized. So in theory, if each hash calculation is done on a different CPU, doing all the Merkle tree updates only takes `time_to_compute_single_hash * tree_height`, which seems to be the minimum that can be achieved.

But maybe the bottleneck is IO? I think the data can be stored however makes sense, so maybe leafs aren't stored in the order of their addresses if that helps?

dong77:
Current performance bottleneck is the update to the Merkel tree (we are maintaining an in-memory Merkel tree in one server). To enable more than 200 trades per second. We need to partition the Merkel tree into multiple shards and perform updates to sub-trees in parallel in multiple servers, then consolidate sub tree roots into the final tree root by a single server. 

[...]

PR #1224, opened 2020-05-29
A demostration of the new way of calculating account ids

dong77:
https://github.com/Loopring/protocols/issues/1218

### DIFF
--- a/packages/loopring_v3/contracts/impl/libexchange/ExchangeAccounts.sol
+++ b/packages/loopring_v3/contracts/impl/libexchange/ExchangeAccounts.sol
@@ -122,7 +122,7 @@ library ExchangeAccounts
         require(S.accounts.length < ExchangeData.MAX_NUM_ACCOUNTS(), "ACCOUNTS_FULL");
         require(S.ownerToAccountId[owner] == 0, "ACCOUNT_EXISTS");
 
-        accountID = uint24(S.accounts.length);
+        accountID = calculateAccountId(uint24(S.accounts.length));
         ExchangeData.Account memory account = ExchangeData.Account(
             owner,
             pubKeyX,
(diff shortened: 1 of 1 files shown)


packages/loopring_v3/contracts/impl/libexchange/ExchangeAccounts.sol
```diff
@@ -140,6 +140,18 @@ library ExchangeAccounts
         );
     }
 
+    function calculateAccountId(uint24 accountIdx)
+        public
+        pure
+        returns (uint24 accountID)
+    {
+        uint24 idx = accountIdx;
+        for (uint i = 0; i < 192; i++) {
```
Brechtpd, line 149: 192?

Brechtpd:
Another possible way to do this is to keep the nicely incrementing accountIDs, but just read the bits of the value flipped when using it as an address. So instead of starting at the LSB, we start at the MSB. Though perhaps that would be more confusing.

packages/loopring_v3/contracts/impl/libexchange/ExchangeAccounts.sol
dong77: my bad.

packages/loopring_v3/contracts/impl/libexchange/ExchangeAccounts.sol
```diff
@@ -140,6 +140,18 @@ library ExchangeAccounts
         );
     }
 
+    function calculateAccountId(uint24 accountIdx)
```
dong77, line 143: @Brecht, do you think it's a good idea to introduce a concept of TreeSharding

interface TreeSharding
{
     function accountIDFromAccountIndex(uint24 accountIndex) returns (uint24 accountID);
}

PR #1226, opened 2020-05-29
allow providing gasLimitSendToken

dong77:
@Brechtpd we listed TRB, whose transfer gas is very high. This makes me think maybe it's more flexible to allow the operator to specify what's the token transfer gas limit.

The other approach is to allow the operator to set the protocol level `gasLimitSendToken` variable.

### DIFF
--- a/packages/loopring_v3/DESIGN.md
+++ b/packages/loopring_v3/DESIGN.md
@@ -313,7 +313,7 @@ We store the deposit information on-chain so users can withdraw from these depos
 
 ## Withdrawing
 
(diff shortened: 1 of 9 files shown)


Brechtpd:
Interesting, even for just a single ERC20 transfer.

I noticed this problem as well for 4.0 with the potential insurance contract (which could make withdrawals also very expensive). Another option I think would be to just allow setting the value on the Deposit contract, so we don't even bother with it on the protocol level and the exchange contract just forwards all gas to the deposit contract for the withdrawal function (where it can fail depending on whatever limit the deposit contract wants to impose).

dong77:
Oh yes, if it is possible on the Deposit contract, it will be even better.
Forget about my PR, feel free to make the change.

PR #1272, opened 2020-06-09
[hebao] allow wallet owner remove its modules

kongliangzhong:


dong77:
only upgrade module can remove existing modules. Users are not allowed to remove modules, too dangerous.

PR #1286, opened 2020-06-11
[hebao]gas usage optimization

kongliangzhong:
About 50k gas saved when creating a wallet with all modules (9 modules) bound.


kongliangzhong:
About 250K gas saved now.

packages/hebao_v1/contracts/lib/AddressSet.sol
```diff
@@ -23,30 +23,23 @@ contract AddressSet
 {
     struct Set
     {
-        address[] addresses;
+        mapping (uint => address) addresses;
         mapping (address => uint) positions;
         uint count;
```
Brechtpd, line 28: Does it make more sense to leave `addresses` as an array and remove `count`? Seems like `addresses` is used as an array.

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -48,6 +59,17 @@ contract GuardianModule is SecurityModule
...
 
+    function bindableMethods()
+        public
+        pure
+        override
+        returns (bytes4[] memory methods)
+    {
+        methods = new bytes4[](2);
+        methods[0] = this.getLock.selector;
+        methods[1] = this.isLocked.selector;
```
Brechtpd, line 70: Maybe these not so useful binded methods can also be removed as they are expensive to set. I don't really see a big need to be able to do `wallet.isLocked` or `wallet.getLock`, may as well just call the methods on the module/storage for that. It's a non-standard interface (unlike `isValidSignature`) that 3rd party contracts that interact directly with the wallet won't depend on.

packages/hebao_v1/contracts/lib/AddressSet.sol
kongliangzhong: It will save about 5K gas for each add operation

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
kongliangzhong: Ok, will do

packages/hebao_v1/contracts/lib/AddressSet.sol
Brechtpd: I think it saves 5K gas for each push/pop because the array length doesn't need to be updated, but then you need to update `count` manually which also costs 5K gas. So I think both methods will use the exact same amount of gas because they essentially do the same things (add element to map/array + update `length`/`count`), but with an array it's just a bit cleaner I think. But if you think the current method using a map uses less gas than great, I did not do any measurements.

PR #1298, opened 2020-06-13
[hebao] demo how to use GSN to support meta tx in Hebao

dong77:


Brechtpd:
I guess we could also use the same Agent design as in protocol 3.5 so we also don't have to build in a specific meta tx type in hebao.

PR #1322, opened 2020-06-23
[Hebao_v1] fix lock/unlock meta-tx fee security issue

dong77:
I'm trying to fix a security issue just identified in the current hebao smart contracts.

## The issue
A guardian can use a meta-transaction to call the `lock` function with a very high gas price to deplete a wallet up to the wallet's daily quota.  If the official guardian's key is leaked from the hot wallet, the sole official guardian can steal all wallets' assets up to their corresponding daily quota.

## Analysis
The `lock` and `unlock` function can be called by any guardian without the knowledge of the wallet's owner. The meta-tx fee is deducted from the wallet's address, not the guardian's address. This should be seen as a security bug.

The reason is that in our executeMetaTx function, we didn't distinguish the logical message sender from the wallet in question. The meta-tx fee should only be deducted from the logical message sender. We should also extract a `reimburseFee` method and make it usable to modules and allow modules to call this method when necessary.

Going forward, I really want to use a better meta-transaction abstraction, something like GSN's codebase.

## About this PR
I'm not 100% sure about the fix, a careful review is demanded.

### DIFF
--- a/packages/hebao_v1/contracts/base/MetaTxModule.sol
+++ b/packages/hebao_v1/contracts/base/MetaTxModule.sol
@@ -189,12 +189,12 @@ abstract contract MetaTxModule is BaseModule
         );
         require(gasSettings.limit > 0, "INVALID_GAS_LIMIT");
 
-        address wallet = extractWalletAddress(data);
+        (address msgSender, address wallet) = extractWalletAddresses(data);
         bytes32 metaTxHash = EIP712.hashPacked(
             DOMAIN_SEPARATOR,
             hash(
                 MetaTransaction(
-                    wallet,
+                    msgSender,
                     address(this),
                     msg.value,
                     data,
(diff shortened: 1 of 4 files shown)


packages/hebao_v1/contracts/base/MetaTxModule.sol
```diff
@@ -233,7 +233,7 @@ abstract contract MetaTxModule is BaseModule
         emit MetaTxExecuted(msg.sender, wallet, nonce, metaTxHash, gasUsed, success, returnData);
 
         if (gasSettings.price != 0) {
-            reimburseGasFee(wallet, gasSettings, gasUsed);
+            reimburseGasFee(msgSender, gasSettings, gasUsed);
```
kongliangzhong, line 236: guardian有可能是一个EOA地址，所以这个地方会失败。

Brechtpd:
> A guardian can use a meta-transaction to call the lock function with a very high gas price to deplete a wallet up to the wallet's daily quota. If the official guardian's key is leaked from the hot wallet, the sole official guardian can steal all wallets' assets up to their corresponding daily quota.

It was actually made to specifically allow guardians to use meta txs. But I can see how a use case like with a centralized service this is deemed dangerous.

Possible easy fix: why don't we just only disallow things that only need a single guardian to be called by a meta tx? This way the guardian needs to call the function on the smart contract himself (which he can do whichever way he wants). For things that need multiple guardians normally (like recovery) we can still allow meta tx's. Basically only allow meta tx to be used when either the wallet owner needs to sign or when a majority of guardians are needed. I think this is only a problem for `lock`/`unlock` (not sure), so just change `onlyFromMetaTxOr(guardian)` to `onlyFrom(guardian)` for these functions.

PR #1323, opened 2020-06-23
[Hebao_v1] fix lock/unlock meta-tx fee security issue (approach#1)

dong77:
- Fix a lock/unlock meta-transaction fee related security issue
- Make `reimburseGasFee` internal in MetaTxModule

### DIFF
--- a/packages/hebao_v1/contracts/base/MetaTxModule.sol
+++ b/packages/hebao_v1/contracts/base/MetaTxModule.sol
@@ -189,12 +189,12 @@ abstract contract MetaTxModule is BaseModule
         );
         require(gasSettings.limit > 0, "INVALID_GAS_LIMIT");
 
-        address wallet = extractWalletAddress(data);
+        (address msgSender, address wallet) = extractWalletAddresses(data);
         bytes32 metaTxHash = EIP712.hashPacked(
             DOMAIN_SEPARATOR,
             hash(
                 MetaTransaction(
-                    wallet,
+                    msgSender,
                     address(this),
                     msg.value,
                     data,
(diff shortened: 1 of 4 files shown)


dong77:
Screen Shot 2020-06-23 at 22 13 04

MetaTx的签名方式需要改一点。

Brechtpd:
Hmm I understand the idea, but I think this actually makes things worse because `senderWallet` doesn't seem to need to authenticate the meta tx. The signatures are checked against the signers as expected on the `targetWallet`. So it seems any wallet can be used for reimbursement with the current code. Perhaps I'm looking over where this is authorized somehow.

But I don't this solves the main problem even `senderWallet` would authorize the meta tx somehow. You would still have the same authorization problem for reimbursement. When can `senderWallet` authorize the reimbursement?

Conceptually I like the idea of a guardian being able to use a meta tx on his own wallet to call any other function it needs. It's only because of the current design that we can't allow direct calls from Loopring wallets to Loopring modules, other smart wallets/EOAs don't have this limitation because they don't need this special case for our modules. Making an exception for this for calls to `executeMetaTx` seems like a workaround, but on the other hand not allowing calls on the modules to begin with is the actual workaround. 

But, I also think this check can be removed. `TransferModule` already contains a check for this: https://github.com/Loopring/protocols/blob/4ba3ac878da95a58ef05b1cc48625a1d3f757b2c/packages/hebao_v1/contracts/modules/transfers/TransferModule.sol#L124

I think that's the only place that users can do arbitrary calls, so I feel like the check can be removed in `BaseWallet`. This would even allow smart wallet guardians to call `lock`/`unlock` directly and no workarounds are needed to begin with.

dong77:
I see the point. Then lets drop the current PRs that attempt to fix this potential security issue and I'll start redesigning the meta transaction module.

PR #1328, opened 2020-06-24
[Hebao_1] redesign meta-transaction modules

dong77:
This is a preliminary PR to show you how I'd like to re-design the meta-transaction modules.

Here are some key points:

- There is one single **MetaTxRelayerModule** that is responsible for relaying all meta-transactions to different meta-tx modules. Meta-tx modules are no longer relayers, and  MetaTxRelayerModule itself is not a meta-tx module.
- Meta-tx modules are not aware of whether a function invocation is from a meta-transaction or a normal transaction. The logical message sender is abstracted by using `msgSender()`, not `msg.sender`. `msgSender()` of all meta-transactions can only be smart wallet addresses, not their guardians or owners (at least for now in this PR). I assume all wallet owners do not have Ether, and even if they do, they are not authorized to transact directly. Wallet owners can sign on behalf of their wallets using EIP1271. So now we allow wallets as `msgSender()` to perform most actions. **createWallet might be an exception**
- All signatures as input to function calls are now fully EIP712 compatible.
- move *addModule* from BaseModule to MetaTxModule, it seems even in the audited version, this function is not used at all.

TODO
- [x] WalletFactoryModule
- [x] better meta-tx nonce mgmt
- [x] better signed request nonce mgmt
- [x] better controller design


packages/hebao_v1/contracts/modules2/core/SignedRequest.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    function hash(Request memory request)
+        public
+        returns (bytes32)
+    {
+        return keccak256(
+            abi.encode(
+                REQUEST_TYPEHASH,
+                keccak256(abi.encode(request.signers)),
+                keccak256(abi.encode(request.signatures)),
```
Brechtpd, line 49: I don't think this can be a part of the data that is signed by all the signers, otherwise we end up with the same problem we had before where the signers array needs to be signed by all signers beforehand. And the signatures also being part of it makes it impossible I guess.

I don't think this is needed and they can just be removed here.

Brechtpd:
I like it, definitely an improvement!

So if I understand everything correctly, the meta tx stuff is basically still a module just for the fee reimbursement, otherwise it's pretty much the same as standard gsn. The trusted relayer is the meta tx module right? So if a user adds a module with meta tx support they basically trust the meta tx module. But to be able to create meta txs themselves they need to add the module to their wallet. Maybe there's some room for simplification here where all the meta tx code is fully pulled out of the main smart wallet contracts with users just authorizing the relayer contract somehow? (which would make sense because we're using the gsn design which allows just that).

We also have to watch out where `msgSender` is used currently. We have to make sure that `msgSender` with the trusted relayer is only used for wallets that have indeed authorized that trusted relayer somehow. Not a problem for functions that do transactions on the wallet or modify the stores because these can only be done by modules registered on the wallet, which is probably all of them, but maybe there are also some edge cases where that isn't the case.

I was also wrong yesterday, there is no problem with meta txs in meta txs because they don't pass through the wallet `transact` function, so no limitations here on how module functions can be called. But in any case, this design is better. :)

packages/hebao_v1/contracts/modules2/core/SignedRequest.sol
dong77: removed

dong77:
> the meta tx stuff is basically still a module just for the fee reimbursement, otherwise it's pretty much the same as standard gsn.

You are right. We need to wrap the meta-tx forwarder (renamed from meta-tx relayer) into a module so that it can reimburse fees, and also because the msgSender() is provided solely by this forwarder which also implies a lot of trusts.

dong77:
> Maybe there's some room for simplification here where all the meta tx code is fully pulled out of the main smart wallet contracts with users just authorizing the relayer contract somehow?

I like this idea, Let me think about how this can be done.

dong77:
> We also have to watch out where msgSender is used currently. 

I strongly agree with this. The code need thorough testing and a lot of reviews :)

PR #1420, opened 2020-07-17
[hebao] Added onlyWalletModule to HashStore/NonceStore functions

Brechtpd:
Adds `onlyWalletModule` in the store functions that need it.

I don't think it's necessary to keep the hashes separate for different wallets in HashStore as long as the hash already has the wallet address mixed in, but I don't think it hurts to do it just in case.

### DIFF
--- a/packages/hebao_v1/contracts/stores/NonceStore.sol
+++ b/packages/hebao_v1/contracts/stores/NonceStore.sol
@@ -31,6 +31,7 @@ contract NonceStore is DataStore
 
     function verifyAndUpdate(address wallet, uint nonce)
         public
+        onlyWalletModule(wallet)
     {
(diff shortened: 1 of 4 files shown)


packages/hebao_v1/contracts/stores/NonceStore.sol
```diff
@@ -31,6 +31,7 @@ contract NonceStore is DataStore
 
     function verifyAndUpdate(address wallet, uint nonce)
         public
+        onlyWalletModule(wallet)
```
dong77, line 34: do we have the same issue in other Stores?

Brechtpd: I think the other stores are fine (and they are still pretty similar to the previous versions I think).

PR #1421, opened 2020-07-18
[hebao_1.1] try to fix txAwareHash related issues.

dong77:


packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -72,10 +73,19 @@ contract ForwarderModule is BaseModule
...
-        // check that txAwareHash is indeed valid.
-        bytes memory data_ = (txAwareHash == 0) ? data : bytes("");
+        bytes memory data_ = data;
+
+        if (txAwareHash != 0) {
+            // If a non-zero txAwareHash is provided, we do not verify signature against
+            // the `data` field. The actual function call in the real transaction will have to
+            // check that txAwareHash is indeed valid.
+            require(nonce == 0, "NONCE_MUST_BE_0");
+            controller.hashStore().verifyAndUpdate(txAwareHash);
```
Brechtpd, line 83: We have to watch out here that the txAwareHash hash value doesn't get marked as used if the meta tx fails. Otherwise the relayer can still provide invalid data (but without getting a reimbursement), and the txAwareHash hash value that could be used with valid data would become unusable before it's actually used correctly.

So I think we either:
- allow the meta tx call to fail (current behaviour), then we need to do this in the meta tx call or only call `verifyAndUpdate` if the meta tx call is successful
- we don't allow the meta tx call to fail so that the complete Ethereum tx is reverted

Brechtpd: Doing `verifyAndUpdate` in the meta tx call doesn't work because it currently also acts as replay protection on the meta tx level if it's used, as you're probably aware as you moved it. :)

Brechtpd:
Everything looks good, but I think `txAwareHash` should only be marked as used when the meta tx call is successful (see comments above).

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dong77: Let me think about it:)

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -63,19 +65,21 @@ contract ForwarderModule is BaseModule
...
         public
-        view
     {
-        require(
-            (to == from) ||
-            (to == controller.walletFactory()) ||
-            (to != address(this) && Wallet(from).hasModule(to)),
-            "INVALID_DESTINATION"
-        );
+        require(to != address(this), "INVALID_DESTINATION");
```
Brechtpd, line 69: Oh no wait, we need to disallow calls on **any** wallet address and stores here! This is still a module with special privileges on those contracts! Hmmm... I guess this is a drawback of having this as a module.

dong77:
I'm done with this PR and it's ready to be reviewed.

@kongliangzhong there is a test failure :)

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dong77: oh.....

Brechtpd: Changing the check to `isModuleRegistered` instead of the wallet specific module check works I think. `callContract` can be used to call non-module contract functions (but not module functions), and meta txs can only be used to call module functions. Makes some sense. :)

dong77: what if 

```
        require(
            to != address(this) &&
            controller.moduleRegistry().isModuleRegistered(to),
            "INVALID_DESTINATION"
        );
```

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -66,16 +68,17 @@ contract ForwarderModule is BaseModule
         view
     {
         require(
-            (to == from) ||
-            (to == controller.walletFactory()) ||
-            (to != address(this) && Wallet(from).hasModule(to)),
+            to == controller.walletFactory() || // 'from' can be the wallet to create (not its owner),
```
dong77, line 71: @kongliangzhong 通过ForwarederModule发送MetaTx来生成钱包（这样可以从wallet扣费），钱包通过MetaTx调用自己的方法，以及调用自己钱包（比如转账），和其他钱包的Module（比如Recovery)这几种情况都没测试到。

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dong77: @Brechtpd please review this part again. this is very important !!!!

dong77:
change this to draft so @kongliangzhong will add more tests.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -65,17 +67,21 @@ contract ForwarderModule is BaseModule
...
         view
     {
+        // Since this contract is a module, we need to prevent wallet from interacting with
+        // Stores via this module. Therefore, we must carefully check the 'to' address as follows,
+        // so no Store can be used as 'to'.
         require(
-            (to == from) ||
-            (to == controller.walletFactory()) ||
-            (to != address(this) && Wallet(from).hasModule(to)),
+            to == controller.walletFactory() || // 'from' can be the wallet to create (not its owner),
```
Brechtpd, line 74: > to == controller.walletFactory() || // 'from' can be the wallet to create (not its owner),

I guess this won't be possible because `verifySignature` is called on the wallet before the wallet is created in the meta tx call.

kongliangzhong: Yes, I got the INVALID_SIGNATURE error now

dong77:
@kongliangzhong 3 failing tests:

```console
 88 passing (7m)
  3 failing

  1) Contract: TransferModule - approvedTransfer
       approveToken
         owner should be able to approve an address using daily quota:
     AssertionError: incorrect available quota
      at /Users/d/Projects/loopring/protocols/packages/hebao_v1/test/testQuotaTransfers.ts:531:5
      at Generator.next (<anonymous>)
      at fulfilled (transpiled/test/testQuotaTransfers.js:5:58)
      at processTicksAndRejections (internal/process/task_queues.js:89:5)

  2) Contract: TransferModule - approvedTransfer
       TransferToken (meta tx)
         owner should be able to transfer tokens using daily quota (meta tx):
     AssertionError: incorrect available quota
      at /Users/d/Projects/loopring/protocols/packages/hebao_v1/test/testQuotaTransfers.ts:143:5
      at Generator.next (<anonymous>)
      at fulfilled (transpiled/test/testQuotaTransfers.js:5:58)
      at processTicksAndRejections (internal/process/task_queues.js:89:5)

  3) Contract: TransferModule - approvedTransfer
       approveToken (meta tx)
         owner should be able to approve an address using daily quota (meta tx):
     AssertionError: incorrect available quota
      at /Users/d/Projects/loopring/protocols/packages/hebao_v1/test/testQuotaTransfers.ts:531:5
      at Generator.next (<anonymous>)
      at fulfilled (transpiled/test/testQuotaTransfers.js:5:58)
      at processTicksAndRejections (internal/process/task_queues.js:89:5)

```

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -48,11 +50,11 @@ contract ForwarderModule is BaseModule
     }
 
     bytes32 constant public META_TX_TYPEHASH = keccak256(
-        "MetaTx(address from,address to,uint256 nonce,address gasToken,uint256 gasPrice,uint256 gasLimit,bytes32 txAwareHash,bytes data)"
+        "MetaTx(address from,address to,uint256 nonce,address gasToken,uint256 gasPrice,uint256 gasLimit,bytes32 txAwareHash,bytes32 dataHash)"
```
Brechtpd, line 53: Why this change? When signing with EIP712 `data` won't be visible any more. `bytes` should work fine.

kongliangzhong: because the data was converted to bytes32 by keccak256(data_), If not change do this change, the signature will fail.

kongliangzhong: Or I can change the  keccak256(data) back to data.

Brechtpd: Hmmm... `keccak256(data)` is correct for calculating the hash because the `bytes` field is hashed internally by EIP712 to a bytes32 value. Not sure why it wouldn't work. Will try it.

Brechtpd: Pushed a commit that changes it to `bytes`, seemingly successful. Please revert if you come across any issues with it for some reason.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -65,17 +67,21 @@ contract ForwarderModule is BaseModule
...
+        // so no Store can be used as 'to'.
         require(
-            (to == from) ||
-            (to == controller.walletFactory()) ||
-            (to != address(this) && Wallet(from).hasModule(to)),
+            to == controller.walletFactory() || // 'from' can be the wallet to create (not its owner),
+                                                // or an existing wallet,
+                                                // or an EOA iff gasPrice == 0
+            controller.moduleRegistry().isModuleRegistered(to) && (to != address(this)) ||
+            controller.walletRegistry().isWalletRegistered(from) && (to == from),
```
Brechtpd, line 78: >  controller.walletRegistry().isWalletRegistered(from) && (to == from)

We cannot allow this, otherwise the wallet owner can call all critical functions on the wallet (transact, setOwner,...). Should be removed and as far as I can tell that should be fine. The only thing we probably want to do is to allow the wallet owner to call `addModule` on the base wallet so he can't end up with no modules and no wait to add any.

dong77: I got this fixed.

dong77:
@Brechtpd do you want a day or two to review the Hebao code before I tell solidified to start the next audit?

Brechtpd:
> @Brechtpd do you want a day or two to review the Hebao code before I tell solidified to start the next audit?

I don't think that will be necessary. An extra couple of days probably doesn't make that much difference, and any issues that are likely to show up at this point are probably pretty contained. So if all tests pass (and the tests test the different scenario's we need so we're sure the design works for the functionality we need) I would go ahead.

PR #1423, opened 2020-07-19
[Loopring 3.6] Multiple improvements and renames

dong77:
@Brecht, I'll not add more changes to this PR. Please review it :)


packages/loopring_v3/contracts/impl/libtransactions/DepositTransaction.sol
```diff
@@ -54,12 +54,14 @@ library DepositTransaction
...
+
+        // Also note the oritinal deposit.amount can be zero!
+        if (amount > 0) {
+            require(deposit.amount >= amount, "INVALID_AMOUNT");
             feeETH = uint(deposit.fee).mul(amount) / deposit.amount;
             deposit.fee = uint64(uint(deposit.fee).sub(feeETH));
+            deposit.amount = uint96(uint(deposit.amount).sub(amount));
         }
-        // Consume what was deposited
-        deposit.amount = uint96(uint(deposit.amount).sub(amount));
```
Brechtpd, line 62: `deposit.amount > 0` check is needed in the if because there is a division by `deposit.amount`;

packages/loopring_v3/contracts/impl/libtransactions/NewAccountTransaction.sol
```diff
@@ -16,45 +14,46 @@ import "../../lib/SignatureUtil.sol";
...
+        // offset += 3;
         //uint16 feeTokenID = data.toUint16(offset);
-        offset += 2;
+        // offset += 2;
         //uint fee = uint(data.toUint16(offset)).decodeFloat(16);
-        offset += 2;
-        uint24 newAccountID = data.toUint24(offset);
+        // offset += 2;
+
+        uint offset = 8;
```
Brechtpd, line 48: I prefer not to do it like this because the `8` is now a special value that comes from nowhere.

packages/loopring_v3/contracts/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -16,6 +16,13 @@ import "../libexchange/ExchangeWithdrawals.sol";
 
 /// @title WithdrawTransaction
 /// @author Brecht Devos - <brecht@loopring.org>
+/// @dev The following 4 types of withdrawals are supported:
+///      - withdrawType = 0: offchain withdrawals with EdDSA signatures
+///      - withdrawType = 1: offchain withdrawals with ECDSA signatures or onchain appprovals
+///      - withdrawType = 2: onchain valid forced withdrawals (owner and accountID matche)
+///      - withdrawType = 3: onchain invalid forced withdrawals (owner and accountID mismatch)
+/// TODO(brecht):
+///      - withdrawType = 4: offchain operator-initiated withdrawals for protocol fees or in shutdown mode.
```
Brechtpd, line 25: The different types are for the different behaviours in the circuit. `withdrawType = 4` like this would not behave differently than types 2/3 in the circuit (so we'd have to have 4/5, but they would do exactly the same thing).

The types are not related to how the onchain verfication is done. Type `1` withdrawals can be authorized onchain with an ECDSA signature or an approved hash, but the actual behaviour is exactly the same, it's just authorized differently, so they use the same type. The operator just needs to provide the correct authentication.

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
```diff
@@ -52,7 +58,7 @@ library ExchangeGenesis
         S.blocks.push(ExchangeData.BlockInfo(bytes32(0)));
 
         // Get the protocol fees for this exchange
-        S.protocolFeeData.syncedAt = uint32(0);
+        S.protocolFeeData.syncedAt = uint32(now);
```
Brechtpd: `0` is probably better because we only set the protocol fees to the max value here, so it would take some time before the actual values are used when blocks are being submitted immediately after the exchange is created. Probably very unlikely because it will take some time before blocks are submitted after the exchange is created, but I think it makes sense to check the actual values the first time a block is submitted to get the recent values.

packages/loopring_v3/contracts/impl/libexchange/ExchangeTokens.sol
```diff
@@ -68,7 +68,7 @@ library ExchangeTokens
         address tokenAddress,
         uint    amountToBurn
         )
-        public
+        internal
```
Brechtpd, line 71: Any reason to make this internal? This inlines the function in the places it is being used, to it will add extra bloat there, increasing contract sizes. Gas use for this function call isn't a concern, so best to only use `internal` for critical functions where it is.

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -71,36 +68,33 @@ contract BasicDepositContract is IDepositContract, ReentrancyGuard
...
+                    uint(amount)
                 );
             }
         }
 
         uint exchangeBalanceBefore = exchangeBalance[token];
         // Keep track how many tokens are deposited in the exchange
-        uint exchangeBalanceAfter = exchangeBalanceBefore.add(amount);
-        // Make sure the total max amount per token in the exchange is capped
-        require(exchangeBalanceAfter <= MAX_TOTAL_TOKEN_BALANCE, "MAX_AMOUNT_REACHED");
```
Brechtpd, line 102: Would like to keep this check to keep circuits easier.

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
```diff
@@ -29,12 +35,12 @@ library ExchangeGenesis
         bytes32 _domainSeperator
         )
         external
+        onlyWhenUninitialized(S)
```
Brechtpd, line 38: Already checked in ExchangeV3.initialize, so probably a bit overkill to do it here as well? (maybe even a bit overboard to have a dedicated `ExchangeGenesis` library contract just for this initialization function...)

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
dong77: Brecht, is it correct to say that each token's balance on layer-2 for each user should be <= MAX_TOTAL_TOKEN_BALANCE, and each deposit/withdrawal's total amount should also be <= MAX_TOTAL_TOKEN_BALANCE, but the total balance in a DepositContract can be > MAX_TOTAL_TOKEN_BALANCE?

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
dong77: Agreed, reverted.

packages/loopring_v3/contracts/impl/libexchange/ExchangeTokens.sol
dong77: Reverted

packages/loopring_v3/contracts/impl/libtransactions/DepositTransaction.sol
dong77: updated, please take another look.

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
Brechtpd: See also other comments on different places about this. But if the index can decrease I may have to implement more checks anyway so the onchain check may not be that useful any more...

packages/loopring_v3/contracts/impl/libtransactions/WithdrawTransaction.sol
dong77: I'lll keep it as-is

PR #1429, opened 2020-07-20
[hebao] Fix test failure

dong77:


dong77:
Screen Shot 2020-07-20 at 18 52 30

I’m not sure why this change matters. Without this change, some tests will fail.

dong77:
Screen Shot 2020-07-20 at 18 52 30

I’m not sure why this change matters. Without this change, some tests will fail.

PR #1433, opened 2020-07-20
[Loopring_3.6] add notifyDepositProcessed and notifyWithdrawalProcessed to IDepositContract

dong77:


Brechtpd:
- To make this work we would also have to mark add an extra parameter in `withdraw`. Now the deposit contract has no idea if the withdrawal was already notified or not so can't really act on it.
- The balance is still inside the deposit contract and is still fully available to be used by the deposit contract if the withdrawal fails. There isn't really any difference with if the balance would still be available inside the rollup.
- It's still not fully accurate because the index used in the circuits may already be older than the onchain index. So I'd say it makes more sense to send the balance index with the withdrawal and keep track of it so we can be sure it's fully accurate (this is needed so support negative index updates, we need to be 100% in that case for scenario's like that to not make the exchange insolvent).
- (If we don't care about negative index updates) Is it actually bad that the amount would still be part of the index calculations? It's still part of the deposit contract balance so it is also still used to earn interest. The deposit contract doesn't need to care. The owner of the balance just doesn't get interest on his part because we don't track it now, but it's still fair for all people inside the rollup. If people withdraw and pay a fee for it I assume they actually want to get their funds out as soon as possible, so doesn't seem to matter much to me any way, they wouldn't get interest for maybe an 1 hour for that balance.

Brechtpd:
There's 2 options here when we have negative index updates (so we need to keep track of balances perfectly):
- Withdrawals always lock in the amount withdrawn at the latest index (which could still differ from the index in the Merkle tree, so we have to recalculate onchain). Afterwards it is not subjected to any index changes any more. The amount may have been stored in a defi app so the operator may always need to eat the cost to remove the balance from the defi app for each withdrawal, even if the withdrawal itself is not successful (so in theory this could be very expensive for the operator if no balance is available). To some extend this is the approach in this PR.
- Withdrawals are not locked to the amount at withdrawal time and are still subjected to index changes. The balance can stay in the defi app until the user manually withdraws the amount himself and the amount withdrawn is recalculated against the latest index value. This will be cheaper when the withdrawal does not happen instantaneously because there is nothing to do on the deposit contract.

Both options are pretty similar, so whatever is the easiest/cleanest to implement.

dong77:
> There's 2 options here when we have negative index updates (so we need to keep track of balances perfectly):
> 
> * Withdrawals always lock in the amount withdrawn at the latest index (which could still differ from the index in the Merkle tree, so we have to recalculate onchain). Afterwards it is not subjected to any index changes any more. The amount may have been stored in a defi app so the operator may always need to eat the cost to remove the balance from the defi app for each withdrawal, even if the withdrawal itself is not successful (so in theory this could be very expensive for the operator if no balance is available). To some extend this is the approach in this PR.
> * Withdrawals are not locked to the amount at withdrawal time and are still subjected to index changes. The balance can stay in the defi app until the user manually withdraws the amount himself and the amount withdrawn is recalculated against the latest index value. This will be cheaper when the withdrawal does not happen instantaneously because there is nothing to do on the deposit contract.
> 
> Both options are pretty similar, so whatever is the easiest/cleanest to implement.

As a user, I expect after my withdrawal is done (even when I didn't get the tokens yet), my tokens are no longer part of the deposit fund pool at and thus are free from DeFi risks. But as the latest index may differ from the latest one onchain, we cannot guarantee a user will be able to withdraw from the contract 100% at all.

I propose that we do not re-calculate the amount that can be claimed by approved withdrawals. Instead, when an unclaimed withdrawal is done, we get money back from DeFi, accumulate these funds together, then each unclaimed withdrawal will share this sub-fund proportionally.

A demo is like this:



contract SomeDepositContract
{
    function notifyUntransferredWithdrawal(
        address token,
        uint    amount)
    {
        unclaimed[token].outstanding += amount;
        unclamied[token].balance += (withdrawFromDeFi(token, 0.4*amount) + 0.6*amount);
    }
    function depositTransfer(
        address from,
        address token,
        uint    amount,
        bytes   calldata auxiliaryData
        )
        external
        payable
        returns (uint actualAmount, uint tokenIndex)
        {
           depositToDeFi(token, 0.4*amount);
           // keep 60% in this contract
        }
    function withdrawTransfer(
        address to,
        address token,
        uint    amount,
        bytes   calldata auxiliaryData,
        bool    unclaimed
        )
        external
        payable {
            uint payamount;
            if (unclamied) {
                payamount = amount * unclaimed[token].balance / unclaimed[token].outstanding);
                
                unclaimed[token].outstanding -= amount;
                unclaimed[token].balance -= payamount;
            } else {
                payamount = (withdrawFromDeFi(token, 0.4*amount) + 0.6*amount);
            }
            transeferOut(token, payamount);
        }
        。。。

}

dong77:
> * To make this work we would also have to mark add an extra parameter in `withdraw`. Now the deposit contract has no idea if the withdrawal was already notified or not so can't really act on it.
> * The balance is still inside the deposit contract and is still fully available to be used by the deposit contract if the withdrawal fails. There isn't really any difference with if the balance would still be available inside the rollup.
> * It's still not fully accurate because the index used in the circuits may already be older than the onchain index. So I'd say it makes more sense to send the balance index with the withdrawal and keep track of it so we can be sure it's fully accurate (this is needed so support negative index updates, we need to be 100% in that case for scenario's like that to not make the exchange insolvent).
> * (If we don't care about negative index updates) Is it actually bad that the amount would still be part of the index calculations? It's still part of the deposit contract balance so it is also still used to earn interest. The deposit contract doesn't need to care. The owner of the balance just doesn't get interest on his part because we don't track it now, but it's still fair for all people inside the rollup. If people withdraw and pay a fee for it I assume they actually want to get their funds out as soon as possible, so doesn't seem to matter much to me any way, they wouldn't get interest for maybe an 1 hour for that balance.

I updated the interface

PR #1444, opened 2020-07-21
[Hebao_1.1] fix forwarderModule bug

dong77:
Previously:

Someone can use meta transaction to call `addGuardian` with `nonce=0` and `txAwareHash !=0`, then change the `data` to call the same function with different parameters.

### DIFF
--- a/packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
+++ b/packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
@@ -28,6 +28,12 @@ abstract contract MetaTxAware
         trustedForwarder = _trustedForwarder;
     }
 
+    modifier txAwareHashNotAllowed()
+    {
+        require(txAwareHash() == 0, "INVALID_TX_AWARE_HASH");
+        _;
+    }
+
     /// @dev Return's the function's logicial message sender. This method should be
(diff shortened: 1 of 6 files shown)


packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
```diff
@@ -28,6 +28,12 @@ abstract contract MetaTxAware
         trustedForwarder = _trustedForwarder;
     }
 
+    modifier txAwareHashNotAllowed()
+    {
+        require(txAwareHash() == 0, "INVALID_TX_AWARE_HASH");
+        _;
+    }
+
```
letsgoustc, line 36: 加这个是不允许MetaTX被重复发送？

kongliangzhong: 是的。

Brechtpd:
I knew this was possible, but I don't see it as a security issue. It's the user that signs the wrong data. We can't really guard against the user signing the wrong things.

That's why we sign the function selector for the meta tx so it's the user that remains in control and the relayer can't mess things up agains the users. If we check this everywhere (quite error prone because easy to forget) I don't think we even have to sign the function selector like we do now.

Issue #1422, opened 2020-07-18
[Protocol 3.6] Propose to remove Dual Authoring in transfers


dong77:
The reason we added Dual Authoring (DA) is to support the following two use cases:

- Red Envelopes: Chinese people have this tradition to send red envelopes to friends. Usually, each envelope contains 0.1-10 USD. The sender will create a RE and share in WeChat (that's the only action he needs to do), people in the group will be able to click on the RE to claim his share, a random percentage of the total amount. The receiver will see the money received in his WeChat wallet immediately. If there is any money left in the RE, it will be returned to the sender after 24 hours.

DA will make sure the sender can create open transfer requests that don't specify the `to` fields. The receiver can fill in the `to` fields and signs the modified transfer request by not sharing the DA key with our relayer, making sure the relayer won't be able to become the receiver to steal the money.

- Payees scan the QR code of payers:

In China, the most popular payment apps are WeChat and Alipay. In many cases, vendors will scan consumer's QR code -- the consumer will simply show the QR code without specifying the amount and he doesn't care about the vendor's account. According to hoss, this payee-scan-payer scenario are most people where the payer doesn't have access to the internet.

## Challenges & Suggestion
The problem with the transfer-based red envelope solution is that if a transfer with a bigger nonce is included in a block, pending open transfers will fail. After a discussion in Shanghai, we realized we have done this red-envelop thing sometime ago, by using trades, not transfers. We believe it's more elegant as it doesn't use nonce at all. Here is how:

- The sender will simply create an ask order whose amountS is the total value for the RE and amountB is 0.
- The receive will create a bid order whose amountS is 0 and amountB is the amount he will get.
- The client will simply match these two orders and send the ring to the relayer. Noted that the receiver's order is fully filled, and the sender's order may be only partially filled.
- The relayer checks the account balance of the sender and if the balance is sufficient, include the trade in the next bock. The relayer cannot put the order in the orderbook.

If we implement DA for trades, we can make this process also entered trustless, but I feel it's an overkill. I don't think we want to pay the price of making the spot-trade circuit bigger simply because of such use cases. Remember REs are mostly used among friends, coworkers, and relatives, and the amount is small.

The true value of DA for trades is to enable trustless OTC (peer-to-peer) trades without trusting the relayer. But again, I don't think we want to complicate the trade circuits for that.

The payee-scan-payer can also be implemented with an open-trade. The payer create an order at price 0. and the payee takes the order by paying 0.

So here are my **suggestions**:

[...]

Issue #1279, opened 2020-06-10
[Loopring 3.* Bounty Idea] Payment receipts (challage and response game)


Brechtpd:
Receipts are given to a user by the operator that does a transfer. The receipt provides him some extra guarantees that the payment will actually go through and the transfer won't just be thrown away. If the transfer isn't included in a block the receipt can be used to still get the transferred amount directly from an insurance contract provided by the operator. 

This provides better finality, but because the operator can create an infinite number of receipts where he promises to include transfers in a block, but the insurance contract only has a limited amount of funds, it's certainly not a 100% guarantee the transfer will actually be done. But still pretty good.

This functionality is provided by zkSync and Starkware, and people are talking about this. If want to make an actual payment platform it's a good thing to have.

dong77:
How receipt is used/verified/guaranteed in the smart contract?

Brechtpd:
That is still an open question as I haven't really looked into this at all. :)

But I guess it could work similarly to light clients for blockchains. 
- For each block a hash of all transactions is saved onchain (either a flat list or in a Merkle tree format).
- The receipt the operator gives to the payer/payee is the block and position the transfer will be done in. 
- The operator submits the block. If the transfer didn't happen at the promised position in the block the payer/payee can submit a fraud proof on the insurance contract showing that a different transfer happened at that position. He can do that by posting all the DA again (or if it's in a Merkle tree format, a Merkle proof) and shows that the receipt contained a different transfer than was actually done.
- If indeed the operator lied, the receipt can be used to still transfer the funds to the payee using funds inside the insurance contract the operator created to provide instant economic finality.

So basically just by saving the data availability hash for each block onchain would already allow us to provide such functionality. We actually do that in 3.1 but don't do that any more in 3.5. 

... so I guess we can actually do this already in 3.1 if we want to! The data availability hash for each block can be queried with `getBlock`. Heh, I actually wasn't aware of this before typing this up haha. :)

But I guess there could be better ways to do it. One downside in 3.1 is that the transfer doesn't contain a nonce to make sure the transfer is unique, but that's not the biggest problem as the recipient of the receipt can just check if the receipt is actually "good" (not cheating by pointing to an old transfer or something like that) before accepting it.

dong77:
So if we append the hash of each block in a list in 3.6, it is still hard to accept a receipt as a challenge because the data (DA) for each block is not structured (queryable), but as call data, isn't it?

[...]

Issue #1209, opened 2020-05-26
[Loopring 4.0] Add `validUntil` to transfers


dong77:
If a layer-2 transfer is not processed before validUntil, the user now can decide to transfer on layer-1. Without this option, the user may end up with a duplicate payment.

dong77:
@Brechtpd  it seems this is not done yet.

dong77:
We also need to add 'validUntil to withdrawal requests.

Brechtpd:
It's there for transfers, but it's not available when the transfer isn't signed with EdDSA. Do you want it to be available everywhere? The only difference is that we also need to put the timestamp onchain which we currently don't do.

> We also need to add 'validUntil to withdrawal requests.

Okay, makes sense.

dong77:
As a user, I need to make sure that my offchain request has a way to become invalid at some point. If I signed the request with my ECDSA signature using an EOA, I can never invalidate my request without a 'validUntil' field. EdDSA signed requests can be invalidated by updating the EdDSA key, but hash-based requests need a way to remove those hashes which is also not possible as of now.

Therefore, maybe we should add `validUntil` to all offchain requests. But I don't think the timestamp will need to be put onchain, as it will be checked in the circuits and is not part of DA. Correct me if I'm wrong.

But if we have to put it onchain, we may want to use a bigger unit, such as 1 day. Then we can save maybe 1 byte per timestamp.

Brechtpd:
> As a user, I need to make sure that my offchain request has a way to become invalid at some point. If I signed the request with my ECDSA signature using an EOA, I can never invalidate my request without a 'validUntil' field. EdDSA signed requests can be invalidated by updating the EdDSA key, but hash-based requests need a way to remove those hashes which is also not possible as of now.

The "normal" way to invalidate a tx (like Ethereum transactions) is to create another tx with the same nonce so the other one can't be used any more. This also works for us for all transactions, even those approved with hashes onchain.

> But I don't think the timestamp will need to be put onchain, as it will be checked in the circuits and is not part of DA. Correct me if I'm wrong.

It doesn't need to be part of the DA because it's not needed to reconstruct the Merkle tree, but it does need to be sent onchain because we have to use it to calculate the transaction hash so we can check the signature/approved hash. So kind of boils down to the same thing in this case. Though if we don't put it into the DA, we will have to check the timestamp onchain because it will not have been checked in the circuits (only the data in the DA is validated directly by the circuit).

[...]

PR #1450, opened 2020-07-21
[Protocol 3.6] Protocol 3.6 improvements

Brechtpd:
- #1462 
- #1457 
- #1456 
- #1422
- #1209 
- #1279 (only things in the protocol)
- Removed index
- Updated to solidity 0.7.0
- Removed new account/transfer account transaction types

And some other small improvements.


dong77:
Brecht, when you add the new field to an Order, can you make sure it means the other order's id: otherOrderID & newField == newField, instead of checking orderOrderID == newField. Or do you think it's a good idea?

Brechtpd:
> Brecht, when you add the new field to an Order, can you make sure it means the other order's id: otherOrderID & newField == newField, instead of checking orderOrderID == newField. Or do you think it's a good idea?

Not sure I follow, why check the orderID? To lock an order to a single other order? But the owner of the order could still make whatever order he wants in that case because he can create as many orders with that orderID as he wants, so I don't think that could work. I guess we could check the other's order hash, but I'm not sure what you're trying to achieve here. :)

dong77:
My bad, wrong expression, I did mean the owner's address of the other order.

> > Brecht, when you add the new field to an Order, can you make sure it means the other order's id: otherOrderID & newField == newField, instead of checking orderOrderID == newField. Or do you think it's a good idea?
> 
> Not sure I follow, why check the orderID? To lock an order to a single other order? But the owner of the order could still make whatever order he wants in that case because he can create as many orders with that orderID as he wants, so I don't think that could work. I guess we could check the other's order hash, but I'm not sure what you're trying to achieve here. :)

My bad, wrong expression. Let me try again.
Assume the new field is called `addressMask`, now what does this means：it means the other order's owner address `order.owner` must satisfy: 
```
ower.owner & addressMask == addressMask
```

instead of：
```
ower.owner == addressMask
```

The benefit is that we can allow the order to be matched by one of N owners in a set, we just need to do:
```
addressMask = owner1 | owner2 | .. |ownerN
```

This is more flexible, but I'm not sure it will be used a lot.

Brechtpd:
Ah, some sort of [bloom filter](https://en.wikipedia.org/wiki/Bloom_filter). Yeah, could be cool and is still easy to implement. Not sure how "safe" it is when the set is quite large, but for a set of a single element it will be as safe as the equality check, so can't seem to hurt.

Brechtpd:
To do bitwise operations we need to break the value up in bits and then do the actual bit operation. So this will cost around 160 * 4 * 2 = 1280 constraints. So not cheap, but also not expensive.

`order.owner & addressMask == addressMask` needs to be `order.owner & addressMask == owner` I think. 

Because the set contains ORd addresses I think the addressMask will contain a lot of 1s even with a few addresses(~75% 1s for the result of ORing 2 addresses?), so I think it won't be that hard to brute force find addresses that will pass the test that were not in the original set if the set is > 1 or 2 (but it does have the nice property that only gets worse with the size of the set).

Brechtpd:
I did some testing with the address mask, and the false positives are simple too high even for extremely small sets. I used this python script to verify:
```
import random

numRounds = 100000
setSize = 4

mask = 0
set = []
for i in range(setSize):
    value = random.SystemRandom().randint(0, 2**160)
    mask |= value
    set.append(value)
print("mask: " + "{0:b}".format(mask))

numFalsePositives = 0
breakingPoint = 0
found = 0
for i in range(numRounds):
    address = random.SystemRandom().randint(0, 2**160)
    if address & mask == address:
        if not (address in set):
            numFalsePositives += 1

print("False positive rate: " + str(numFalsePositives*100/numRounds) + "%")
```

It takes a fraction of a second to find a false positive even with just 4 values in the set. And even a set of 1 has quite a lot of false positives (these are just addresses that have some of the bits not set to 1, which there are also many of).

Making it an actual bloom filter isn't trivial and will also cost a lot more constraints.

Anther option is to just have a fixed, limited set of addresses (e.g. max 16 addresses) and just hash those to create a mask, the operator than needs to show that the target address is part of this fixed list of addresses, but this also quite a bit of work to implement.

So would stick to the simple equality check unless a simple alternative is found which at least works and is 100% secure for the common case it just needs to target a single address. In any case the number of addresses in the set will be limited no matter the method (unless with great cost). A limited set of target addresses can also be made possible with order aliasing: the exact same order (with the same orderID) is simply signed multiple times with different taker addresses.

dong77:
> I did some testing with the address mask, and the false positives are simple too high even for extremely small sets. I used this python script to verify:
> 
> ```
> import random
> 
> numRounds = 100000
> setSize = 4
> 
> mask = 0
> set = []
> for i in range(setSize):
>     value = random.SystemRandom().randint(0, 2**160)
>     mask |= value
>     set.append(value)
> print("mask: " + "{0:b}".format(mask))
> 
> numFalsePositives = 0
> found = 0
> for i in range(numRounds):
>     address = random.SystemRandom().randint(0, 2**160)
>     if address & mask == address:
>         if not (address in set):
>             numFalsePositives += 1
> 
> print("False positive rate: " + str(numFalsePositives*100/numRounds) + "%")
> ```
> 
> It takes a fraction of a second to find a false positive even with just 4 values in the set. And even a set of 1 has quite a lot of false positives (these are just addresses that have some of the bits not set to 1, which there are also many of). The problem is that the mask just contains too many 1s even for very small sets. The chance a bit is 1 is ` 1 - (1/2)^n`, which quickly goes to 1.
> 
> Making it an actual bloom filter isn't trivial and will also cost a lot more constraints.
> 
> Anther option is to just have a fixed, limited set of addresses (e.g. max 16 addresses) and just hash those to create a hash of the set, the operator than needs to show that the target address is part of this fixed list of addresses, but this also quite a bit of work to implement and doesn't really add that much.
> 
> A limited set of target addresses can also be made possible with order aliasing: the exact same order (with the same orderID) is simply signed multiple times with different taker addresses. I think this is currently the best method.
> 
> So would stick to the simple equality check unless a simple alternative is found which at works for decent sized sets and is 100% secure for the common case it just needs to target a single address. In any case the number of addresses in the set will be limited no matter the method (unless with great cost).

So lets keep things simple by checking equality then.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
```diff
@@ -142,6 +128,23 @@ library OwnerChangeTransaction
             // If the account has a wallet, use it to recover the account
             require(accountTransfer.walletHash != 0, "ACCOUNT_HAS_NO_WALLET");
 
+            // Verify authorization from the new owner
```
dong77, line 131: I don't think we need to get authorization from the new owner, otherwise, the user experience will be bad. Even in hebao, we don't need a signature from the new owner.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
```diff
@@ -153,6 +156,14 @@ library OwnerChangeTransaction
                     )
                 )
             );
+            // Verify that the data was signed by the current owner
```
dong77, line 159: In the case of using stateless wallet, the current owner may have lost access, so no need to check authorization from the current owner.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
```diff
@@ -142,6 +128,23 @@ library OwnerChangeTransaction
             // If the account has a wallet, use it to recover the account
             require(accountTransfer.walletHash != 0, "ACCOUNT_HAS_NO_WALLET");
 
+            // Verify authorization from the new owner
+            if (auxData.signature.length > 0) {
+                require(
```
dong77, line 133: Across many files, we find this pattern of verifying either ECDSA signature or an approval hash. Maybe we can extract a common function?

packages/loopring_v3.js/src/constants.ts
```diff
@@ -3,19 +3,18 @@ import { FloatEncoding } from "./float";
...
 
 export class Constants {
-  static readonly BINARY_TREE_DEPTH_TRADING_HISTORY = 14;
-  static readonly BINARY_TREE_DEPTH_ACCOUNTS = 24;
+  static readonly BINARY_TREE_DEPTH_STORAGE = 14;
+  static readonly BINARY_TREE_DEPTH_ACCOUNTS = 32;
   static readonly BINARY_TREE_DEPTH_TOKENS = 12;
 
-  static readonly TX_DATA_AVAILABILITY_SIZE = 104;
+  static readonly TX_DATA_AVAILABILITY_SIZE = 106;
```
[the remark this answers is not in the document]
Brechtpd: Yes, no way around that, we have to hash the same amount of data for each transaction type. The zeros are compressed away where it makes sense (though the compression can certainly still be approved).

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -115,15 +116,15 @@ contract BasicDepositContract is IDepositContract, ReentrancyGuard, Claimable
         override
         payable
         onlyExchange
-        nonReentrant
+        //nonReentrant
```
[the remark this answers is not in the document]
Brechtpd: Never, because only the exchange can call these functions and the exchange is always nonReentrant. Basically the deposit contract is like an upgradeable library. Removing this saves ~3,000 gas per call into the deposit contract.

packages/loopring_v3/contracts/impl/agents/FastWithdrawalAgent.sol
```diff
@@ -100,34 +99,59 @@ contract FastWithdrawalAgent is ReentrancyGuard
...
+                        fastWithdrawal.validUntil
+                    )
+                )
+            );
+
+            // Check the signature
+            require(hash.verifySignature(fastWithdrawal.from, fastWithdrawal.signature), "INVALID_SIGNATURE");
+
+            // Check the time limit
+            require(block.timestamp <= fastWithdrawal.validUntil, "TX_EXPIRED");
```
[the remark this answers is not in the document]
Brechtpd: Yeah, the` validUntil` here isn't passed into `setWithdrawalRecipient`, and the user never signs any of the data used in that case so basically at this point any data can be used, of course as the risk of the liquidity provider.

packages/loopring_v3/contracts/impl/agents/OpenGSN2Agent.sol
```diff
@@ -1,169 +1,169 @@
...
-    {
-        (bool success, bytes memory returnData) =
-            exchange.call{value: msg.value}(msg.data);
-
-        if (!success) {
-            assembly { revert(add(returnData, 32), mload(returnData)) }
-        }
-    }
-}
+// import "@opengsn/gsn/contracts/BaseRelayRecipient.sol";
```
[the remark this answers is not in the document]
Brechtpd: Doesn't compile yet with 0.7.0, so uncommented until it gets upgraded I'm afraid...

packages/loopring_v3/contracts/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -170,6 +170,23 @@ library WithdrawTransaction
...
+        // Check if there is a withdrawal recipient
+        address recipient = S.withdrawalRecipient[withdrawal.owner][withdrawal.to][withdrawal.tokenID][withdrawal.amount][withdrawal.nonce];
+        if (recipient != address(0)) {
+            // Auxiliary data is not supported
+            require (auxData.auxiliaryData.length == 0, "AUXILIARY_DATA_NOT_ALLOWED");
+
+            // Set the new recipient address
+            withdrawal.to = recipient;
+            // Allow any amount of gas to be used on this withdrawal (which allows the transfer to be skipped)
+            withdrawal.minGas = 0;
```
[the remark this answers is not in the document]
Brechtpd: Otherwise your idea won't work to do cheap redirects without actual token transfers. If the user does a normal withdrawal then surely in most cases some minimum amount of gas will be specified. So if we still keep enforcing this gas limit then the operator can't choose to use the cheaper path of just updating `amountWithdrawable` multiple times for the liquidity provider and he would still be forced to do the (likely) more expensive full token transfer withdrawals.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
Brechtpd: I removed the requirement when the old owner can still sign, but kept this one because someone needs to sign something. Basically we have the same problem as with our smart wallet: who can sign meta txs for an account. The new owner is required to sign because he authorizes:
- The fee payment (using funds from the account being recovered, so only the old or the new owner can do that)
- Which data is used to recover the account

I guess we could let another account pay for the recovery, but I think that makes things a bit more complicated, and in any case someone needs to sign the transaction so why not the new owner and keep things simple?

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
Brechtpd: The wallet data would be signed before it is set on the account while the owner still has access to his keys. So if the owner wants to set the data on his account, he would also sign it and store the signature for it.

I added this for extra security when the user's EdDSA keys are lost or shared between different users. Currently it is allowed to update the EdDSA keys/wallet data for an account with just the EdDSA keys because it's significantly cheaper than needing an ECDSA signature. If the users's EdDSA keys are stolen all funds in his account are at risk, but when the wallet can also be updated he would also risk losing his account completely because the owner could be changed by setting some dummy wallet. By also requiring an ECDSA signature for the wallet data this is not possible. At least the user's account is safe from being overtaken completely, so the user's accountID is still secure as long as his Ethereum key is safe.

So not too important, but allows us to keep more common operations as cheap as possible while having some extra safe guards against theft of the EdDSA keys.

packages/loopring_v3/contracts/impl/libexchange/ExchangeWithdrawals.sol
```diff
@@ -126,23 +126,22 @@ library ExchangeWithdrawals
...
         external
     {
         uint16 tokenID = S.getTokenID(token);
-        ExchangeData.Deposit storage deposit = S.pendingDeposits[owner][tokenID][index];
+        ExchangeData.Deposit storage deposit = S.pendingDeposits[owner][tokenID];
         require(deposit.timestamp != 0, "DEPOSIT_NOT_WITHDRAWABLE_YET");
 
         // Check if the deposit has indeed exceeded the time limit
-        require(now >= deposit.timestamp + S.maxAgeDepositUntilWithdrawable, "DEPOSIT_NOT_WITHDRAWABLE_YET");
+        require(block.timestamp >= deposit.timestamp + S.maxAgeDepositUntilWithdrawable, "DEPOSIT_NOT_WITHDRAWABLE_YET");
```
dong77, line 138: should we skip check this if the exchange is in WithdrawalMode? something like:

```solidity
require(isWithdrawalMode ||
             block.timestamp >= deposit.timestamp + S.maxAgeDepositUntilWithdrawable, "DEPOSIT_NOT_WITHDRAWABLE_YET");
```

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
dong77: "The wallet data would be signed before it is set on the account while the owner still has access to his keys"
Unless the signature of the current owner is saved inside the Merkle tree and becomes part of DA, there is no guarantee that his/her signature can be found when the user's wallet is used.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
dong77: We can/should always assume that if the current owner configs his wallet to have guardians, then a majority of the guardian's approval does authorize the operator to charge the recovery fee out of this account. I don't think the new owner needs to sign. Most likely one of his guardians is another address of his.

"Which data is used to recover the account"
I think we can simply reset the wallet data after the recovery, let the new owner to config his wallet later.

dong77:
Please see - https://github.com/Loopring/protocols/issues/1523

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
Brechtpd: The same is true for the actual wallet data, which is also not part of the Merkle tree or the DA, but needs to be stored somewhere else by the user/platform (publicly or privately), so no difference there.

packages/loopring_v3/contracts/impl/libtransactions/AccountTransferTransaction.sol
Brechtpd: That's a good point, but they'd have approve the fee payment as well which they currently do not, but that makes sense.

PR #1465, opened 2020-07-22
[Loopring_3.6] Disable empty blocks

dong77:
I don't have a strong reason why empty blocks should be disabled, but I feel it's wrong to allow different blocks having the same root.

I guess the check may not be necessary if the relayer account's nonce must +1. But I just want to double check.


Brechtpd:
> I guess the check may not be necessary if the relayer account's nonce must +1. But I just want to double check.

Yeah this will never be the case, but I guess an extra check can't hurt.

Issue #1251, opened 2020-06-04
[Loopring 3.6] Add convinient function for user to withdraw more than one tokens


dong77:
`function withdrawFromApprovedWithdrawal(address owner, address[] tokens)`

This will enable users to fire one single transaction to get all his/her withdrawals back.

dong77:
@Brechtpd Do you think this requirement makes sense? We currently have `withdrawFromApprovedWithdrawals`, but maybe we should simply provide `withdrawFromApprovedWithdrawal`(without the `s).  We have `forceWithdraw`, not `forceWithdrawals` (with a `s`). So maybe we should always provide ABI that operates on one single token/user, as a principle

Brechtpd:
`withdrawFromApprovedWithdrawals` is useful because even if you implement batching externally it will make things more expensive because of the contract call overheads. And this method may be used to do withdrawals for lots of users at the same time, so efficiency is important. This can do withdrawals for multiple different users at once.

`forceWithdraw` only does a withdrawal request, and it needs to be called by the account owner himself, so they are very different operations. And `forceWithdraw` is only there for safety, it's not going to be used at all normally because it doesn't make sense when the operator does what he needs to do, so it doesn't really matter how efficient it is.

dong77:
> `withdrawFromApprovedWithdrawals` is useful because even if you implement batching externally it will make things more expensive because of the contract call overheads. And this method may be used to do withdrawals for lots of users at the same time, so efficiency is important. This can do withdrawals for multiple different users at once.
> 


> `forceWithdraw` only does a withdrawal request, and it needs to be called by the account owner himself, so they are very different operations. And `forceWithdraw` is only there for safety, it's not going to be used at all normally because it doesn't make sense when the operator does what he needs to do, so it doesn't really matter how efficient it is.

Agreed.

PR #1466, opened 2020-07-22
[Loopring_3.6] Enable forcing withdrawal of multiple tokens in one L1 tx

dong77:
https://github.com/Loopring/protocols/issues/1251


Brechtpd:
Because the fee is paid in ETH, the user will need to send the forced withdrawal fee * tokens.length, so needs some extra logic for that in `state.forceWithdraw`.

I'm probably repeating myself a lot haha, but I'm not expecting `forceWithdraw` to be used at all.

dong77:
OK，I'm closing this one.

Issue #1463, opened 2020-07-22
[Loopring 3.*] mass migration


dong77:
I fee this is a very critical feature.

https://ethresear.ch/t/mass-migration-to-prevent-user-lockin-in-rollup/7701
https://ethresear.ch/t/batch-deposits-for-op-zk-rollup-mixers-maci/6883

dong77:
Maybe we can support a special migration to move all existing users to a new Merkel tree whose history/token tree heights are the same, but the account tree height is bigger. So the existing Merkel tree will become the left-most sub-tree of the new Merkel tree. 

![Screen Shot 2020-07-23 at 16 41 28](https://user-images.githubusercontent.com/1710087/88267407-63222700-cd03-11ea-9a1a-5d06dad6c026.jpg)

We don't even need to provide an proof for it, just need to make sure when users want to provide a Merkel proof for the new tree, it can withdraw the fund. This can simply be done by making the height of the account tree a variable instead of a fix value in the smart contract.

dong77:
We'd decided to implement this feature in the smart contract (danielw), Brecht will change the accountID to use 32 bits instead of 24 bits so that we can support up to 4 billion users.

PR #1472, opened 2020-07-23
[Loopring_3.6] enable the account-sub-tree to grow from 12 to 16 layer

dong77:
This is the first part of the changes required to support this feature: https://github.com/Loopring/protocols/issues/1463#issuecomment-662889886

The change in circuit will be changing accountID from 24 bits to 32 bits. Then this rollup will initially support 4^12 users (16 million), and the capacity can be increased up to 4^16 users (4 billion)


packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
```diff
@@ -62,4 +73,41 @@ library ExchangeGenesis
...
+        bytes32 t0,
+        bytes32 t1,
+        bytes32 t2,
+        bytes32 t3
+        )
+        private
+        pure
+        returns (bytes32)
+    {
+        Poseidon.HashInputs5 memory inputs = Poseidon.HashInputs5(
```
dong77, line 104: @Brechtpd do we need to do `>> 3` trick here?

Brechtpd: No, the Poseidon hashes are limited by the field size automatically.

Brechtpd:
@dong77 I'm going to measure how much the number of constraints increase if we make the accounts tree 32bits so we have a better idea on that.

Brechtpd:
@dong77 Currently the increase/tx is 6885 constraints to go from a depth of 12 to a depth of 16, or an increase of just ~4.5%. It's likely this can be reduced further to ~3%. So I think we can just always use the full 32bit tree.

dong77:
> @dong77 Currently the increase/tx is 6885 constraints to go from a depth of 12 to a depth of 16, or an increase of just ~4.5%. It's likely this can be reduced further to ~3%. So I think we can just always use the full 32bit tree.

Greate, lets do it.

dong77:
@Brechtpd shall I close this PR or merge it to your active branch?

Brechtpd:
I think this PR can be closed. My branch is using a fixed 32bit tree, so no need for dynamic depth changes.

dong77:
> I think this PR can be closed. My branch is using a fixed 32bit tree, so no need for dynamic depth changes.

Brecht, my understanding is that we can save some money when we initially use a 24bit tree after launch, then gradually increase the height of the tree only when necessary. Are you worried about the extra trusted setups?

Brechtpd:
> Greate, lets do it.

Seems like I misunderstood this. :)

I'm bit worried about trusted setups, yes. But mainly just not trying to do things that aren't super easy and don't make too much of difference performance wise so we can spend this time optimizing things with the biggest impact and the lowest risk.

I'm not really concerned about proving costs. Previously in 3.1 and $200 ETH + 1 Gwei gas prices the prover costs were already only ~1/3 of the total costs (and this was even in the best case scenario onchain which we're currently very far away from). In 3.6 transactions do use around 2x the constraints, but of course gas prices have skyrocketed and will probably not decrease for the foreseeable future and will probably never be as cheap again. So onchain costs are x50 to x100 higher than what they were, which basically means proving costs are not important because the prover cost is only ~1-2% of the total cost now.

So I think time is best spent on optimizing the onchain costs. Compress the calldata as good as we can, don't do any unnecessary things when submitting blocks etc... Those are easier and less complex optimizations that will make a much bigger difference and are useful in all cases. An increase in number of constraints does impact the number of transactions we can do inside a single block of a certain number of constraints, so in that way it also impacts the gas costs a bit. But this feature would only save quite a small number of constraints so the difference is very minimal. And this can be mitigated easier by just making the blocks bigger/optimize the prover some more when/if we need it, which will make a much bigger difference in costs (or some specialized hardware provers may be available at that time which makes this even less of a concern).

dong77:
> > Greate, lets do it.
> 
> Seems like I misunderstood this. :)
> 
> I'm bit worried about trusted setups, yes. But mainly just not trying to do things that aren't super easy and don't make too much of difference performance wise so we can spend this time optimizing things with the biggest impact and the lowest risk.
> 
> I'm not really concerned about proving costs. Previously in 3.1 and $200 ETH + 1 Gwei gas prices the prover costs were already only ~1/3 of the total costs (and this was even in the best case scenario onchain which we're currently very far away from). In 3.6 transactions do use around 2x the constraints, but of course gas prices have skyrocketed and will probably not decrease for the foreseeable future and will probably never be as cheap again. So onchain costs are x50 to x100 higher than what they were, which basically means proving costs are not important because the prover cost is only ~1-2% of the total cost now.
> 
> So I think time is best spent on optimizing the onchain costs. Compress the calldata as good as we can, don't do any unnecessary things when submitting blocks etc... Those are easier and less complex optimizations that will make a much bigger difference and are useful in all cases. An increase in number of constraints does impact the number of transactions we can do inside a single block of a certain number of constraints, so in that way it also impacts the gas costs a bit. But this feature would only save quite a small number of constraints so the difference is very minimal. And this can be mitigated easier by just making the blocks bigger/optimize the prover some more when/if we need it, which will make a much bigger difference in costs (or some specialized hardware provers may be available at that time which makes this even less of a concern).

I'm convinced.

PR #1474, opened 2020-07-24
[Loopring_3.6] add LoopringV3_6_vs_V3_1.pdf

dong77:


Brechtpd:
Very good overview. Some minor things:

- Paralleled transfers: 3.1 does require a strict nonce + 1 for each operation that uses a nonce, transfers included. In 3.6 currently it will also not be possible to jump over slots. This is so we can still enforce transfers to be done sequentially as well by always writing over the same slot for the transfers, and this also decreases the amount of data that needs to be put onchain (just the slot and if the slot gets overwritten and thus the stored Id in the slot increases by 2^14). So if you have transfers with storageID = [X + 1 * 2^14, X + 2 * 2^14, X + 3 * 2^14], they will have to be done sequentially or not at all.
- Implicit account creation: 3.2 -> 3.6
- Deposit/Withdraw To any address: 3.1 can deposit to a different address.
- Receipts: 3.1 also exposes the public data hash for a block onchain so receipts could be implemented on 3.1 as well. But of course we haven't written any of the necessary code, but that code isn't part of the main protocol so not sure if it needs to be part of the list for explicit differences.

PR #1475, opened 2020-07-24
[Loopring_3.6] Simplify BasicDepositContract

dong77:
This contract can now handle burn-over-transfer tokens


packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -110,12 +113,17 @@ contract BasicDepositContract is IDepositContract, ReentrancyGuard
...
-        // Transfer the tokens from the contract to the recipient
-        transferOut(to, token, amount);
+        if (isETHInternal(token)) {
+            to.sendETHAndVerify(amount, gasleft());
+        } else {
+            // Just in case the balance is unexpectedly smaller
+            uint balance = ERC20(token).balanceOf(address(this));
+            uint _amount = balance < amount ? balance : amount;
+            token.safeTransferAndVerify(to, _amount);
+        }
```
Brechtpd, line 126: I think we can make this a bit more efficient by using a try/catch. Now we always check the balance which will increase gas costs ~2,500 gas or something like that for each withdrawal. We can just always do the transfer, and only when it fails do this additional logic to cap it to the max available.

Another option is to let it fail, and let the user withdraw with `withdrawFromApprovedWithdrawals`, but we modify the function to take an extra amountToWithdraw parameter so it can be set to whatever is still available in the deposit contract. This way we never have to worry about this case onchain, at least not in this simple use of deposit contracts.

If we do this check for ERC20 tokens, why not for ETH?

dong77:
```solidity
try token.safeTransferAndVerify(to, amount) {} catch {
                uint balance = ERC20(token).balanceOf(address(this));
                require(balance < amount, "UNEXPECTED_STATE");
                token.safeTransferAndVerify(to,balance);
            }
```
we get:
```
Try can only be used with external function calls and contract creation calls.
```

dong77: > If we do this check for ERC20 tokens, why not for ETH?
Because this contract can never send ETH out except by calling the withdrawal function.

dong77: > Another option is to let it fail, and let the user withdraw with withdrawFromApprovedWithdrawals, but we modify the function to take an extra amountToWithdraw parameter so it can be set to whatever is still available in the deposit contract. This way we never have to worry about this case onchain, at least not in this simple use of deposit contracts.

I would prefer to only allow the user to withdraw the full amount inside the `withdrawal` record, not allow them to specify the amount.

Brechtpd: > Try can only be used with external function calls and contract creation calls.

Ah yes sorry, in this case just use `safeTransfer` which returns a boolean.

> Because this contract can never send ETH out except by calling the withdrawal function.

Just like ERC20 tokens no? Where is there a difference between ETH/ERC20? I thought this was for unexpected losses on the the deposit contract.

> I would prefer to only allow the user to withdraw the full amount inside the withdrawal record, not allow them to specify the amount.

Okay, makes sense.

Brechtpd: > Because this contract can never send ETH out except by calling the withdrawal function.

> Just like ERC20 tokens no? Where is there a difference between ETH/ERC20? I thought this was for unexpected losses on the deposit contract.

Or do you mean with ETH nothing unexpected can happen like on a token contract. That would indeed be true.

dong77: please take another look

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -110,12 +120,21 @@ contract BasicDepositContract is IDepositContract, ReentrancyGuard
...
-        // Keep track how many tokens are deposited in the exchange
-        exchangeBalance[token] = exchangeBalance[token].sub(uint(amount));
-
-        // Transfer the tokens from the contract to the recipient
-        transferOut(to, token, amount);
+        actualAmount = amount;
+        if (isETHInternal(token)) {
+            to.sendETHAndVerify(actualAmount, gasleft());
+        } else {
+            try ERC20(token).transfer(to, actualAmount) {} catch {
```
Brechtpd, line 128: This may not work correctly for those "bad ERC20 tokens` that don't follow the standard correctly (like not throwing and only returning a boolean or things like that). I think it's safer to use `safeTransfer` here instead which will handle all those special cases.

dong77: safeTransfer is not working, not an external function :(

dong77: I figured it out.

Brechtpd: `safeTransfer` returns a boolean you can use, it does the try/catch internally.

packages/loopring_v3/contracts/impl/BasicDepositContract.sol
```diff
@@ -3,66 +3,77 @@
...
+        else { _; }
     }
 
+    event CheckBalance(
+        address indexed token,
+        bool            checkBalance
+    );
+
+    event Deposit   (address token, uint amountReceived);
+    event Withdrawal(address token, uint amountPaid);
```
Brechtpd, line 51: We would now throw deposit/withdrawal events here and in the exchange contract, seems unnecessary to throw them on both contracts.
The exchange contract is better fit for the withdrawal events because it can also throw an event when a withdrawal isn't successful (or we could log successful withdrawals in the deposit contract and unsuccessful withdrawals on the exchange contract... but that seems a bit confusing).

dong77: Agreed, removed.

PR #1479, opened 2020-07-25
[Loopring_3.6] extract IAgentRegistry

dong77:


packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
```diff
@@ -44,8 +45,10 @@ library ExchangeGenesis
         S.genesisMerkleRoot = _genesisMerkleRoot;
         S.DOMAIN_SEPARATOR = _domainSeperator;
 
-        ILoopringV3 loopring = ILoopringV3(_loopringAddress);
+        ILoopringV3 loopring = ILoopringV3(_loopring);
         S.loopring = loopring;
+
+        S.agentRegistry = IAgentRegistry(loopring.agentRegistryAddress());
```
Brechtpd, line 51: Saw this more of a per exchange kind of thing, but everything will probably happen inside a single exchange contract any way so probably doesn't matter. :)

packages/loopring_v3/test/testExchangeAgents.ts
```diff
@@ -1,286 +0,0 @@
...
-    ownerB = ctx.testContext.orderOwners[1];
-    ownerC = ctx.testContext.orderOwners[2];
-    ownerD = ctx.testContext.orderOwners[3];
-  });
-
-  after(async () => {
-    await ctx.stop();
-  });
-
-  describe("Agents", function() {
```
Brechtpd, line 104: These test will need some updating, but will still be needed. So will restore and update later if deleted now.

packages/loopring_v3/contracts/impl/AgentRegistry.sol
```diff
@@ -0,0 +1,116 @@
...
+    function registerInternal(
+        bytes32 key,
+        address agent,
+        bool    toRegister
+        )
+        private
+    {
+        require(agent != address(0), "ZERO_ADDRESS");
+        if (toRegister) {
+            addAddressToSet(key, agent, true);
```
Brechtpd, line 103: Do we need to maintain a list for these agents? This currently makes things quite a bit more expensive. Instead of a fixed 20,000 gas cost to register an address using just a map, the list will cost 80,000 for the first address, and 50,000 for the following addresses. So 30,000 gas to 60,000 gas more than just a simple map. At current gas prices an increase in costs in the order of $1.

Without maintaining the list the cost is still 20,000 gas higher for the first address and 5,000 gas higher for the following addresses compared to just a map.

dong77: Ok, I'll disable the list-query.

packages/loopring_v3/test/testExchangeAgents.ts
dong77: I restored the file but comment out the current tests.

packages/loopring_v3/contracts/impl/libexchange/ExchangeGenesis.sol
dong77: Makes a lot of sense. Code changed accordingly.

packages/loopring_v3/contracts/impl/AgentRegistry.sol
Brechtpd: Still up to 2x the cost just to be able to query how many agents someone has registered, but I guess pretty reasonable now.

dong77: Yes, but these queries are not frequent I guess.

PR #1480, opened 2020-07-25
[Loopring_3.6] Remove 'operator' from IExchangeV3 & introduce AccessManager

dong77:
While we are operating our Loopring.io exchange, we want to have 3 roles:

- the owner - the only address who can send some critical translations
- a token registration manager - who manages what token to be enabled by the exchange, this should be a cold wallet
- the operator's address.

Currently, our engineers will have access to the owner's private key to register tokens. Not good.

Now, we can manage access in another contract outside of the core exchange.

Warning: this will break most tests.


Brechtpd:
> Currently, our engineers will have access to the owner's private key to register tokens. Not good.

This isn't a limitation of the current system though, both the current system and the system in this PR are identical in functionality, the only difference is the efficiency that can be achieved. We just haven't bothered making a contract like that (though we have bothered with making other better owner contract with delays but we're not using those either...). You could set a single `SelectorBasedAccessManager` contract as both the operator and the owner in our live 3.1 protocol and it would work exactly the same way as in this PR. 

The main thing this PR changes is that it abstracts the concept of the operator away in the protocol, which at this point makes sense I think! The only reason it wasn't done like that was to make `submitBlocks` as efficient as possible because it could be an EOA address (that also make it easier to parse the DA data), which isn't possible with the owner because the owner is security oriented so would need to be a contract (and indeed would also need to contain extra logic to make e.g. the token registration permissionless using some process). But at this point we also want to make the operator more flexible so a contract there is inevitable as well. 

It could make sense to save a bit of gas on block submissions to allow function calls with compressed data directly on `SelectorBasedAccessManager` (like possible in `submitBlocksCompressed` in `Operator`), this way we don't necessarily have to go through yet another contract (with an overhead for block submissions I think maybe 1-2% overhead per contract hop) if we just submit blocks from an EOA (and the open operator contract that would allow other parties to submit blocks can be done from a different address). But maybe that's not the best place to do it to because upgrading `SelectorBasedAccessManager` will be a lengthy/more difficult process because it also manages all the owner related functions, operator contracts can be switched out immediately without any problem. So let's just keep it as it is now, I will do some measurements later.

packages/loopring_v3/contracts/impl/owners/LoopringIOExchangeOwner.sol
```diff
@@ -0,0 +1,51 @@
...
+        SelectorBasedAccessManager(_exchange)
+    {
+    }
+
+    function submitBlocksCompressed(
+        bytes calldata data
+        )
+        external
+    {
+        bytes4 selector = IExchangeV3(0).submitBlocks.selector;
```
dong77, line 33: @Brechtpd Ideally I want to use a constant selector instead of doing this read. How to calculate/find this selector?

dong77:
> > Currently, our engineers will have access to the owner's private key to register tokens. Not good.
> 
> > Now, we can manage access in another contract outside of the core exchange.
> 
> This isn't a limitation of the current system of course, both the current system and the system in this PR are identical in functionality, the only difference is the efficiency that can be achieved with both.
> 
> This abstracts the concept of the operator away in the protocol, which at this point makes sense I think! The only reason it wasn't done like that was to make `submitBlocks` as efficient as possible because it could be an EOA address (that also make it easier to parse the DA data), which isn't possible with the owner because the owner is security oriented so would need to be a contract (and indeed would also need to contain extra logic to make e.g. the token registration permissionless using some process). But at this point we also want to make the operator more flexible so a contract there is inevitable as well.
> 
> It could make sense to save a bit of gas on block submissions to allow function calls with compressed data directly on `SelectorBasedAccessManager` (like possible in `submitBlocksCompressed` in `Operator`), this way we don't necessarily have to go through yet another contract (with an overhead for block submissions I think maybe 1-2% overhead per contract hop) if we just submit blocks from an EOA (and the open operator contract that would allow other parties to submit blocks can be done from a different address). But maybe that's not the best place to do it to because upgrading `SelectorBasedAccessManager` will be a lengthy/more difficult process because it also manages all the owner related functions, operator contracts can be switched out immediately without any problem. So let's just keep it as it is now, I will do some measurements later.

I renamed Operator.sol to LoopringIOExchangeOwner.sol. Then I realized that many of our codes are not the core-protocol code, such as DelayedTransaction, *AccessManager, and many others. Maybe we should re-organize the directories so that these auxiliary contract can live their own directories.

packages/loopring_v3/contracts/impl/owners/LoopringIOExchangeOwner.sol
kongliangzhong: This is a sample: bytes4(keccak256("isValidSignature(bytes,bytes)")

dong77: How about functions with Structure parameters.

Brechtpd: > Ideally I want to use a constant selector instead of doing this read

It's not a read, the value is calculated at compile time. There's no difference in cost so I don't see any reason not to use it.

PR #1483, opened 2020-07-27
[Loopring_3.6] extract IDecompressor

dong77:


packages/loopring_v3/contracts/impl/owners/LoopringIOExchangeOwner.sol
```diff
@@ -54,4 +60,15 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
         open = _open;
         emit SubmitBlocksAccessOpened(_open);
     }
+
+    function setDecompressor(address _decompressor)
```
kongliangzhong, line 64: owner 如果能setDecompressor的话相当于放大的owner的权限，owner可以用恶意的Decompressor合约来修改用户提交的参数。

Brechtpd:
This is how it works in 3.1, but this will do another contract call + data/return data copy. I don't think it makes a lot of sense to do it like this I think, if we take the cost to do another contract call like this we may as well do the decompression in an operator contract that can be updated easily and immediately (one of the reasons I kept the operator a separate instance).

Issue #1297, opened 2020-06-13
[Loopring 3.6] Support Gas Station Network for user actions


dong77:
https://docs.opengsn.org/tutorials/

As of now, if a user must have some Ether to perform any exchange related layer-1 actions. If we support gas-less transactions relayed by GSN, users can perform those actions without Ether.

dong77:
Without much knowledge in GSN, I have the following questions: 

1. what's the minimal work we have to do besides modifying our contract to support GSNv2, as described in https://docs.opengsn.org/tutorials/#converting_a_contract_to_support_gsnv2
2. It seems a paymaster is only necessary if we want to pay the fees for our users; otherwise, we don't have to develop or deploy such a contract. The question I have is whether it is easy to frequently change the paymaster contract -- the rules should change based on our operations.
3. What's the best way to handle GSN major/incompatible upgrades? Like many apps, we don't like dependencies :) 
4. How does GSN support (verify signatures of ) contract addresses as `msg.sender`s internally, ERC1271? This seems transparent to us though.

dong77:
@Brechtpd I have this feeling that GSN will probably make our smart wallet module design simpler...

Brechtpd:
One of the reasons Agents exist is to support GSN or other meta transactions. I would say we want to keep these specific implementations out of the main protocol. If we do want to support one of these out of the box we can just whitelist some agents on the protocol contract so they are enabled for all users without them having to enable it specifically.

So I would say:
1. None. We just need to create an Agent contract that implements support for one or more of these meta transactions.
2. Easy to deploy new Agent contracts.
3. Easy to deploy new Agent contracts.
4. Not sure.

For payments, well, if we want people to keep their funds inside the rollup, we kind of also have to charge people for meta transactions inside the rollup I would think, that's where their funds are available. So meta txs done by us using conditional transfers for fee payments makes sense to me, and is as easy as calling `approveOffchainTransfer`. On the other hand, not everyone will store all their funds there, so some other way to pay may be useful as well. But I have to look more into how the Paymaster works.

> I have this feeling that GSN will probably make our smart wallet module design simpler...

Probably, but on the other hand it's also one of the most important parts of a smart wallet. So if you create a smart wallet from scratch it kind of makes sense to have this built in. But I'm all for using existing services/smart contract where we can!

dong77:
Brecht, if we create a Meta-transaction Agent, does each user has to enable that Agent before using it? The user may not have any Ether at all...

[...]

PR #1484, opened 2020-07-27
[Loopring_3.6] Slightly change Agent's design and add OpenGSN2Agent

dong77:
Checkout this comment: https://github.com/Loopring/protocols/issues/1297#issuecomment-664349413

The new Agent design requires:
- if the agent is an EOA, then the behavior should be the same
- If the agent is a contract, it must append the calldata with the real logic sender.


Brechtpd:
The idea with Agents is that the authorization is checked inside the Agent contract. So this could be either with a signature, by checking the address that authorized some meta tx, or even something completely different. For GSN it would be done like this:

```
function approveTransaction(
    address owner,
    bytes32 transactionHash
    )
    external
    override
{
    require(owner == _msgSender(), "UNAUTHORIZED");
    exchange.approveTransaction(owner, transactionHash);
}
```

Basically the GSN design is the same as the exchange design, we just don't use the trick of putting the actual `msg.sender` hidden in the calldata, just directly in the function parameters, which simplifies some things.

This new design is a bit more complicated on the protocol level compared to the old one, but for these simple forwarder contracts every function does need to be implemented manually (though there's only like 6 functions so no biggie).

I like the simplicity of the old approach over the one in this PR (no EOA/contract differences, no packing extra data in the calldata), but essentially they are the same so it boils down to what implementation we prefer.

dong77:
Actually I like the previous design as well. I'll drop this PR.

PR #1497, opened 2020-07-29
more

dong77:


packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -480,7 +474,7 @@ class SubGadget : public GadgetT
         unsafeSub(pb, A, B, FMT(prefix, ".unsafeAdd")),
         rangeCheck(pb, unsafeSub.result(), n, FMT(prefix, ".rangeCheck"))
     {
-        assert(n + 1 <= NUM_BITS_FIELD_CAPACITY);
+        assert(n < NUM_BITS_FIELD_CAPACITY);
```
Brechtpd, line 477: The sum totals n + 1 bits, and it's the sum that needs to be kept in range.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -539,12 +532,8 @@ class TernaryGadget : public GadgetT
...
     }
 
-    void generate_r1cs_constraints(bool enforceBitness = true)
+    void generate_r1cs_constraints()
     {
-        if (enforceBitness)
-        {
-            libsnark::generate_boolean_r1cs_constraint<ethsnarks::FieldT>(pb, b, FMT(annotation_prefix, ".bitness"));
-        }
         pb.add_r1cs_constraint(ConstraintT(b, y - x, y - selected), FMT(annotation_prefix, ".b * (y - x) == (y - selected)"));
```
Brechtpd, line 537: This is quite dangerous. I'm pretty sure this check cannot be removed with how things are now, the circuits probably depend on this in some places.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1719,6 +1744,11 @@ struct SelectorGadget : public GadgetT
     }
 };
 
+// if selector=[1,0,0] and values = [a,b,c], return a
+// if selector=[0,1,0] and values = [a,b,c], return b
+// if selector=[0,0,1] and values = [a,b,c], return c
+// special case,
+// if selector=[0,0,0] and values = [a,b,c], return a
```
Brechtpd, line 1751: As the selector we expect the output of the `SelectorGadget`, so the special case should never trigger, but if that works very much not intended. :)

PR #1533, opened 2020-08-03
[Loopring 3.6] support erc-1155

dong77:
Tests have not been updated yet.

I'm not sure how popular this standard can be in the near future, but if it will, then we can list some of these tokens to distinguish our platform from others that only support ERC20. The extra effort seems to be really small.

### DIFF
--- a/packages/loopring_v3/contracts/iface/ExchangeData.sol
+++ b/packages/loopring_v3/contracts/iface/ExchangeData.sol
@@ -30,7 +30,7 @@ library ExchangeData
     struct Token
     {
         address addr;
-        uint    tid;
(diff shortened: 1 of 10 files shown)


Brechtpd:
I think ERC-1155 may already be quite popular hearing from Matthew. :)

For the implementation this is certainly how it could work by exposing more things to the protocol, and seems to work out okay. It's also possible to not change the main protocol at all and just register the token + id together packed in an address like `address(keccak256(token, id))` and just do the conversion back to (token, id) in the deposit contract because that's the only place it's used. Both have cons/pros so not sure what would be best.

PR #1536, opened 2020-08-03
[Loopring 3.6] format c++ code using LLVM style 

dong77:
`clang-format -i --style=LLVM --sort-includes=false *`

All files compile. @Brechtpd could you run tests, please?

After installing `clang-format`, you can run `npm run formatc` to format all C++ code (except ThirdParty/*)


Brechtpd:
Pushed unrelated fix for a test case caused by outdated test data.

But yes, I am no fan of auto-formatting. :) Especially the 80 character per line limit is mind-boggling ([linus](https://lkml.org/lkml/2020/5/29/1038) agrees). I much prefer consistent code over consistent style, and seeing more of a single function on my screen instead of being able to see 5 different screens next to each other (which I never do), but all of this is of course highly subjective

Issue #1481, opened 2020-07-26
[Loopring 3.6] reoragnize contract code into `core` and `aux`


dong77:
As we are extracting logic from the core protocol into customizable contracts, such as IDepositContract, IAgentRegistry, etc. We should really keeping only the core logics into its own directory `core` and more other non-core contracts into an `aux` directory. The aux directory should contains:
- ownership/operator/access related contracts
- customized deposit contracts
- price feed/sellers
- protocol fee vault
- compressors

So eventually we should have:

- core/iface
- core/impl
- staking
- aux
- lib
- thirdparty

Brechtpd:
Good idea. Let's do that when all open PRs we have now are merged in the protocol-36 branch, otherwise merging will probably be hard.

dong77:
@Brechtpd Lets get this done so that we can resume conversations with auditors to get a new quote.

Brechtpd:
Will do this today.

dong77:
Done

PR #1538, opened 2020-08-03
[protocol-3] Reorganize contracts

Brechtpd:
#1481


dong77:
I pushed two commits:1) rename BasicDepositContract to DefaultDepositContract, 2) sort imports in solidity files.

PR #1548, opened 2020-08-04
[Loopring_3.6] improve SignatureUtil

dong77:
Screen Shot 2020-08-04 at 16 21 57

### DIFF
--- a/packages/loopring_v3/contracts/lib/SignatureUtil.sol
+++ b/packages/loopring_v3/contracts/lib/SignatureUtil.sol
@@ -170,12 +170,11 @@ library SignatureUtil
                 if (recoverECDSASigner(hash, stripped) == signer) {
                     return true;
                 }
-            } else {
-                bytes32 hash = keccak256(
-                    abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(data))
-                );
-                return recoverECDSASigner(hash, stripped) == signer;
             }
+            bytes32 hash = keccak256(
+                abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(data))
+            );
+            return recoverECDSASigner(hash, stripped) == signer;
         } else {
             return false;
         }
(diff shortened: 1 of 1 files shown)


packages/loopring_v3/contracts/lib/SignatureUtil.sol
```diff
@@ -155,15 +158,24 @@ library SignatureUtil
...
+                );
+                if (recoverECDSASigner(hash, stripped) == signer) {
+                    return true;
+                }
+            } else {
+                bytes32 hash = keccak256(
+                    abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(data))
+                );
+                return recoverECDSASigner(hash, stripped) == signer;
+            }
```
Brechtpd, line 178: Will this work as expected? Seems like this will still have the old behaviour where it is assumed bytes of length 32 are already hashed. Totally not sure, it's a bit hard to keep track. :)

dong77: Yes, it works. Basically it does this:
- if the data is 32 byte-long, try to treat it as hash, if it fails, try to use as the raw data of 32 bytes.
- if the data is not 32 long, use it as raw data.

Brechtpd: > if the data is 32 byte-long, try to treat it as hash, if it fails, try to use as the raw data of 32 bytes.

If the data is 32 bytes long, and `recoverECDSASigner(hash, stripped) == signer` is false, it won't get into the `else` no? So it won't try with `keccak256(data)`?

dong77:
I should not have added “else”:(

发自我的iPhone

packages/loopring_v3/contracts/lib/SignatureUtil.sol
dong77: Sharp eyes. I fixed the issue by removing an "else".

PR #1549, opened 2020-08-04
[Hebao_1.1] option-2: meta-tx fees do not consume daily quota

dong77:
The main purpose of this change is to reduce meta-transaction cost overhead.
But this PR actually introduces a security risk, a hacker who steals a wallet's owner private key can deplete the wallet by giving away tokens as meta-tx fees.

Please also see https://github.com/Loopring/protocols/pull/1551
[Please do not merge yet!!!]

### DIFF
--- a/packages/hebao_v1/contracts/modules/base/BaseModule.sol
+++ b/packages/hebao_v1/contracts/modules/base/BaseModule.sol
@@ -202,17 +202,4 @@ abstract contract BaseModule is ReentrancyGuard, Module
     {
         return Wallet(wallet).transact(uint8(3), to, 0, data);
     }
-
-    function reimburseGasFee(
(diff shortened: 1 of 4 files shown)


kongliangzhong:
Without quota updating, GAS_OVERHEAD can be reduced to 40000.

kongliangzhong:
Maybe we can define a maxGasPrice property in Controller,  convert metaTx.gasPrice to ETH price and make sure it is less than the maxGasPrice value.

dong77:
> Maybe we can define a maxGasPrice property in Controller, convert metaTx.gasPrice to ETH price and make sure it is less than the maxGasPrice value.

You don't know which token they will pay for meta-tx fees so you don't know the range of the price.

PR #1552, opened 2020-08-04
[hebao_1.1] meta-tx fee do not consume quota in some situations

dong77:
Meta-tx fees won't consume daily quota in the following 2 situations:
- the target address is the wallet factory
- the meta-tx has a non-zero txAwareHash (which means at least one signature is required from guardian)
All tests pass

### DIFF
--- a/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
+++ b/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
@@ -163,7 +163,9 @@ abstract contract ForwarderModule is BaseModule
 
             // Do not consume quota if the target address is the factory
             // or the meta-tx's txAwareHash is not zero which means it will
-            // be signed by at least a guardian
+            // be signed by at least a guardian, therefor, even if the owner's
+            // private keyis leaked, the hacker won't be able to deplete ether/tokens
+            // as high meta-tx fees.
             bool skipQuota = metaTx.txAwareHash != 0 ||
                 metaTx.to == controller().walletFactory();
 
--- a/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
+++ b/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
@@ -160,12 +160,17 @@ abstract contract ForwarderModule is BaseModule
         // wallet funds.
         if (metaTx.gasPrice > 0 && (metaTx.txAwareHash == 0 || success)) {
             uint gasAmount = gasUsed < metaTx.gasLimit ? gasUsed : metaTx.gasLimit;
+
+            // Do not consume quota if the target address is the factory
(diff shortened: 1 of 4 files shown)


Brechtpd:
> the target address is the wallet factory

Doesn't this mean that if the owner keys are stolen the full smart wallet can be drained to the relayer's wallet just by doing calls to the wallet factory contract? The attacker can just call any function on the factory without any limit, which seems to be a problem. For `createWallet` we could fix that because only the first call to it would be successful so we could force only successful meta-txs, but the attacker may as well just call other public functions like `computeWalletAddress` which are useless but will be successful, so the attacker can create as many successful meta-txs he wants.

dong77:
How about I also check the function selector to make sure we only skip using quota when we can createWallet?

One thing I’m not sure is: will the gas be zero if a wallet transacts on view or pure functions? 

发自我的iPhone

Brechtpd:
> How about I also check the function selector to make sure we only skip using quota when we can createWallet? 

That works if you also require the call to be successful (similar to the `metaTx.txAwareHash != 0` case).

> One thing I’m not sure is: will the gas be zero if a wallet transacts on view or pure functions?

View/pure functions cost gas as usual when used onchain.

dong77:
> > How about I also check the function selector to make sure we only skip using quota when we can createWallet?
> 
> That works if you also require the call to be successful (similar to the `metaTx.txAwareHash != 0` case).
> 
> > One thing I’m not sure is: will the gas be zero if a wallet transacts on view or pure functions?
> 
> View/pure functions cost gas as usual when used onchain.

Fixed, please take another look.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -160,12 +161,23 @@ abstract contract ForwarderModule is BaseModule
...
             uint gasAmount = gasUsed < metaTx.gasLimit ? gasUsed : metaTx.gasLimit;
+
+            // Do not consume quota when call factory's createWallet function or
+            // when a successful meta-tx's txAwareHash is non-zero (which means it will
+            // be signed by at least a guardian). Therefor, even if the owner's
+            // private key is leaked, the hacker won't be able to deplete ether/tokens
+            // as high meta-tx fees.
+            bool skipQuota = metaTx.txAwareHash != 0 && success ||
+                metaTx.data.toBytes4(0) == WalletFactory.createWallet.selector &&
+                metaTx.to == controller().walletFactory();
```
Brechtpd, line 172: Will need to have an extra `&& success` for the `createWallet` case as well, we can't let it fail because that would allow it to be called an infinite number of times.

dong77: Good point.

dong77: fixed

Issue #1503, opened 2020-07-30
[Loopring 3.6] code review DA encoding and decoding


dong77:
Currently, the exchange data is not structured, so we have many `data.toXXX(offsite)` reads. Should we use `abi.decode(data, RollupData)` instead?  

For auxiliary data of each transaction (take withdrawal for example), we have this:
```solidity
 WithdrawalAuxiliaryData memory auxData = abi.decode(auxiliaryData, (WithdrawalAuxiliaryData));
```

Brechtpd:
Standard ABI encoding is pretty wasteful because it adds a lot of zeros. This is no problem with the auxiliary data because zeros are cheap and will be compressed away (and is also not used for the most scalable transactions like trades/transfers). But this is a big problem for the DA because in the circuits all these zeros do need to be passed through SHA256 which will be very expensive.

There is `abi.encodePacked` which doesn't do the padding, but there is no `abi.decodePacked` which can revert to the original struct (this would even be impossible when containing dynamic arrays with how it currently works). So if we would want a function like that we'd have to write it ourselves. Perhaps we can write a function that can do packed -> unpacked so we can use abi.decode, but that doesn't seem to be that much easier because it would need some meta information about the struct we'd have to write manually. Or we could use code generation for this (or even the complete decoding), but that makes things much more complex in the background.

PR #1557, opened 2020-08-05
[protocol 3.6] Increased max num tokens + decreased max num bytes DA

### LINKED ISSUE #1564: [Loopring_3.6] review: missing connections between two inputs in
```cpp
class EdDSA_HashRAM_Poseidon_gadget : public GadgetT {
public:
  Poseidon_gadget_T<6, 1, 6, 52, 5, 1> m_hash_RAM; // hash_RAM = H(R, A, M)
  libsnark::dual_variable_gadget<FieldT> hash;

  EdDSA_HashRAM_Poseidon_gadget(ProtoboardT &in_pb, const Params &in_params,
                                const VariablePointT &in_R,
                                const VariablePointT &in_A,
                                const VariableT &in_M,
                                const std::string &annotation_prefix)
      : GadgetT(in_pb, annotation_prefix),
        // Prefix the message with R and A.
        m_hash_RAM(in_pb, var_array({in_R.x, in_R.y, in_A.x, in_A.y, in_M}),
                   FMT(annotation_prefix, ".hash_RAM")),
        // --- below, hash not connected from/to m_hash_RAM
        hash(pb, NUM_BITS_MAX_VALUE, FMT(annotation_prefix, ".hash")) {}

  void generate_r1cs_constraints() {
    m_hash_RAM.generate_r1cs_constraints();
    hash.generate_r1cs_constraints(true);
  }

  void generate_r1cs_witness() {
    m_hash_RAM.generate_r1cs_witness();
    hash.bits.fill_with_bits_of_field_element(pb, pb.val(m_hash_RAM.result()));
    hash.generate_r1cs_witness_from_bits();
  }

  const VariableArrayT &result() { return hash.bits; }
};
```

It seems from the circuit perspective, there is no connections at all between `m_hash_RAM` and `hash`. Which means the real witness to `hash` can be chosen by the operator without restrictions.

Brechtpd:
- #1546 
- Decreased max number of bytes used by the DA from 106 bytes to 68 bytes. This is done by removing more stuff from the DA that's not actually used in the circuit directly for withdrawals, and replacing it with a short hash. `validUntil` was also removed in the DA from all transactions and passed through the auxiliary data instead, which saves a couple of bytes but now needs to be checked onchain because it's not validated in the circuit any more. The `data` parameter that was available for transfers was also removed as it couldn't be used any more to fill up any left over bytes in the DA for transfers. There are now 3 different transaction types hovering around 68 bytes, which is close to the limit because of the account update tx (requires at least an address + 32 bytes for the public key + some other data).
- Fix #1564 
- Fix https://github.com/Loopring/team/issues/107

Started at 155,000 constraints/tx, increased to 163,000 constraints/tx because the balances tree got bigger, but now at 135,000 constraints/tx after the DA optimizations.

I think I can fairly easily save another ~4,000 constraints by reusing EdDSA point validations, but that'll be another PR.


dong77:
What about the gas cost per trade and transfer tx?

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -79,20 +85,29 @@ library WithdrawTransaction
...
+        withdrawal.to = auxData.to;
+        withdrawal.minGas = auxData.minGas;
+        withdrawal.data = auxData.data;
+        withdrawal.validUntil = auxData.validUntil;
 
         if (withdrawal.withdrawalType == 0) {
             // Signature checked offchain, nothing to do
         } else if (withdrawal.withdrawalType == 1) {
+            // Check validUntil
+            require(block.timestamp < withdrawal.validUntil, "WITHDRAWAL_EXPIRED");
```
dong77, line 110: I'm not sure if we should pass-in the block time as a parameter so for each block, we only read 'block.timestamp' once!

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
```diff
@@ -510,7 +509,6 @@ abstract contract IExchangeV3 is IExchange
         uint96  amount,
         address feeToken,
         uint96  fee,
-        uint    data,
```
dong77, line 513: Brecht, will {from, token, storageID} uniquily identify a transfer? If so, maybe we can simply use these tree variables as the input to approveOffchainTransfer.

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
```diff
@@ -56,12 +62,17 @@ library TransferTransaction
     {
         // Read the transfer
         Transfer memory transfer = readTransfer(data, offset);
+        TransferAuxiliaryData memory auxData = abi.decode(auxiliaryData, (TransferAuxiliaryData));
+
+        // Check validUntil
+        require(block.timestamp < auxData.validUntil, "WITHDRAWAL_EXPIRED");
```
dong77, line 68: Actually I have another idea that may work:

1. Conver the root into a Root Node which contains some global layer2 state variables, such as the smallest and largest `validUntil` values. Then on L2, we only need to check two values against block.timestamp. 

2. `validUntil` should compare with `inputTimestamp`, not with 'block.timestamp'.

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
```diff
@@ -56,12 +62,17 @@ library TransferTransaction
...
+        transfer.validUntil = auxData.validUntil;
 
         // Calculate the tx hash
         bytes32 txHash = hash(ctx.DOMAIN_SEPARATOR, transfer);
 
         // Check the on-chain authorization
-        S.requireAuthorizedTx(transfer.from, auxiliaryData, txHash);
+        S.requireAuthorizedTx(transfer.from, auxData.signature, txHash);
```
dong77, line 75: For transfers with EdDSA signature authorization, the validUntil in the aux data will not be checked online so the operator can use any timestamp value

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -441,7 +441,6 @@ contract ExchangeV3 is IExchangeV3
         uint96  amount,
         address feeToken,
         uint96  fee,
-        uint    data,
```
dong77, line 444: The operator can create more than one transfer with the same parameters but just different "data", so maybe we should include this field. Otherwise, we can simply just use {from, token, storageID} as the unique transfer identifier.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
```diff
@@ -50,12 +57,17 @@ library AccountUpdateTransaction
     {
         // Read the account update
         AccountUpdate memory accountUpdate = readAccountUpdate(data, offset);
+        AccountUpdateAuxiliaryData memory auxData = abi.decode(auxiliaryData, (AccountUpdateAuxiliaryData));
```
dong77, line 60: I feel really bad for asking the same question again: why not use the same way to decode aux data as for the transaction itself? I think we are paying the same gas for aux data as for DA.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -79,20 +85,29 @@ library WithdrawTransaction
...
-            // Hashes are stored using only 253 bits so the value fits inside a SNARK field element.
-            require(
-                uint(keccak256(auxData.auxiliaryData)) >> 3 == uint(withdrawal.dataHash),
-                "INVALID_WITHDRAWAL_AUX_DATA"
-            );
-        }
+        // Validate the withdrawal data not directly part of the DA
+        bytes32 dataHash = keccak256(
+            abi.encodePacked(
+                auxData.minGas,
```
dong77, line 91: maybe we should also include validUntil here.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -191,21 +206,16 @@ library WithdrawTransaction
...
+        withdrawal.tokenID = data.toUint16(offset);
+        offset += 2;
         withdrawal.amount = data.toUint96(offset);
         offset += 12;
+        withdrawal.feeTokenID = data.toUint16(offset);
+        offset += 2;
         withdrawal.fee = uint(data.toUint16(offset)).decodeFloat(16);
         offset += 2;
-        withdrawal.to = data.toAddress(offset);
+        withdrawal.dataHash = data.toBytes20(offset);
```
dong77, line 217: move datahash to the end?

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -33,7 +33,7 @@ library WithdrawTransaction
     using ExchangeWithdrawals  for ExchangeData.State;
 
     bytes32 constant public WITHDRAWAL_TYPEHASH = keccak256(
-        "Withdrawal(address owner,uint32 accountID,uint16 tokenID,uint256 amount,uint16 feeTokenID,uint256 fee,address to,bytes32 dataHash,uint24 minGas,uint32 validUntil,uint32 nonce)"
+        "Withdrawal(address owner,uint32 accountID,uint16 tokenID,uint256 amount,uint16 feeTokenID,uint256 fee,address to,bytes data,uint minGas,uint32 validUntil,uint32 nonce)"
```
dong77, line 36: the order doesn't match the struct definition.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -45,19 +45,25 @@ library WithdrawTransaction
         uint    amount;
         uint16  feeTokenID;
         uint    fee;
+        bytes20 dataHash;
```
dong77, line 48: maybe rename `dataHash` to `auxDataHash`, move to the end of the struct, and add comment like:
"// not part of the original Withdrawal data, not signed but supplied by the operator.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: Checked to make sure, this doesn't read from storage (which is the expensive one) and is practically free, cost is maybe 3-5 gas or something like that, hard to measure exactly because of that.

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
Brechtpd: It won't, because the operator could choose whatever the other transfer parameters are because `from` hasn't signed anything when the transfer is approved like this. So basically it has to contain pretty much all transfer data.

dong77: So may be 'data' should be put inside "TransferAuxiliaryData" and be signed？

dong77:
A more general question: decreasing the number of bytes for DA doesn't mean the number of constraints will be smaller, does it? My understanding is the data cost is the sum of DA + aux data. And they are priced the same.

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
Brechtpd: > Actually I have another idea that may work:
> 
>     1. Conver the root into a Root Node which contains some global layer2 state variables, such as the smallest and largest `validUntil` values. Then on L2, we only need to check two values against block.timestamp.

But we'd still have to check that `minValidUntil <= tx.validUntil <= maxValidUntil` onchain for each conditional tx, and do all that work to just do `minValidUntil < inputTimestamp < maxValidUntil` on layer 2 which would also be practically free on layer 1? Not sure I understand the idea. The simple time check is very cheap either way (50 gas).

>     2. The offchain validation of `validUntil` for EdDSA signed off-chain transactions is done by comparing it with the `inputTimestamp` of the block, so the onchain validation of `validUntil` should also  compare with `inputTimestamp`, not with 'block.timestamp'.

I guess that's right, but why not use the more accurate time when we can, `inputTimestamp` is only there because we can't use 'block.timestamp' in the circuit. Fine with me either way. :)

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
Brechtpd: It should. :) Nothing has actually been changed in the circuit, `validUntil` just isn't passed into the DA any more, so we can't count on it being correct on-chain.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
Brechtpd: See above, needs to contain all the transfer parameters, the approval needs to approve all the data of the transfer. `storageID` is used to make the transfer unique when all else is the same.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
Brechtpd: Would pad too many zeros when encoding the data to be used with `abi.decode` which we need to hash in the circuit which uses a lot of constraints! (as can be seen in this PR, removing ~40 bytes from the DA saves ~25,000 constraints, if they are zero of not doesn't matter for the circuit). And indeed explained here earlier: #1503 :)

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
dong77: I prefer to use `inputTimestamp` as there will be fewer failures and make the logic consistent.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: It's not necessary because the hash is used for approving the data in the circuit with EdDSA. And when EdDSA signatures are used the timestamp is checked in the circuit and `validUntil` is directly signed by the user.

When not using EdDSA the dataHash is actually not used, because all that data that is not part of the DA is approved directly by the user using an ECDSA signature/approval which directly contains `validUntil`/`minGas`/`data`/`to` as parameters of the onchain hash.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: The style I tried to follow is to always do the nonce at the end. Any reason you think `datahash` is better at the end here?

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
dong77: Oh, I actually meant the other way- do not use abi.decode even for aux data.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: The struct tries to make it clear that part of the data is from the DA and the other part of the data is from the aux data. I thought it would be clearer this way, otherwise the parameters from the 2 different sources are mixed. `WITHDRAWAL_TYPEHASH` also does not contain all the data like `dataHash`, so they are not completely the same either way.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
dong77: I agree, that's why I think `data` should not be removed from this method.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: Can do, but `auxDataHash` may also be a bit confusing because it's not all of the data inside the `WithdrawalAuxiliaryData`, and it's also not really "aux data" because it's just standard withdrawal data that always needs to be there, it's just not necessary for the logic inside the circuit. What about `onchainDataHash`?

Brechtpd:
> A more general question: decreasing the number of bytes for DA doesn't mean the number of constraints will be smaller, does it? 

It most certainly does! Data in the DA needs to be hashed with sha256, onchain and in the circuit. Hashing data in the circuit costs ~15,000 constraints/32 bytes. So very expensive. Auxiliary data is not hashed by the circuit.

> My understanding is the data cost is the sum of DA + aux data. And they are priced the same.

That's right, the gas cost for both is the same because both are calldata.

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
Brechtpd: Not sure I follow what you want to achieve here. Why is `data` necessary? And how can it be signed if it is approved using some onchain mechanism?

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
dong77: I agree, i just want to see them as in sync as possible to improve readability.

packages/loopring_v3/circuit/Circuits/WithdrawCircuit.h
```diff
@@ -39,7 +37,7 @@ class WithdrawCircuit : public BaseTransactionCircuit {
   DualVariableGadget nonce;
 
   // Signature
-  Poseidon_gadget_T<12, 1, 6, 53, 11, 1> hash;
+  Poseidon_gadget_T<10, 1, 6, 53, 9, 1> hash;
```
letsgoustc, line 41: Modify the Poseidon hash parameter to save constraint?

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
Brechtpd: I'm not sure I see how `data` helps. `storageID` is the important one here because once the `storageID` is used none of the other transactions with that `storageID` can be executed afterwards. So in theory you could use `data` to make it unique, but only one will be able to be executed. Is that what you want? Using `validUntil` for that would also work then I think.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
Brechtpd: Yeah sure that would work. But `abi.decode` is nice? :) Makes it easy to decode dynamically sized data and stuff, the only reason not to use it would be gas cost, but I don't think that's an issue here because the overhead of abi.decode is quite low, probably comparable with manually extracting the data.

packages/loopring_v3/circuit/Circuits/WithdrawCircuit.h
Brechtpd: Just less inputs to hash because a few were removed. A Poseidon hash is only like 300 constraints, not much too save on in any case for these single hashes. :)

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
dong77: It seems the dataHash doesn't need to be calculated on client-side (and even if it is provided by the user, we may still ignore the value). The operator can calculate this value before putting the tx into a block. That's why I feel it is very special.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
dong77: If the argument is correct, then we can say it's also ok to use abi.decode to decode the `block.data` inside `commitBlock` and the read function of all the XXXTransaction.sol.

dong77: I see the points now.

dong77:
Is it possible to only use one single hash for the DA for each tx, and move all DA data to Aux Data?

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
dong77: > Is it possible to only use one single hash for the DA for each tx, and move all DA data to Aux Data?

Maybe too expensive: gas cost for sha256 is `30 + 6 * (size of input in words)`

Brechtpd:
> Is it possible to only use one single hash for the DA for each tx, and move all DA data to Aux Data?

Basically this is already what is done, but more efficiently by using a single hash so we only need to calculate a single hash onchain and because sha256 has some limitations on how many bytes it can process, so it's more efficient to calculate the hash on longer strings of data.

All DA data of all txs is turned into a single hash, the difference between DA and aux data is just what data passes through the circuit and which data does not. 

So putting a hash in the DA for each tx:
- is pretty wasteful, because now you need to put the hash onchain (~32 non-zero bytes! Much more expensive than the ~15 non-zero bytes for most transfers) AND you still need to put the actual data inside the aux data. So a big increase in costs in most common transactions.
- Also not safe (at least how we do things now). With just a hash we don't enforce any data-avaiability because the hash is not enough to know the data, so we would have to force the aux data to also contain the actual data for each tx. But most of our transactions won't be processed onchain at all (like trades) so these don't need any auxiliary data. But now we'd still have to process those as well to make sure the DA hash matches the actual DA data, so everything gets much more expensive.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
```diff
@@ -50,12 +57,17 @@ library AccountUpdateTransaction
...
+
+        // Check validUntil
+        require(block.timestamp < auxData.validUntil, "WITHDRAWAL_EXPIRED");
+        accountUpdate.validUntil = auxData.validUntil;
```
letsgoustc, line 64: Why need to double check validUntil unchain? In AccountUpdateCircuit, we already has constraint to check:
 // Validate
        requireValidUntil(pb, state.timestamp, validUntil.packed, NUM_BITS_TIMESTAMP, FMT(prefix, ".requireValidUntil")),

This constraint will ensure validUntil < circuit's timestamp. And we already check the circuit's timestamp to consistent with block's timestamp when processing block.

        // Validate timestamp
        uint32 inputTimestamp = _block.data.toUint32(offset);
        offset += 4;
        require(
            inputTimestamp > block.timestamp - ExchangeData.TIMESTAMP_HALF_WINDOW_SIZE_IN_SECONDS() &&
            inputTimestamp < block.timestamp + ExchangeData.TIMESTAMP_HALF_WINDOW_SIZE_IN_SECONDS(),
            "INVALID_TIMESTAMP"
        );

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: The main problem is I think the dynamically sized bytes. To decode those we'd basically have to implement our own simple ABI and decoder for it, at which point it seems to make more sense to just use the built in stuff when possible.

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
Brechtpd: This worked when `validUntil` was part of the DA, which made sure `validUntil` passed through those checks in the circuit. But I removed it from the DA to save some bytes, and now put it into the auxiliary data instead. This data can be freely chosen the the operator and is not validated by the circuit in any way. The `validUntil` in the circuit and the `validUntil` onchain are not related and could have different values (doesn't make sense to do so, but the operator could do that).

dong77: So should we simply remove the logic from the circuit? Or it is used by for example, trades.

Brechtpd: It's still used in the most common cases where an EdDSA signature is used. In that case, the `validUntil` used in the circuit is directly approved by the user so can be used directly. And if it's not a conditional transaction no `validUntil` is ever checked onchain.

letsgoustc: So if we decide to check validUntil onchain, we should remove the requireValidUntil from off chain circuit. Otherwise it's double checked.

packages/loopring_v3/circuit/Circuits/WithdrawCircuit.h
```diff
@@ -362,9 +353,8 @@ class WithdrawCircuit : public BaseTransactionCircuit {
 
   const VariableArrayT getPublicData() const {
     return flattenReverse({type.bits, owner.bits, accountID.bits, tokenID.bits,
-                           feeTokenID.bits, amount.bits, fFee.bits(), to.bits,
-                           dataHash.bits, minGas.bits, validUntil.bits,
-                           nonce.bits});
+                           amount.bits, feeTokenID.bits, fFee.bits(),
+                           dataHash.bits, nonce.bits});
   }
```
letsgoustc, line 358: Missing validUntil.bits？

packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
Brechtpd: It's never double checked, we only check `validUntil` onchain when an onchain signature is used because we cannot depend on the circuit check to be valid. The operator can use any `validUntil` in the circuit unless it is used in combination with an EdDSA signature, because `validUntil` is part of the data signed by the user. If there is no EdDSA signature the operator can put in any `validUntil` into the circuit because it is never authorized anywhere on layer 2 and because we don't put it into the DA there is no link between layer 1 <-> layer 2.

packages/loopring_v3/circuit/Circuits/WithdrawCircuit.h
Brechtpd: Removed to save a couple of bits for optimization.

If `validUntil` would be put here, then yes, we don't have to check `validUntil` onchain. This is how it worked before this PR, but I changed it to save on some sha256 hashing constraints.

I removed it because it's technically not data-availability data, it's not never needed to reconstruct the Merkle tree.

It this is too confusing I can add it back, but it does save quite some constraints (~2,000 constraints/tx).

Brechtpd:
> What about the gas cost per trade and transfer tx?

DA increased by 1 byte because both tokenIDs now need to use the full 16 bits/token. But of course tokenIDs < 256 only use a single non-zero byte so not much change in that case. So depending on how you look at it ~12-16 gas extra for each trade/transfer. But constraints have decreased because of the unrelated DA optimization so more transactions now fit in a single block, which will also save some gas (depending on what the max block size is).

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: Added it to the end.

yueawang:
> It most certainly does! Data in the DA needs to be hashed with sha256, onchain and in the circuit. Hashing data in the circuit costs ~15,000 constraints/32 bytes. So very expensive. Auxiliary data is not hashed by the circuit.

Hi Brecht, for the SHA256 calculation, I think as it first preprocesses the input bits to `448 + 512*K mod 512 bit` aligned blocks and then calculate 1 by 1. Is that to say if we trim DA to the nearest `512*K + 448 bits`, we can save a whole new round of 256 hash kernel calculation?
Considering that, I think both 68bytes(544bit) DA in this PR and previous 106bytes(848bits) DA take 2 rounds of sha256 hash, so the saving of this PR mostly comes from the reduction of input bit assignment rather than the sha256 calculation? Correct me if my understanding is incorrect, Thanks!

Brechtpd:
The data alignment of SHA256 isn't really important for us (unless for very small blocks) because all the tx data is hashed in a single go, not for each tx separately. So every byte that can be saved per tx matters, because blocks will of course contain many hundreds of txs. So for a 1024 tx block, saving 1 byte will end up saving 1024 bytes in that single SHA256 call, directly resulting in saving many rounds in SHA256 over the complete block.

So you're absolutely right that if we wouldn't hash all the data for the block together this could result in many wasted constraints.

packages/loopring_v3/circuit/Circuits/TransferCircuit.h
```diff
@@ -450,11 +440,9 @@ class TransferCircuit : public BaseTransactionCircuit {
   const VariableArrayT getPublicData() const {
     return flattenReverse(
         {type.bits, fromAccountID.bits, toAccountID.bits, tokenID.bits,
-         feeTokenID.bits, fAmount.bits(), fFee.bits(),
-         nonce.getShortStorageID(), da_To.result(), da_ValidUntil.result(),
-         da_StorageID.result(), da_From.result(),
-         VariableArrayT(256 - NUM_BITS_FIELD_CAPACITY, state.constants._0),
-         data.bits});
+         fAmount.bits(), feeTokenID.bits, fFee.bits(),
```
letsgoustc, line 443: Why switch fAmount and feeTokenID sequence? If we switch here, we also need to modify TransferTransaction.sol to keep them consistent

letsgoustc: Forget about the comment, I see the TransferTransaction.sol is also modified.

PR #1558, opened 2020-08-05
[hebao 1.1] enable OfficialGuardian contract to transact

kongliangzhong:


packages/hebao_v1/contracts/modules/OfficialGuardian.sol
```diff
@@ -26,4 +27,31 @@ contract OfficialGuardian is OwnerManagable, ERC1271
         (address addr1, address addr2) = _data.recoverECDSASigner(_signature);
         return isManager(addr1) || isManager(addr2) ?  ERC1271_MAGICVALUE : bytes4(0);
     }
+
+    function addWhitelist(address target)
```
dong77, line 31: add another parameter so we can also remove an address from the whitelist.

PR #1574, opened 2020-08-06
[hebao_1.1] allow wallet to change controller via upgrade

dong77:
这个可能需要好好测试一下才能merge。


packages/hebao_v1/contracts/modules/core/UpgraderModule.sol
```diff
@@ -48,13 +52,37 @@ contract UpgraderModule is BaseModule {
     {
     }
 
+    function upgradeWalletImplementation(address wallet)
+        external
+    {
+        require(msg.sender == address(this), "PROHIBITED");
+
```
letsgoustc, line 59: This means only this module can call it?

dong77: yes

PR #1586, opened 2020-08-09
[Loopring_3.6] (auto) reformat all cpp files

dong77:
@Brechtpd now it's a good time to finetune the .clang-format file for better readability.


Brechtpd:
I experimented a bit and ended up with this slightly changed format:

```
Language: Cpp
BasedOnStyle: LLVM
IndentWidth: 4
ContinuationIndentWidth: 2
ColumnLimit: 140
BinPackArguments: false
SortIncludes: false
BinPackParameters: false
AlignAfterOpenBracket: AlwaysBreak
IndentCaseLabels: true
AllowAllParametersOfDeclarationOnNextLine: false
AllowShortBlocksOnASingleLine: Never
AllowShortCaseLabelsOnASingleLine: false
AllowShortIfStatementsOnASingleLine: false
AllowShortFunctionsOnASingleLine: false
AllowShortLoopsOnASingleLine: false
BreakConstructorInitializers: BeforeComma
BraceWrapping:
  AfterClass:      true
  AfterControlStatement: true
  AfterEnum:       true
  AfterFunction:   true
  AfterNamespace:  true
  AfterObjCDeclaration: true
  AfterStruct:     true
  AfterUnion:      true
  BeforeCatch:     true
  BeforeElse:      true
  IndentBraces:    false
BreakBeforeBraces: Custom
```

If we choose something like this we'll need to do some manual touch ups at some places (but still fully compatible with the style so we only have to do that once) because the formatting makes some bad decisions caused by some of the current line breaks for the constructor initializers so the placement of the commas are wrong. Having multiple of those constructor initializers on the same line or not, depending on how long the line is, is super confusing.

Basically this is me trying to get closer to how it was to be honest (I wrote the code like I preferred it of course). Still not great at a couple of places, but I don't see how those can be fixed with static rules because it depends on context (and clang-format doesn't support everything, like https://reviews.llvm.org/D33029).

dong77:
@Brechtpd I used your new styles, please take a look and feel free to finetune.

dong77:
Screen Shot 2020-08-10 at 09 08 14

I found the colon being at the front of the line is weird though.

Brechtpd:
> I found the comma being at the front of the line is weird though

I know, it is, but clang-format is pretty limited in what it can do. :( I think the other options are worse because they are super inconsistent. `BeforeColon` (and others as far as I know) don't always put the initializers on a new line, so then you get this where sometimes there's more than one on a single line, but mostly just one, and it's super inconsistent and you have to search where they are:

example

Or even a lot of cases where this happens:

example2

Which is just extremely confusing and makes no sense. `BeforeComma` at least puts everything on a new line so it is consistent. I haven't found a way to achieve that in a different way.

dong77:
adding 'ConstructorInitializerAllOnOneLineOrOnePerLine: true' seems to work.

Brechtpd:
`ConstructorInitializerAllOnOneLineOrOnePerLine` I feel still pretty inconsistent in rare cases, I don't know why everything would need to be on a single line when by chance it's below the column width limit. But this luckily doesn't happen too often so perhaps the best option we have.

PR #1587, opened 2020-08-09
[willdrop] rename some gadgets and varables

dong77:


packages/loopring_v3/circuit/Gadgets/OrderGadgets.h
```diff
@@ -96,6 +96,8 @@ class OrderGadget : public GadgetT
 
   void generate_r1cs_witness(const Order &order)
   {
+    pb.val(taker) = order.taker;
```
Brechtpd, line 99: Why move this to the top? It's not in the same order any more as everywhere else?

dong77: restored

PR #1602, opened 2020-08-11
[Loopring_3.6] use mapping for S.blocks

dong77:
@Brechtpd I'm actually not sure this will save gas or not :(

### DIFF
--- a/packages/loopring_v3/contracts/core/iface/ExchangeData.sol
+++ b/packages/loopring_v3/contracts/core/iface/ExchangeData.sol
@@ -176,7 +176,8 @@ library ExchangeData
         bytes32 merkleRoot;
 
         // List of all blocks
-        BlockInfo[] blocks;
+        mapping(uint => BlockInfo) blocks;
+        uint  numBlocks;
 
         // List of all tokens
         Token[] tokens;
--- a/packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
+++ b/packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
@@ -280,7 +280,7 @@ contract ExchangeV3 is IExchangeV3
         view
         returns (uint)
     {
-        return state.blocks.length;
+        return state.numBlocks;
     }
 
     function getBlockInfo(uint blockIdx)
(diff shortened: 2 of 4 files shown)


Brechtpd:
Did some testing with this:
```
uint[] public values;
function addArray(
    uint data
    )
    external
{
    values.push(data);
}

mapping(uint => uint) blocks;
uint  numBlocks;
function addMap(
    uint data
    )
    external
{
    if (data != 0) {
        blocks[numBlocks] = data;
    }
    numBlocks++;
}
```

- addArray(0): 6,832 gas 
- addArray(!0): 26,032 gas
- addMap(0):  6,080 gas
- addMap(!0): 26,958

So a bit surprising. Arrays are ~800 gas cheaper when adding non-zero values (I assume because we don't have to eat the cost for reading the length from storage for some reason), Maps are ~900 gas cheaper when adding zero values (no need to store the 0, which is quite cheap at ~800 gas, but not free). The assumption is that we'll mostly not store the values, so in that case using maps is cheaper. :)

dong77:
> Did some testing with this:
> 
> ```
> uint[] public values;
> function addArray(
>     uint data
>     )
>     external
> {
>     values.push(data);
> }
> 
> mapping(uint => uint) blocks;
> uint  numBlocks;
> function addMap(
>     uint data
>     )
>     external
> {
>     if (data != 0) {
>         blocks[numBlocks] = data;
>     }
>     numBlocks++;
> }
> ```
> 
> * addArray(0): 6,832 gas
> * addArray(!0): 26,032 gas
> * addMap(0):  6,080 gas
> * addMap(!0): 26,958
> 
> So a bit surprising. Arrays are ~800 gas cheaper when adding non-zero values (I assume because we don't have to eat the cost for reading the length from storage for some reason), Maps are ~900 gas cheaper when adding zero values (no need to store the 0, which is quite cheap at ~800 gas, but not free). The assumption is that we'll mostly not store the values, so in that case using maps is cheaper. :)

You are right, we will not store data for quite some time.
Also, we are not actually adding 0s to mapping at all, we simply skip it.

Brechtpd:
> Also, we are not actually adding 0s to mapping at all, we simply skip it.

Yep, and that's the only reason it's cheaper. If the 0 would also be stored for the map the array would be cheaper as well. Storing `0` never costs an actual SSTORE, but it can cost an SLOAD. And even 0 -> 0 costs an SLOAD, which is higher than I thought it would be (but that's the rule: https://eips.ethereum.org/EIPS/eip-2200)

PR #1604, opened 2020-08-11
[Loopring 3.6] log computation vs calldata gas

dong77:


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -32,7 +33,14 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
             hasAccessTo(msg.sender, SUBMITBLOCKS_SELECTOR) || open,
             "PERMISSION_DENIED"
         );
+        uint _gasleft = gasleft();
         bytes memory decompressed = LzDecompressor.decompress(data);
+
+        // Log the gas reduction in calldata and the cost in data decompression.
+        // If (decompressed.length - data.length) * 16 > (_gasleft - gasleft()), then
```
Brechtpd, line 40: Not really accurate like this because non-zero bytes will cost 16 gas, but zero bytes will cost 4 gas. And the compression will mostly target the zero bytes.

I don't think we'll be able to measure this well on-chain, I think we'll have to do the measurements on the actual Ethereum tx, there are also hidden costs elsewhere that can easily be overlooked.

Brechtpd:
The current compression is not good enough, the last few days I've been experimenting with ways to improve it. Any general compression scheme like currently used is simply not good enough to achieve decent compression of the zero bytes. The compression will need to be specific for the tx data so we can make some assumptions. 

It's actually pretty hard to do something that works well, calldata is quite cheap, doing even the most basic thing in the EVM already quickly costs 6-9 gas (an `add` may cost 3 gas, but the EVM is a stack machine so values need to pushed and popped, which also costs 3 gas per op).

dong77:
> The current compression is not good enough, the last few days I've been experimenting with ways to improve it. Any general compression scheme like currently used is simply not good enough to achieve decent compression of the zero bytes. The compression will need to be specific for the tx data so we can make some assumptions.
> 
> It's actually pretty hard to do something that works well, calldata is quite cheap, doing even the most basic thing in the EVM already quickly costs 6-9 gas (an `add` may cost 3 gas, but the EVM is a stack machine so values need to pushed and popped, which also costs 3 gas per op).

I understand what you say, and I think that's the reason I was looking into the decompression code and tests.

Do you think we can do something in the circuits so we somehow can 'select' only the set of bytes that are corresponding to one single sub-circuit? maybe with leading bytes to indicate the type of transactions.

Brechtpd:
> Do you think we can do something in the circuits so we somehow can 'select' only the set of bytes that are corresponding to one single sub-circuit? maybe with leading bytes to indicate the type of transactions.

Isn't this exactly what we currently do? :) Or do you mean that the number of bytes would also be different depending on the transaction type? We always have to support the worst case, and in the worst case all bytes are used. Anything dynamic in the circuit will be very expensive I'm afraid, I can't think of any method that pushing non-zero bytes away (at the end for example) for each tx separately in a way that depends on the transaction type. That would mean we'd have to write the bytes at different positions depending on the transaction type of all transactions before it, which doesn't seem possible in an efficient way.

What I have been thinking about is simply split the DA in 2 parts. The first part is the first ~25 bytes of all txs (generally the number of bytes used for trades/transfers), the second part are the remaining byes for all txs. This way, because a block will be mostly transfers/trades, only the first part will contain mostly non-zero bytes. The 2nd part will mostly be long ranges of zeros that can easily and cheaply be compressed. The downside of this approach is that is makes reading the tx data for the conditional transactions messier, we'd now have to read from 2 separate positions positions and we'd have to make sure none of the data is stored over the boundaries of the 2 parts, otherwise it makes things even messier. It also doesn't work optimally in all cases.

It's still possible to achieve good compression though. I can currently decompress the zero bytes (~44 non-zero bytes at 4 gas = 176 gas) of a trade for ~27 gas/tx without splitting the data up (which is already less than the cost of 2 non-zero bytes, so not too bad). And this is more flexible because it can decompress as many bytes that are actually needed, not just a hard cut in the DA data. We we would do the split at ~25 bytes, that would mean the transfers would still often contain ~10 zero bytes, which costs 40 gas.

Brechtpd:
Doing the split like [0, 24[ and [24, 68[ seems very doable (easy to make sure no data is stored over the boundary at byte 24) and should make the decompression simple and very cheap.

dong77:
I don't like the splitting of DA into 2 parts idea. It's more complicated for data lookup and very bad for third-parties to pass the transactions.

> "in the worst-case all bytes are used" 

I didn't know this. I thought even in the worst-cases there are still many 0 bytes. And that's why I suggested to select out a fixed size subarray of data as large as the worst-case scenario. But seems the current size of DA is already the worst cases.

dong77:
Is it easy to reorder the DA bytes inside the circuit in a tx-type specific way so that we have all zeros at the end of the DA for each tx (For each tx types, we are almost certain which fields are 0s)? If this is possible, we can do very simple domain-specific decompression onchain without using a generic decompression algorithm.

Brechtpd:
> I don't like the splitting of DA into 2 parts idea. It's more complicated for data lookup and very bad for third-parties to pass the transactions.

We actually already do this more aggressively in 3.1 where we group similar data together, so this split in 2 parts is a simplification of that. Easy enough to decode, even on Dune. But I do agree this makes things a bit more complicated, and should be avoided if possible.

> Is it easy to reorder the DA bytes inside the circuit in a tx-type specific way so that we have all zeros at the end of the DA for each tx (For each tx types, we are almost certain which fields are 0s)?

This is already done, zeros are always added to the end (and fields that have the most chance of being zero are added at the back).

> If this is possible, we can do very simple domain-specific decompression onchain without using a generic decompression algorithm - my instinct is that any generic decompression algorithm will be too costly compared to persisting 0s in calldata.

I agree, and that's what my testing has also shown, but 27 gas seems to be the lowest I can get it, even when everything is hard coded. There's a hard limit in what can be achieved this way because we have to do some decompression for each tx, so we have to process each tx on-chain. The minimum operation is `calldatacopy(add(ptr, mOffset), add(pos, cOffset) 25)`, which copies the non-zero bytes for the tx from calldata to memory, for each tx. And just this is 27 gas/tx (and there will be some extra overhead because we'll have to send some meta data as well). If that's acceptable we can do that. Otherwise I currently don't see any other way accept the breaking into 2 parts to get the decompression cost to practically 0 in the most common cases, because that has the advantage that we can do the decompression on-chain in bulk, and not for each tx.

dong77:
Ok, so maybe we should try the 2-part approach. But I think we can consider it as 3.6.x, a new minor release in the near future. We can move forward with the audit using  the current implementation.

PR #1611, opened 2020-08-13
[Loopring 3.6] remove deposit fee

dong77:
https://github.com/Loopring/team/issues/123
Still allow users to send more ETH than required. The extra ETH will be fees.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
+++ b/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
@@ -66,10 +66,7 @@ contract DefaultDepositContract is IDepositContract, Claimable
         external
         onlyOwner
     {
-        require(
-            _feeRecipient != address(0) && _feeRecipient != feeRecipient,
-            "INALID_ADDRESS"
-        );
+        require(_feeRecipient != feeRecipient, "INALID_ADDRESS");
         feeRecipient = _feeRecipient;
     }
 
--- a/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
+++ b/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
@@ -26,6 +26,7 @@ contract DefaultDepositContract is IDepositContract, Claimable
     using MathUint          for uint;
 
     address public exchange;
+    address public feeRecipient;
 
     mapping (address => bool) needCheckBalance;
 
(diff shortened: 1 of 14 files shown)


packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -78,22 +75,17 @@ library ExchangeDeposits
...
-        uint depositValueETH = 0;
-        if (msg.value > 0 && (tokenAddress == address(0) || depositContract.isETH(tokenAddress))) {
-            depositValueETH = amount;
-            fee = uint64(msg.value.sub(amount));
+        if (tokenAddress == address(0) || depositContract.isETH(tokenAddress)) {
+            require(msg.value == amount, "INVALID_AMOUNT");
         } else {
-            fee = uint64(msg.value);
+            require(msg.value == 0, "INVALID_AMOUNT");
         }
```
Brechtpd: I think this can now be removed entirely, the deposit contract can and does check this as well.

packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
```diff
@@ -95,11 +95,9 @@ contract DefaultDepositContract is IDepositContract, Claimable
         returns (uint96 amountReceived)
     {
         if (isETHInternal(token)) {
-            require(msg.value == amount, "INVALID_ETH_DEPOSIT");
+            require(msg.value >= amount, "INVALID_ETH_DEPOSIT");
```
Brechtpd, line 98: Any reason why `>=` now?

dong77: We still allow users to tip the operator for faster deposit processing.

Brechtpd: But
- this sends ETH to `feeRecipient` no matter if the deposit is processed. It's just a donation to the fee recipient with no guarantee any work will be done.
- Deposit aren't processed faster/slower even in 3.1, we just wait for N number of block confirmations before making the amount available. So it's more of do/don't process, not a slower/faster one.

If we want to allow tipping the current system makes more sense. The fee is paid only when that specific deposit is processed, and it's probably even cheaper because this will do an additional ETH transfer for each deposit with tip.

dong77: OK. updated.

PR #1617, opened 2020-08-14
[Loopring_3.6] improve exchange staking

dong77:
Allow the owner of the exchange to stake LRC later but no later than the 10,000 blocks.


dong77:
fixed tests

PR #1633, opened 2020-08-20
[Hebao_1.1] Optimize wallet factory to create uninitialized wallets when gas price is low

dong77:
The general idea is to allow uninitialized wallets (blanks) to be created when gas price is low, then these blanks can be used later when the gas price is high.

When we create blanks and wallets, we can use 'salt's to make sure the address are 'nicer', Chinese people love addresses with certain features, for example, those with 888, 666...


Brechtpd:
So this is basically our own [gas token that uses contract creation](https://medium.com/@1inch.exchange/1inch-introduces-chi-gastoken-d0bd5bb0f92b)? :) Seems like this could have some benefit over just burning gas tokens directly when creating the smart wallet because of some overheads/limitations, but not sure how much.

In any case, gas tokens can be used for everything (meta txs, submitting blocks, etc...) so probably also something we should use in the future.

packages/hebao_v1/contracts/modules/core/WalletFactory.sol
```diff
@@ -98,23 +114,30 @@ contract WalletFactory is ReentrancyGuard
...
             keccak256(_ensApproval),
             _ensRegisterReverse,
             keccak256(abi.encode(_modules))
         );
 
-        bytes32 txHash = EIP712.hashPacked(DOMAIN_SEPERATOR, encodedRequest);
-        require(txHash.verifySignature(_owner, _signature), "INVALID_SIGNATURE");
+        require(
+            EIP712.hashPacked(DOMAIN_SEPERATOR, encodedRequest)
+                .verifySignature(_owner, _signature),
```
kongliangzhong, line 125: the owner can be empty here, so verifySignature is impossible sometimes.

kongliangzhong:
> So this is basically our own [gas token that uses contract creation](https://medium.com/@1inch.exchange/1inch-introduces-chi-gastoken-d0bd5bb0f92b)?

I don't think so. we just want to create some `Adobe wallets` without the owner being set when the ethereum gasprice is low. And we can assign an adobe wallet to some user later. :)

dong77:
> So this is basically our own [gas token that uses contract creation](https://medium.com/@1inch.exchange/1inch-introduces-chi-gastoken-d0bd5bb0f92b)? :) Seems like this could have some benefit over just burning gas tokens directly when creating the smart wallet because of some overheads/limitations, I guess it depends on the gas price by how much.
> 
> In any case, gas tokens can be used for everything (meta txs, submitting blocks, etc...) so probably also something we should use in the future.

Hoss had this idea to create a bunch of uninitialized wallets (but configured with modules) when gas price is low. When a user wants to register and the gas price is higher, we can use one of these uninitialized wallets to save gas.

packages/hebao_v1/contracts/modules/core/WalletFactory.sol
```diff
@@ -58,18 +61,36 @@ contract WalletFactory is ReentrancyGuard
         allowEmptyENS = _allowEmptyENS;
     }
 
-    function computeWalletAddress(
```
kongliangzhong, line 61: We need this function because we must get the wallet address ahead of its creation so we can compute the ensApproval.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -172,8 +181,9 @@ abstract contract ForwarderModule is BaseModule
             // private key is leaked, the hacker won't be able to deplete ether/tokens
             // as high meta-tx fees.
             bool skipQuota = success && (
-                metaTx.txAwareHash != 0 ||
-                metaTx.data.toBytes4(0) == WalletFactory.createWallet.selector &&
+                metaTx.txAwareHash != 0 || (
+                    metaTx.data.toBytes4(0) == WalletFactory.createWallet.selector ||
+                    metaTx.data.toBytes4(0) == WalletFactory.createWallet2.selector) &&
```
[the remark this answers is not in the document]
dong77: Good catch. I created an issue for Kongliang to fix.

dong77:
> I think if the owner is set at zero for the blank wallets then currently everyone can create valid signatures for the wallet because of a strange behavior in SignatureUtil. If the signer is `0`, then it seems `verifySignature` will return true regardless, which very likely can be abused:
> 
> * https://github.com/Loopring/protocols/blob/5b307cd1d4ff05c67b85d54ff0b3b70a680b60e7/packages/hebao_v1/contracts/lib/SignatureUtil.sol#L100-L132
> * https://github.com/Loopring/protocols/blob/5b307cd1d4ff05c67b85d54ff0b3b70a680b60e7/packages/hebao_v1/contracts/lib/SignatureUtil.sol#L148
> * https://github.com/Loopring/protocols/blob/5b307cd1d4ff05c67b85d54ff0b3b70a680b60e7/packages/hebao_v1/contracts/modules/core/ERC1271Module.sol#L56
> 
> I previously added an extra check for that for the protocol to be sure this can't be abused, but probably the lib should do it internally as well so this can't be forgotten:
> 
> * https://github.com/Loopring/protocols/blob/5b307cd1d4ff05c67b85d54ff0b3b70a680b60e7/packages/loopring_v3/contracts/core/impl/libexchange/ExchangeSignatures.sol#L27

I created another issue for Kongliang to fix.

kongliangzhong (approved):

PR #1658, opened 2020-08-25
[protocol 3.6] AMM

Brechtpd:
Core protocol additions:
- AMM token weight in balance leaf
- AMM feeBips in account leaf
- `AMM_UPDATE` transaction which
  - Updates the AMM fee bips in an account leaf
  - Updates the AMM weight in a balance leaf
  - Exposes the balance in the DA so it can be read on L1 (see below)
- Forced withdrawals resets the AMM weight of the token being withdrawn to automatically disable AMM for that token.
- The relayer can now create arbitrary orders for accounts between 2 tokens with non-zero token AMM weight, as long as the fill amounts match the minimum rate of the AMM, which is the balancer/bancorV2 formula.
- Extra getters were added for most of the onchain operations on the exchange so progress can be watched onchain in other smart contracts.

All in all, quite small a number of changes.

Extra additions:

The biggest challenge is how to create a pool on top of the exchange which works as easy as possible. LP tokens need to be minted/burned when depositing/withdrawing from the pool, which I think only really works when the actual balance in the AMM is known. This is also how uniswap/balancer do the calculations so it would be great if we can stick to that as close as possible. On its own the exchange doesn't offer a lot of certainty: operations may or may not be done, at any time in the future at some different state,...

Currently I implemented it in a way that is built on top of the exchange instead of inside the protocol. It's also very simple onchain, flexible and efficient, but counts on some extra offchain logic to create the necessary rollup transactions. Some extra logic offchain might be unavoidable because of the necessary synchronization at some point. The idea is that the DA of the transactions is used to synchronize L1 and L2. On L1 we know exactly what needs to be done, and so on L2 we know it as well, but on L1 we do need some data of L2 ( the AMM balances) to do the necessary calculations and we need to make sure L2 transactions are done in the expected order and at the expected time. Basically it's just a different reason/method L2 transactions are included in a block, and instead of waiting until they are approved on L1, the approval on L1 is delayed until we know exactly how in which order they are already processed in the block. We also need to make sure the pool can be safely exited by all LPs in all circumstances, which means exposing forced withdrawals and making sure we do the necessary bookkeeping at all times. 

Most important part can be seen in `onSubmitBlocks` in `AmmPool.sol` (completely untested). Any number of pool joins/exits can be done in `2*numTokensInPool` L2 operations at any position in a block with other operations. At any point the relayer can choose to insert the L2 transactions necessary to update pools in a block. I'm quite happy with the onchain part of things, but it would be great if the offchain logic would not need to check additional smart contracts for potential L2 work, I'm not sure how much extra complexity this adds to the relayer (though maybe it would simply be an extra service that would do this and create the L2 transactions).


dong77:
> Any number of pool joins/exits can be done in 2*numTokensInPool L2 operations at any position in a block with other operations. 
Brecht, can you elabrate on it.

Brechtpd:
LPs first need to lock up their deposits in the pool contract so we can be sure the deposited amounts are available when we actually deposit to the AMM account when we have all the necessary info.

Joins and exits to the pool are queud on L1 so we can properly process them with the latest balances in the AMM account on L2 and in the expected order.

If `x` people want to deposit and `y` people want to withdraw, all these operations are queued, and after processing this queue we only need to 1 deposit/withdraw per token for the actual balance change to the AMM account on L2, everything else is just internal bookkeeping in the pool contract. This is where `1*numTokensInPool` L2 transacitons come from.

The other `1*numTokensInPool` are AMM update transactions which do 2 things:
- Make sure the weights/fee are properly set on the AMM account on L2
- Make all the balances of the AMM account available on L1 for the different pool tokens. This allows us to work on the actual balances in the AMM account on L1 without having to withdraw or something similar to get this info on L1.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1884,6 +1861,430 @@ class OwnerValidGadget : public GadgetT
...
+        rangeCheck.generate_r1cs_constraints(true);
+    }
+
+    const SignedVariableT result() const
+    {
+        return SignedVariableT(sign.result(), value.result());
+    }
+};
+
+// A - B = sum with A, B and sum < 2^n
```
letsgoustc, line 2011: sum --> sub

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1884,6 +1861,430 @@ class OwnerValidGadget : public GadgetT
...
+
+    LeqGadget a_leq_b;
+
+    EqualGadget signsEqual;
+    TernaryGadget temp;
+    TernaryGadget value;
+
+    AndGadget signB_and_a_leq_b;
+    AndGadget signA_and_not_a_leq_b;
+    OrGadget sign_math;
```
letsgoustc, line 1923: sign_math --> sign_match ?

packages/loopring_v3/contracts/lib/Poseidon.sol
```diff
@@ -345,86 +345,103 @@ library Poseidon
...
-        o.t4 = addmod(o.t4, mulmod(i.t4, 13001155522144542028910638547179410124467185319212645031214919884423841839406, q), q);
-        o.t4 = addmod(o.t4, mulmod(i.t5, 16037892369576300958623292723740289861626299352695838577330319504984091062115, q), q);
-        o.t5 = mulmod(i.t0, 15162889384227198851506890526431746552868519326873025085114621698588781611738, q);
-        o.t5 = addmod(o.t5, mulmod(i.t1, 13272957914179340594010910867091459756043436017766464331915862093201960540910, q), q);
-        o.t5 = addmod(o.t5, mulmod(i.t2, 9416416589114508529880440146952102328470363729880726115521103179442988482948, q), q);
-        o.t5 = addmod(o.t5, mulmod(i.t3, 8035240799672199706102747147502951589635001418759394863664434079699838251138, q), q);
-        o.t5 = addmod(o.t5, mulmod(i.t4, 21642389080762222565487157652540372010968704000567605990102641816691459811717, q), q);
-        o.t5 = addmod(o.t5, mulmod(i.t5, 20261355950827657195644012399234591122288573679402601053407151083849785332516, q), q);
+        HashInputs7 memory o;
+        o.t0 = mulmod(i.t0, 14183033936038168803360723133013092560869148726790180682363054735190196956789, q);
```
letsgoustc, line 365: How these bit number is calcualted? Any material for introduction?

Brechtpd: Automatically calculated from the python implementation: https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/util/generate_poseidon_EVM_code.py

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+        bytes                memory auxiliaryData
+        )
+        public
+        override
+        online
+        onlyExchangeOwner
+        returns (uint numTransactionsConsumed)
+    {
+        AuxiliaryData memory auxData = abi.decode(auxiliaryData, (AuxiliaryData));
+        ExchangeData.Block memory _block = blocks[blockIdx];
```
dong77, line 359: Is it true that all the layer-2 join/exit transactions for an AMM account must be included into the same block?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+
+        // Process work in the queue
+        for (uint n = queuePos; n < queuePos + auxData.numItems; n++) {
+            QueueItem memory item = queue[n];
+            // TODO: all the necessary pool logic, mint/burn LP tokens,...
+            if (item.workType == WorkType.JOIN) {
+                PoolJoin memory join = item.join;
+                for (uint i = 0; i < tokens.length; i++) {
+                    uint amount = join.maxAmountsIn[i];
+                    ammBalances[i] = ammBalances[i].add(amount);
```
dong77, line 391: I think the logic here is incomplete: we need to calculate for each Join/Exit exactly how much token can be added to/removed from ammBalances[i]/locked[i] based on the new layer-2 AMM balances.

Brechtpd:
> A couple of questions regarding the weight. What is the range of valid weight values?

There is currently no real restriction, except ones that are implementation dependent because of overflow. And in this case the actual weight value doesn't matter, just the ratio between the weights of the 2 tokens. As far as I currently know, `2**(96-14)/(10**18)` is the max ratio before the swap formula will overflow, which will simply make swaps impossible until the weight are updated again.

> I wonder if it is possible to allow adding liquidity of only one token into the pool which requires changing the weight value on L2 and L1.

This is possible for sure, depends on the pool implementation. But in no way is it required to do so (balancer doesn't do it, the only risk is slippage which should be fine when deposit << pool balance. Depositing a single token just means using the pool itself to buy the other tokens in the pool automatically).

> Also, it seems the core protocol doesn't have a restriction on whether AMMUpdate will actually change the weight of a token or not, therefore, we can design a AMM pool contract to allow update of single token's weight. Is my understanding correct?

That's right, the only thing the protocol enforces is the formula, nothing else. Weights can be updated and set to any values as required in the pool contract.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,519 @@
...
+        _;
+    }
+
+    modifier offline()
+    {
+        require(!isOnline(), "NOT_OFFLINE");
+        _;
+    }
+
+    function setupPool(
```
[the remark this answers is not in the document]
Brechtpd: Depends on if we want pools to be upgradeable or not.

But just that we're on the same page of this pool implementation: for now this is just for testing (this is why I added the contract to the test folder). While some of this code will definitely be used in some form, it'll depend on how we go forward on the actual pool implementations, otherwise some of this code will simply be used as the glue between Loopring and e.g. the balancer pool contracts.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+                require(withdrawal.to == address(this), "INVALID_TX_DATA");
+                require(withdrawal.extraData.length == 0, "INVALID_TX_DATA");
+                // Now approve this withdrawal
+                withdrawal.validUntil = 0xffffffff;
+                bytes32 txHash = WithdrawTransaction.hashTx(exchangeDomainSeparator, withdrawal);
+                exchange.approveTransaction(address(this), txHash);
+                numTransactionsConsumed++;
+                // Total balance in this contract increases by the amount withdrawn
+                totalBalance[tokens[i].addr] = totalBalance[tokens[i].addr].add(amount);
+            }
```
dong77, line 458: Let me know if my understanding is correct:

1. The relayer determines how many queued Join/Exit requests to process. Once the starting and ending indices are chosen, the relayer knows the new balances in this AMM account based on the most up-to-date L2 balances.

1. But before processing the Joins/Exists, the relayer must roll up the most up-to-date L2 balance per token using an AmmUpdate transaction (per token). These **n** (*n is the number of tokens in the pool*)  transactions are put in the block next to each other.

1. The relayer then includes (up to) **n** Deposits/Withdrawals into the block in case any token's balance changed. Same as the AmmUpdate transactions, these deposit/withdraw transactions don't exist yet, but the relayer is assured that once notified, the AMM contract will create these requests onchain before block submission. If a token's balance doesn't change at all (the aggregated deposit amount == the aggregated exit amount), then there is no deposit/withdrawal for the corresponding token.

The AMM contract must be notified with the AmmUpdate transactions first, so it will be able to calculate (in the same way as the relayer) how much to deposit/withdraw per token, so it can authorize deposit/withdrawal requests onchain. It is also important for the AMM contract to min/burn ownership tokens.

packages/loopring_v3/circuit/Circuits/AmmUpdateCircuit.h
```diff
@@ -0,0 +1,129 @@
...
+#include "../Utils/Data.h"
+
+#include "ethsnarks.hpp"
+#include "utils.hpp"
+
+using namespace ethsnarks;
+
+namespace Loopring
+{
+
```
dong77, line 17: Add a comment: Brings a token's L2 balance to Layer1.

Question1: is this tx only available to AMM account? 

Question2: besides the nonce, what data will also be changed after such a tx?

dong77 (commented):
> LPs first need to lock up their deposits in the pool contract so we can be sure the deposited amounts are available when we actually deposit to the AMM account when we have all the necessary info.
> 
> Joins and exits to the pool are queud on L1 so we can properly process them with the latest balances in the AMM account on L2 and in the expected order.
> 
> If `x` people want to deposit and `y` people want to withdraw, all these operations are queued, and after processing this queue we only need to 1 deposit/withdraw per token for the actual balance change to the AMM account on L2, everything else is just internal bookkeeping in the pool contract. This is where `1*numTokensInPool` L2 transacitons come from.
> 
> The other `1*numTokensInPool` are AMM update transactions which do 2 things:
> 
> * Make sure the weights/fee are properly set on the AMM account on L2
> * Make all the balances of the AMM account available on L1 for the different pool tokens. This allows us to work on the actual balances in the AMM account on L1 without having to withdraw or something similar to get this info on L1.

AMM Update transactions come before the deposit/withdrawal transactions in a block, correct?

dong77:
@Brechtpd It seems we can change the weight of a single token in an AMM account if the AMM contract allows it. The core protocol doesn't even care. Correct?

Brechtpd:
> AMM Update transactions come before the deposit/withdrawal transactions in a block, correct?

Yes, so we know the balance in the AMM account before deposits/withdrawals are processed.

> It seems we can change the weight of a single token in an AMM account if the AMM contract allows it. The core protocol doesn't even care. Correct?

Correct.

packages/loopring_v3/contracts/aux/access/ISubmitBlocksCallback.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "../../core/iface/ExchangeData.sol";
+
+/// @title ISubmitBlocksCallback
+/// @author Brecht Devos - <brecht@loopring.org>
+abstract contract ISubmitBlocksCallback
```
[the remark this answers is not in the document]
Brechtpd: Fine with me, though the `DataReceiver` part may also be a bit confusing in that it's mostly used passively while it actively can do a lot of things.  `ISubmittedBlockReceiver`? I guess it's also more like a processor on the transaction level and not so much the full block level (because it also takes a transaction idx). `ISubmittedBlockProcessor`?

> IBeforeSubmissionBlockDataReceiver? maybe in the future we can also add IPostSubmissionBlockDataReceiver

If we need it we I think we may as well add a before/after to the same interface (with only small overhead if only one of the functions is needed).

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -44,6 +47,46 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
         }
     }
 
+    function submitBlocksWithCallbacks(
+        ExchangeData.Block[]    memory blocks,
+        ISubmitBlocksCallback[] memory callbacks,
+        uint[]                  memory blockIndices,
+        uint[]                  memory txIndices,
+        bytes[]                 memory auxiliaryData
```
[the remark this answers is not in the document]
Brechtpd: > For multiple receivers, the order to notify them doesn't matter. But for the same receiver, a strict ascent order of block indices and for each block a strict ascent order of transaction indices matters, and can be checked by a pure function that takes ReceiverNotification as input.

Not saying we need to make this completely generalized because we can always update this logic when needed, but I think this does depend on what we currently do and is not necessarily always the case (no strict 1-to-1 mapping between contract and account). Though even in the current use I think it only matters in very select cases where we need to make sure block transactions can only be used once.

So this is for optimizing the number of calls needed to a single pool right? If we think this scenario will happen a lot (which I assumed it wouldn't) this makes sense to save a bit of gas in special circumstances.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+    uint public constant MAX_AGE_REQUEST_UNTIL_POOL_SHUTDOWN = 7 days;
+
+    IExchangeV3 public exchange;
+    uint32      public accountID;
+
+    uint        public shutdownTimestamp = 0;
+
+    struct PoolJoin
+    {
+        uint poolAmountOut;
```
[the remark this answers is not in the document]
Brechtpd: The amount of liquidity tokens the LP expects to receive/spend. Used what balancer does here to check that we can do what we need to do, but I haven't actually implemented LP tokens here because hopefully we won't have to do that ourselves.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+                );
+            }
+        }
+
+        shutdownTimestamp = block.timestamp;
+    }
+
+    // Only used to withdraw from the pool when shutdown.
+    // Otherwise LPs should withdraw by doing normal queued exit requests.
+    function withdrawFromPoolWhenShutdown(uint poolAmountIn)
```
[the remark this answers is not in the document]
Brechtpd: Amount of LP tokens that will be spent in exchange to withdraw the corresponding % from the pool.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,461 @@
...
+            require(update.tokenID == tokens[i].tokenID, "INVALID_TX_DATA");
+            require(update.feeBips == feeBips, "INVALID_TX_DATA");
+            require(update.tokenWeight == tokens[i].weight, "INVALID_TX_DATA");
+            // Now approve this AMM update
+            update.validUntil = 0xffffffff;
+            bytes32 txHash = AmmUpdateTransaction.hashTx(exchangeDomainSeparator, update);
+            exchange.approveTransaction(address(this), txHash);
+            numTransactionsConsumed++;
+            // AMM account balance now available onchain
+            initialAmmBalances[i] = update.balance;
```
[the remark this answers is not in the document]
Brechtpd: Renamed to `ammBalancesInAccount` because I found `refreshed` to be a bit confusing as well. :)

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: With the current implementation that's certainly the case. Makes things much easier and because the max amount of L2 transactions needed is just `2*numTokensInPool` regardless of the number of joins/exits it doesn't seem to be much of a limitation.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Yep for sure, lots of things missing here. Amount of funds actually deposited/withdrawn, mint/burn LP tokens based on the amounts deposited/withdrawn, checking if the LP actually wants to deposit/withdraw at the current balances in the account, if the number is worse than expected don't deposit/withdraw anything at all, etc...

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: All correct. :)

packages/loopring_v3/circuit/Circuits/AmmUpdateCircuit.h
Brechtpd: > is this tx only available to AMM account?

This transaction is available to all account owners at any time. There isn't really a thing like an AMM account. AMM orders between 2 tokens are enabled on an account when those 2 tokens have a non-zero weight, that's the only requirement.

>  besides the nonce, what data will also be changed after such a tx?

Only the expected:
- AMM fee bips stored in the account
- AMM weight of the specified token

This transaction simply sets what's defined in the transaction itself in the Merkle tree, no other checks are done. Data is just copied to the Merkle tree.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1884,6 +1861,476 @@ class OwnerValidGadget : public GadgetT
...
+        sign.generate_r1cs_constraints();
+    }
+
+    const SignedVariableT result() const
+    {
+        return SignedVariableT(sign.result(), value.result());
+    }
+};
+
+// Calculates [0, 1]**[0, inf[ using an approximation. Best accuracy is achieved around 1**y.
```
dong77, line 2129: `[0, 1]**[0, inf]` typo?

Do you mean calculating power(a, b), where 0<=a<=1 and b >=0?

dong77: what's `1**y`?

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1884,6 +1861,476 @@ class OwnerValidGadget : public GadgetT
...
+        rangeCheck.generate_r1cs_constraints(true);
+    }
+
+    const SignedVariableT result() const
+    {
+        return SignedVariableT(normalizedSign.result(), value.result());
+    }
+};
+
+// A + (-B) = sum with abs(A), abs(B) and abs(sum) < 2^n
```
dong77, line 1994: maybe also change to:
```
// sA + (-sB) = sum with abs(sA), abs(sB) and abs(sum) < 2^n
```

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1884,6 +1861,476 @@ class OwnerValidGadget : public GadgetT
...
+        : sign(make_variable(pb, FMT(prefix, ".sign"))), value(make_variable(pb, FMT(prefix, ".value")))
+    {
+    }
+
+    SignedVariableT(const VariableT &_sign, const VariableT &_value) : sign(_sign), value(_value)
+    {
+    }
+};
+
+// A + B = sum with abs(A), abs(B) and abs(sum) < 2^n
```
dong77, line 1888: // sA + sB = sum with abs(sA), abs(sB) and abs(sum) < 2^n

packages/loopring_v3/contracts/aux/access/ISubmitBlocksCallback.sol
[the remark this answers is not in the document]
dong77: *XxxBlockReceiver* is fine.

packages/loopring_v3/contracts/test/AmmPool.sol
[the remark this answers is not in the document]
dong77: I would suggest to remove them or rename them to something like "ownershipTokenBurnt/Mint"

packages/loopring_v3/contracts/test/AmmPool.sol
dong77: Ok, just make sure what's needed to be done for a complete impl.

dong77:
@Brechtpd  I like the current design which doesn't change the core protocol a lot. What I'm not certain is how accurate AMM trading related calculation will be. I would suggest to add enough test cases, if you haven't done that already, to ensure the circuit does the verification job correctly - the power calculation is new.

packages/loopring_v3/circuit/Circuits/AmmUpdateCircuit.h
```diff
@@ -0,0 +1,131 @@
...
+          owner(pb, state.accountA.account.owner, NUM_BITS_ADDRESS, FMT(prefix, ".owner")),
+          accountID(pb, NUM_BITS_ACCOUNT, FMT(prefix, ".accountID")),
+          tokenID(pb, NUM_BITS_TOKEN, FMT(prefix, ".tokenID")),
+          feeBips(pb, NUM_BITS_AMM_BIPS, FMT(prefix, ".feeBips")),
+          tokenWeight(pb, NUM_BITS_AMOUNT, FMT(prefix, ".tokenWeight")),
+          nonce(pb, state.accountA.account.nonce, NUM_BITS_NONCE, FMT(prefix, ".nonce")),
+          balance(pb, state.accountA.balanceS.balance, NUM_BITS_AMOUNT, FMT(prefix, ".balance")),
+
+          // Increase the nonce
+          nonce_after(
```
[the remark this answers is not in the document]
Brechtpd: It' can currently only be approved onchain, no offchain signature is possible.

First I did it similarly to deposits, but I changed it because we have to make sure the request is unique somehow (which the nonce easily does) and we need to be able to have multiple requests open at the same time to not run into strange issues. Unlike deposits, we can't just process the AMM updates partially or only the latest data, so we need to store data for each request separately, otherwise it's impossible to have multiple requests open/account which can give all sorts of problems when the pool needs to be updated multiple times. That's why I think it's much easier to just do the same as we do for other requests and not do anything we don't do for any other request types, withdrawals included, which increase the nonce as well, so it's actually more uniform with other transactions. The good thing about the current system is that the contract doesn't even have to worry about nonces at all, these come directly from the block data so no need to track them onchain.

> and we implicit restrict the owner of any Amm account to be a contract address.

Why does this matter for the protocol? An Ethereum address is an Ethereum address? The requests don't even have to be processed so it's up to the relayer what it wants to do.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
Brechtpd: > [0, 1]**[0, inf] typo?

Infinity is not part of the range because it's impossible, so the bracket is open to show that, but not 100% sure if that's the correct notation in this case.

> Do you mean calculating power(a, b), where 0<=a<=1 and b >=0?

Yes.

> what's 1**y?

`power(1, y)`. Just trying to say that the closer the base is to 1, the better the accuracy.

Brechtpd:
> @Brechtpd I like the current design which doesn't change the core protocol a lot. What I'm not certain is how accurate AMM trading related calculation will be. I would suggest to add enough test cases, if you haven't done that already, to ensure the circuit does the verification job correctly - the power calculation is new.

Yeah, a lot more testing still needed. That's what I'll be focussing on the next days, both for the actual trading logic and the circuit gadgets (added more tests for the gadgets today, more still needed).

The accuracy also depends on the number of iterations which is currently set to 4, but will need to make sure that's sufficient in all cases. The spot price gadgets doesn't have accuracy problems, and they are only there to ensure nothing can go drastically wrong by ensuring the price after a swap can never decrease, which provides a good sanity check.

packages/loopring_v3/contracts/test/AmmPool.sol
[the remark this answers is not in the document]
Brechtpd: Uniswap calls them "liquidity tokens", Balancer/Bancor calls them "pool tokens". Would stick with one of them so we don't name things differently.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
dong77: `[0, inf[` =>`[0, inf)`

dong77: >Just trying to say that the closer the base is to 1, the better the accuracy.
Maybe just say the above in the sourcefile.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -0,0 +1,465 @@
...
+import "../lib/AddressUtil.sol";
+import "../lib/ERC20.sol";
+import "../lib/ERC20SafeTransfer.sol";
+import "../lib/MathUint.sol";
+
+import "../core/impl/libtransactions/AmmUpdateTransaction.sol";
+import "../core/impl/libtransactions/DepositTransaction.sol";
+import "../core/impl/libtransactions/WithdrawTransaction.sol";
+
+contract AmmPool is IBlockReceiver {
```
dong77, line 19: @Brechtpd please just add a simple comment saying that this contract is not fully implemented and serves as a demo of how AMM pool may be created on top of Loopring.

packages/loopring_v3/test/testExchangeRingSettlement.ts
```diff
@@ -43,6 +43,52 @@ contract("Exchange", (accounts: string[]) => {
...
+          tokenB: "GTO",
+          amountS: new BN(web3.utils.toWei("99", "ether")),
+          amountB: new BN(web3.utils.toWei("200", "ether")),
+          balanceS: new BN(web3.utils.toWei("10000", "ether")),
+          balanceB: new BN(web3.utils.toWei("20000", "ether")),
+          feeBips: 0,
+          amm: true
+        },
+        orderB: {
+          tokenS: "GTO",
```
dong77, line 60: so orderB represents an AMM pool? Where is the account ID/owner?

dong77: Is it possible to match two AMM pools?

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
Brechtpd: > [0, inf[ =>[0, inf)

Didn't know a parenthesis could also be used for this, both are standard notation it seems: https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

packages/loopring_v3/test/testExchangeRingSettlement.ts
Brechtpd: > so orderB represents an AMM pool? 

OrderA is the AMM order (+ initial balances) that can be used without a signature of the order owner, orderB is the taker of the AMM order.

> Where is the account ID/owner?

To make writing tests easier these don't need to be filled in if they just need default values, `setupRing` will fill in the missing data.

> Is it possible to match two AMM pools?

Yes, AMM can currently be enabled for both accounts at the same time.

dong77:
@Brechtpd with the current design, is it possible to add liquidity to an AMM pool by using layer-2 transfers, then layer-1 AMM pool contract can access which addresses and how much liquidity has been added to /removed from the pool, then mint or burn liquidity tokens accordingly?

Brechtpd:
Should be no problem, but just doing the offchain transfers as usual with an offchain signaturen there are some problems which the user can't control and the pool contract may depend upon (needing to know the exact balance of the pool etc...). And because the user has no control over when the transfer is included in a block it may have unexpected or incorrect results for the pool logic (which cannot be stopped, the transfer would be completely valid from the protocol's view).

So I think the best approach would probably be something very similar to the current approach, but the L1 deposits/withdrawals are done using L2 transfers and the user just provides an onchain signature for the L2 transfer that approves the transfer from/to the pool account only when the transactions are in the expected order (the pool contract needs to be an Agent of the user to approve these transfers).

dong77:
@Brechtpd In the current AmmPool contract, AmmUpdate transactions come before the actual deposits/withdrawals of liquidity. But we can also implement the contract in such a way that *n* AmmUpdate transactions are followed by *m* (*m <= n*) deposit/withdrawals, then followed by *m* AmmUpdate transactions. Then the Amm Pool contract will learn the old weights/balances before addition/removal of liquidities, and the new ones, so that it can verify the weights are modified correctly.

Brechtpd:
Yeah sure, that's totally possible if weights need to be updated for some reason.

Btw I think basically any logic is possible with the current system. The smart contract generates layer 2 operations when possible/wanted and provides the higher level logic building on the basic building blocks available on layer 2. You could build all kind of advanced functionality on top of this (or even reimplement things like doing a trade by using L2 transfers), with the downside of course that it's less efficient than doing it fully on layer 2.

dong77 (approved):

yueawang:
Should we separate the Spot trade and AMM trade? so we have 2 different types of trade tx, and each manages its trading and fee separately. I thought the current implementation added more complexity to SpotTrade. if there is a AmmTrade tx, in circuit we can fix AMM account/order to maker, and check the maxFeeBips of the opponent. So far I did not find such a check, i.e., is it possible that ammFeeBips > maxFeeBips?

2 more questions about the AMM price/volume calculation. 
1. What's the purpose of BASE_FIXED?? I think values like balanceIn/balanceOut already based on 10**18 unit, is it??
Also, why do weightIn/Out need to multiply BASE_FIXED, see below 2 code snippets which come from the comments of volume and price calculation. 
```c++
//   const weightRatio = (weightIn * BASE_FIXED) / weightOut;
```

```c++
//     const numer = balanceIn * weightOut;
//     const denom = balanceOut * weightIn;
//     const ratio = (numer * BASE_FIXED) / denom;
```

2. Is the PowerGadget the Taylor expansion of pow(x, y)?
```c++
/*
    const x = (_x - BASE_FIXED);
    const bn = [BASE_FIXED, BASE_FIXED];
    const cn = [BASE_FIXED, y];
    const xn = [BASE_FIXED, x];
    let sum = xn[0]*cn[0] + xn[1]*cn[1];
    for (let i = 2; i < iterations; i++) {
        const v = y - bn[i-1];
        bn.push(bn[i-1] + BASE_FIXED);
        xn.push(Math.floor((xn[i-1] * x) / BASE_FIXED));
        cn.push(Math.floor((cn[i-1] * v) / bn[i]));
        sum += xn[i]*cn[i];
    }
    return Math.floor(sum / BASE_FIXED);
*/
```
 It looks promising if BASE_FIXED == 1, but here I think BASE_FIXED == 10**18, is the result still valid?
Thanks!

Brechtpd:
> Should we separate the Spot trade and AMM trade? so we have 2 different types of trade tx, and each manages its trading and fee separately. I thought the current implementation added more complexity to SpotTrade.

Very little added complexity I feel, and because both need to do very similar things (some unneeded work for AMM orders, but nothing that really gets in the way) it's probably more complex to have separate circuits that behave slightly different (also potentially less flexible if the order types don't match and an order can only be used in one of the circuits). 

> if there is a AmmTrade tx, in circuit we can fix AMM account/order to maker, and check the maxFeeBips of the opponent. So far I did not find such a check, i.e., is it possible that ammFeeBips > maxFeeBips?

Those two fees are actually very different. `feeBipsAMM` is the fee **received** by the AMM by the taker (changes the price of the curve). `maxFeeBips` is the fee paid by the account owner to the operator. AMMs currently never pay any fee to the operator, so if AMM is used the order feeBips is required to be 0 (regardless of `maxFeeBips` because the operator can choose that value freely).

> 2 more questions about the AMM price/volume calculation.
> 
>     1. What's the purpose of BASE_FIXED?? I think values like balanceIn/balanceOut already based on 10**18 unit, is it??

Not necessarily so, depends on the token.

>        Also, why do weightIn/Out need to multiply BASE_FIXED, see below 2 code snippets which come from the comments of volume and price calculation.
> 
> 
> ```c++
> //   const weightRatio = (weightIn * BASE_FIXED) / weightOut;
> ```
> 
> ```c++
> //     const numer = balanceIn * weightOut;
> //     const denom = balanceOut * weightIn;
> //     const ratio = (numer * BASE_FIXED) / denom;
> ```

Just a uniform way to do the fixed point math. Maybe not everywhere needed when additional requirements are added to the format of the weights, but definitely required because we need to be able to support values between [0,1], so we have to represent the fractional parts somehow. Easy example: `3 // 2 = 1`, which wouldn't be accurate. With the multiplication we accurately represent it as `1.5`.

>     1. Is the PowerGadget the Taylor expansion of pow(x, y)?
> 
> 
> ```c++
> /*
>     const x = (_x - BASE_FIXED);
>     const bn = [BASE_FIXED, BASE_FIXED];
>     const cn = [BASE_FIXED, y];
>     const xn = [BASE_FIXED, x];
>     let sum = xn[0]*cn[0] + xn[1]*cn[1];
>     for (let i = 2; i < iterations; i++) {
>         const v = y - bn[i-1];
>         bn.push(bn[i-1] + BASE_FIXED);
>         xn.push(Math.floor((xn[i-1] * x) / BASE_FIXED));
>         cn.push(Math.floor((cn[i-1] * v) / bn[i]));
>         sum += xn[i]*cn[i];
>     }
>     return Math.floor(sum / BASE_FIXED);
> */
> ```
> 
> It looks promising if BASE_FIXED == 1, but here I think BASE_FIXED == 10**18, is the result still valid?
> Thanks!

Remember that `x` in our case is [0, 1] so most of the math here is using fixed point math. Which doesn't matter for the formula, it's just one way to represent the numbers (if the implementation is correct of course :).

yueawang:
> Those two fees are actually very different. feeBipsAMM is the fee received by the AMM by the taker (changes the price of the curve). maxFeeBips is the fee paid by the account owner to the operator. AMMs currently never pay any fee to the operator, so if AMM is used the order feeBips is required to be 0 (regardless of maxFeeBips because the operator can choose that value freely).

I see that. I just thought from a user's perspective, one places an order with the chosen maxFeeBips without knowing AMM or another trader takes the fee, but now I think ammFeeBips can be greater than maxFeeBips, it seems an unexpected result to the user. Though we could rely on the operator to avoid this case, should the circuit also add such constraint? Or it's already in there I might overlook. :)

Also I found that AMM can be the taker in current design, and trading could happen between 2 AMMs, is that necessary? I think those AMMs in L1 are all makers waiting for the takers coming, should we fix our AMM to the maker in the circuit? The operator can do that of course, just to see if it simplify the circuit implementation.

Thanks!

Brechtpd:
> > Those two fees are actually very different. feeBipsAMM is the fee received by the AMM by the taker (changes the price of the curve). maxFeeBips is the fee paid by the account owner to the operator. AMMs currently never pay any fee to the operator, so if AMM is used the order feeBips is required to be 0 (regardless of maxFeeBips because the operator can choose that value freely).
> 
> I see that. I just thought from a user's perspective, one places an order with the chosen maxFeeBips without knowing AMM or another trader takes the fee, but now I think ammFeeBips can be greater than maxFeeBips, it seems an unexpected result to the user. Though we could rely on the operator to avoid this case, should the circuit also add such constraint? Or it's already in there I might overlook. :)

Because fees both mean different things, I don't think it's really confusing for users because the user doesn't pay a higher fee than expected even if `ammFeeBips > maxFeeBips`. Because `ammFeeBips` directly impacts the price, it's just like the user just trading with a limit order against an order with a different price. `ammFeeBips` is just a way to increase the price above the one required by the curve. Not that that `maxFeeBips` is still used by non-AMM orders to pay the operator. So if a user trades with an AMM order 2 fee are paid
- `ammFeeBips`: but this fee is paid as part of the trading amounts itself, the user does not know or even care what  `ammFeeBips` is for an AMM, the user just cares what the resulting price is. And because the taker has put his maximum price as part of the limit order, the user is okay with whatever the resulting price is (which includes the fee).
- `feeBips`: ``feeBips` < `maxFeeBips`, the fee paid to the operator as a separate transfer.

> 
> Also I found that AMM can be the taker in current design, and trading could happen between 2 AMMs, is that necessary? I think those AMMs in L1 are all makers waiting for the takers coming, should we fix our AMM to the maker in the circuit? The operator can do that of course, just to see if it simplifies the circuit implementation.

Would be fine with me. Would save some contraints (I think around ~5,000 currently) so if we don't think it's necessary we should definitely remove it.

> from the comment, I think here y (input x of p) is in [0, 1] * BASE_FIXED, not [0,1].

Yep, but  [0, 1] * BASE_FIXED is defined as [0, 1] with fixed point math when using 18 decimals. No floats in circuits. :)

> And also I just can not figure out what's does y ** (weightRatio) mean, especially the BASE_FIXED factor of weightRatio. As weightRatio is exponent, I guess the final result should go through something like pow(res, 1/BASE_FIXED), but I did not find such process.
> Do you have a reference code to show the calculation?

You can refer to https://github.com/Loopring/team/issues/126#issuecomment-676786954 where I did some experimentations. Seems to work out and multiplications/divisions should be done where needed because of the fixed point math if I didn't make any mistakes.

PR #1676, opened 2020-08-31
[hebao 1.1] fix fee calculate issue

dong77:


kongliangzhong (approved):

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -171,15 +172,38 @@ abstract contract ForwarderModule is BaseModule
                 metaTx.to == controller().walletFactory()
             );
 
+            // MAX_REIMBURSTMENT_OVERHEAD covers an ERC20 transfer and a quota update.
+            if (skipQuota) {
+                gasUsed -= 48000;
+            }
+
+            if (metaTx.gasToken == address(0)) {
+                gasUsed -= 15000; // diff between an regular ERC20 transfer and an ETH send
```
Brechtpd, line 181: These gas calculations depend on a lot of assumptions. Wouldn't it be easier just to let the user sign a fixed gas overhead cost which we could accurately calculate offchain so we don't have to do try and calculate this onchain?

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -141,23 +144,21 @@ abstract contract ForwarderModule is BaseModule
...
+
+        uint gasUsed = gasLeft - gasleft() +
+            (signature.length + metaTx.data.length + 7) * 16 + // data input cost
+            447 + // cost of MetaTxExecuted = 375 + 9 * 8
+            21000; // transaction cost;
+
+        if (needReimburst) {
+            gasUsed = gasUsed +
+                MAX_REIMBURSTMENT_OVERHEAD + // near-worst case cost
+                1500; // 3 * 500 for three SLOAD for calculating skipQuota below
```
Brechtpd, line 161: SLOAD cost should be [800 gas](https://github.com/sorpaas/EIPs/blob/c88b73889e66669db0e7094dda7907661f54d873/EIPS/eip-2200.md#specification) I think.  If this is for `controller().walletFactory()` I think the cost is `2*SLOAD+1*CALL = 2*800+1*700=2300` but I don't really know how accurate that will be either.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dong77: I like the idea.

Brechtpd (approved):

kongliangzhong (approved):

Issue #1677, opened 2020-09-01
[Loopring 3.6] Implement a Managing Contract for Loopirng Layer-2 AMM Liquidity Pool


dong77:
In this issue, I'd like to outline a few design objectives of the layer-1 managing contract for our AMM liquidity pool accounts on  L2. These objectives are based on a discussion in the Shanghai office and my updated understanding of the new AMM support in our circuits.

## Objectives

1.  Allow users to add liquidity to and remove liquidity from the managing contract on L1. ERC20-based liquidity tokens can be mint or burnt. The liquidity tokens may be transferrable and can be deposited to L2 if the token is registered.

2.  Allow regular L2 accounts to add liquidity to the AMM account by using L2 transfers. On L1, the correct bookkeeping should be done or liquidity token should be mint to the owner. These add-liquidity transfers' data must be delivered to the AMM managing contract before the corresponding block is submitted. *As of now the operator's behaviors are not guaranteed by the protocol and is more of a centralized trust on the relayer [see the possible solution below]. If the relayer doesn't notify the managing contract of these add-liquidation transfers, then the user will not get his share of the pool, and his fund will be lost.*

> One possible solution is to make the managing contract enforce an opening AmmUpdate tx and a closing AmmUpdate tx, and all non-trading AMM account balance update txs must be wrapped by these AmmUpdate tx pairs (no other tx can be put in between these AmmUpdate pairs of course). By doing this, the managing contract will not miss any liquidation addition and removal no matter the requests are on L1 or L2. (note, this doesn't seem to require us to change the current implementation)
>
> If the operator doesn't put all transfer-based, L2 liquidation addition/removal txs between the pair, there will be data inconsistency in the AmmUpdate pair because all those txs must add up.

3. Allow L2 accounts to remove liquidity from AMM account back to the user's L2 account via a transfer. The transfer will be created by the relayer without any signature, and it will be validated by the managing contract on L1 to create a hash-based approval. If the operator refuses the request, the L2 account may choose to interact with the managing contract to force a withdrawal, this should be very similar or even exactly as point 1.

[...]

PR #1683, opened 2020-09-02
[Loopring_v3] Two step deposit

dong77:
Users can use a dapp (AMM pool for example) to deposit funds into Loopring Exchange for custody, then later, the dapp can use these funds directly for deposits without another ERC20 transfer. 

Another benefit is that once a dapp becomes a global agent, it can use a user's existing ERC20 authorization.

### DIFF
--- a/packages/loopring_v3/contracts/core/iface/ExchangeData.sol
+++ b/packages/loopring_v3/contracts/core/iface/ExchangeData.sol
@@ -202,6 +202,8 @@ library ExchangeData
         // A map from an account owner to a destination address to a tokenID to an amount to a nonce to a new recipient address
         mapping (address => mapping (address => mapping (uint16 => mapping (uint => mapping (uint32 => address))))) withdrawalRecipient;
 
+        // A map from apps to users to tokens to their balances
+        mapping (address => mapping(address => mapping(address => uint))) custody;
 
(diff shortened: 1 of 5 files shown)


Brechtpd:
I think this works, potentially lowering the deposits costs from ~60,000 gas to ~45,000 (no token transfer + no call to the deposit contract), I guess for ETH deposits the SSTORE is potentially more expensive, but we still save a call to the deposit contract so probably will cost largely the same.

f the pool contract needs to process a couple of joins/exits than it will need to update storage/etc... to keep track of all the balances/minting liquidity tokens/.... which will be expensive. And then only a max of `n` (with `n` the number of pool tokens in the pool) deposits to do, so percentage wise probably not that much gas saved in total in most cases. 

So some gas savings possible, with some extra cases in the protocol we need to support but easy ones...At this point I'm undecided. :)

dong77:
> ...so percentage wise probably not that much gas saved in total in most cases.

I agree. Let's keep the PR pending here for a while. I'm also undecided. 

But what you said above makes me feel like to add a fixed join/exit processing fee (in ETH) in the AMM contract. The fee will be charged by the operator from each liquidity provider. Let's discuss it here https://github.com/Loopring/protocols/issues/1677

Issue #1688, opened 2020-09-03
Change feebips in transfer/withdraw(off-chain) to max-feebips


hosschao:
Just like what we did in order request. Max feebips makes an opportunity for relayer to decided exactly feebips based on some infos (like VIP level).

dong77:
Brecht is working on changing the withdrawal fee as the max fee a user would like to pay, the actual fee will be determined by the relayer.

There may be future use cases to change the transfer to also have a max fee and allow a discount by the relayer. But I don't know what's the extra cost on-chain, and whether it will make all transactions more expensive.

dong77:
So for EdDSA signed offchain transfers and withdrawals, we know support maxFee (meaning fee <= maxFee), and for transfers and withdrawals authorized in other means, we will always use maxFee (does the relayer need to provide a fee value but it's just not used?)

Brechtpd:
When not using an EdDSA signature the relayer can fill in any `maxFee`, but that's okay because the user needs to sign/approve the actual fee amount. So the relayer can fill in any `maxFee >= fee` so the circuit is valid, but it has no real impact.

dong77:
> When not using an EdDSA signature the relayer can fill in any `maxFee`, but that's okay because the user needs to sign/approve the actual fee amount. So the relayer can fill in any `maxFee >= fee` so the circuit is valid, but it has no real impact.

Haha, I see the point.

PR #1689, opened 2020-09-03
[protocol 3.6] Add maxFee to withdrawals/transfers

Brechtpd:
#1688

Currently only supports `fee != maxFee` when using an EdDSA signature. I can add support for it with onchain signatures as well if requested. The reason for this is that only the actual fee is part of the DA, but if `fee != maxFee` we need to calculate the hash for the signature with `maxFee` instead of `fee`, so we need to put the `maxFee` onchain (using the auxiliary data) and check that `fee < maxFee` onchain.

### DIFF
--- a/packages/loopring_v3/circuit/Circuits/TransferCircuit.h
+++ b/packages/loopring_v3/circuit/Circuits/TransferCircuit.h
@@ -53,6 +53,7 @@ class TransferCircuit : public BaseTransactionCircuit
     DualVariableGadget payer_toAccountID;
     DualVariableGadget payer_to;
     DualVariableGadget payee_toAccountID;
+    DualVariableGadget maxFee;
 
     // Check if the inputs are valid
     EqualGadget isTransferTx;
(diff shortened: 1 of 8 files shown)


dong77 (approved):

dong77:
> #1688
> 
> Currently only supports `fee != maxFee` when using an EdDSA signature. I can add support for it with onchain signatures as well if requested. The reason for this is that only the actual fee is part of the DA, but if `fee != maxFee` we need to calculate the hash for the signature with `maxFee` instead of `fee`, so we need to put the `maxFee` onchain (using the auxiliary data) and check that `fee < maxFee` onchain.

Not sure I understand it. can `fee > maxFee` be true in the case of EdDSA?

Brechtpd:
> > #1688
> > Currently only supports `fee != maxFee` when using an EdDSA signature. I can add support for it with onchain signatures as well if requested. The reason for this is that only the actual fee is part of the DA, but if `fee != maxFee` we need to calculate the hash for the signature with `maxFee` instead of `fee`, so we need to put the `maxFee` onchain (using the auxiliary data) and check that `fee < maxFee` onchain.
> 
> Not sure I understand it. can `fee > maxFee` be true in the case of EdDSA?

The user just always needs to sign the actual fee that is paid. There isn't any knowledge of `maxFee` onchain, only the real fee paid.

dong77 (approved):

PR #1691, opened 2020-09-04
[Hebao 1.1] fix bug in WalletFactory

dong77:


packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
```diff
@@ -34,6 +34,13 @@ abstract contract MetaTxAware
         _;
     }
 
+    function initTrustedForwarder(address _trustedForwarder)
+        public
+    {
+        require(trustedForwarder == address(0), "INITIALIZED_ALREADY");
```
dong77, line 40: move this to WalletFactory

Issue #1692, opened 2020-09-04
[Loopring 3.6] Improve fast withdrawl handling logic


dong77:
When a fast withdrawal token transfer transaction and a block submission transaction (that includes a corresponding offchain withdrawal tx) are both pending. Liquidity provider may end up losing money, if the block submission tx is confirmed first.

We can supply an extra data in aux data for each withdrawal transaction to indicate if a fast withdrawal transfer is pending on layer1, if so, inside the processing logic of submitBlocks, we set the redirect target (recipient) to a special value, such as 0x1. This value indicates to the token transfer transaction that it should fail. Then we can guarantee the liquidy provider will never lose money (except the failed transfer fees), and in the meanwhile, the relayer will not need to monitor those transfer transactions from the liquidity provider.

@Brechtpd please let me know what you think?

Another two features that Kongliang is implementing in the fast withdrawal contract:

1. Allow the liquidity provider to authorize a fast withdrawal token transfer with a signature. With this approval, anybody can create the transfer tx as msg.sender, but each approval can only be used once. A user has the option to speed up his incoming fast withdrawal transfer tx by using the authorization in a pending tx.

2. A fast withdrawal transfer tx will have an expiry. The liquidity provider knows when such transactions will fail with confidence. A fast withdrawal transfer that gets confirmed too late doesn't' make sense.

dong77:
@Brechtpd  @kongliangzhong as we have communicated, Brecht will work on the first part to allow the operator to hint selected offchain withdrawal txs (requests) so for those withdrawals, the fast withdrawal token transfer from the liquidity provider will fail gracefully with as little cost as possible.

Kongliang will work on the two features to allow people to accelerate pending incoming fast withdrawal token/ether transfers on layer1 and to allow the transfer to expire gracefully.

dong77:
Also see https://github.com/Loopring/protocols/pull/1701 and https://github.com/Loopring/protocols/pull/1706

Issue #1238, opened 2020-06-02
[Loopring 3.6] use storage for offchain withdrawals


dong77:
This will allow the operator to drop certain requests in the middle of the queue and still be able to use/handle requests with larger nonces from the same user. Otherwise, the relayer will have to invalidate all blocks that contains larger nonces from the same user.

This will cause a small problem in a fully decentralized operator environment. With the current nonce solution (+1 strictly), operator A cannot process a user's request with nonce 3 if he hasn't seen a request from the same user with nonce 2, so he has to hold request with nonce 3. Now he will be more aggressive and will process the request which may invalid a request with nonce2 the user send to operator B. But I think this is a small problem as most users will not use two operators simutanously.

Brechtpd:
So we want this even if all requests that modify the nonce are handled by the same circuit?

I'm not totally completely sure when this can happen if we provide instant apparent finality. Transactions that use nonces are executed immediately after the tx is signed by the user, no? (as in the internal relayer state is modified and the tx is added to a block) How can a tx suddenly become invalid after the user signed a different tx? Doesn't this mean it was once valid and added to a block, but then removed for some reason and the state changes it caused are reverted in the internal state?

dong77:
Hmm.. I auctually don't have a strong preference. I think +n will probably offer the relayer some flexibility, but I cannot convince myself that reason is strong enough.

dong77:
After talking to @hosschao and @lydy, the relayer team will drop this feature request.

hosschao:
Hi brecht, we hope removing nonce for withdraw request. (use storage id) So that we have ability to fail some user's offchain requests. (Or user may want to cancel that.) Is it easy to achieve that?

dong77:
I like this idea as it will give relayer more flexibility to handle withdrawals of large values first...

dong77:
@hosschao I talked about this request with Brecht and it seems we can do it. Brecht will figure out whether there is an impact on the size of DA.

PR #1703, opened 2020-09-05
[protocol 3.6] Use storageIDs for withdrawals + Store recipient option

Brechtpd:
#1238
#1692

If a recipient is stored for a withdrawal, it's currently not possible to ever set the value back to 0 to get the gas refund. This is so that we can always depend on it as replay protection. 

We could store some extra timestamp (in the same 32 bytes so no extra cost there), like the signature `validUntil` value, and then allowi the data to be reset back to 0 when e.g. that timestamp > now, so that we are sure the value is there for replay protection as long as needed. But then it acts more like a gas token, so will have to have logic to reset that specific value when doing other gas expensive work onchain. This will always have some overhead, so in the best case we could probably still get a ~10,000 gas refund this way when this would be possible, but the relayer will have to keep track of all of this and then call the correct function with the correct data to get the refund when it makes sense.


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -554,17 +554,34 @@ contract ExchangeV3 is IExchangeV3
...
-        uint32  nonce,
+        uint32  storageID,
         address newRecipient
         )
         external
         override
         nonReentrant
         onlyFromUserOrAgent(from)
     {
+        require(newRecipient != address(0), "INVALID_DATA");
```
dong77, line 565: should we allow the recipient to be set to 0? I'm thinking of the following scenario:
- The withdrawal is done, the recipient is set to `withdrawal.to`
- The FastWithdrawalContract reads the recipient, find it's non-zero, then set the new value to 0, to get some gas refund, then abord the operation (but still mark the approval as used)

Brechtpd: > (but still mark the approval as used)

I guess this is the problem. Unless time (or something similar) is used somehow, somewhere we need to store that the fast withdrawal approval cannot be used any more. With the current method we only need to depend on the stored recipient for that (if non-zero it cannot be used), no other hash/nonce needs to be stored elsewhere. Otherwise we reset storage to zero in one place, but need to store a non-zero value elsewhere, which makes things more expensive.

Brechtpd: Also the recipient can be set from the account owner himself, or any of it's agents. So it's important that it cannot be overridden because the owner can directly do that himself or just register another agent contract that can do it as well. Something to keep in mind as well. Which is why in the opening post I stored the "store until" timestamp in the protocol itself, because we cannot depend on the logic inside a single agent to keep it safe.

We could make it so that the data with the address that set it, is also the only that address can reset it again, but that makes things a bit more complex again (and will probably also make it so we need to use 2 storage slots inside of the 1 we need now to store that extra address).

dong77: I was thinking the FastWithdrawal contract will also do a bookkeeping. If our solution can make it unnecessary, that's even better.

Brechtpd: Currently none is needed, no storage is currently used for fast withdrawals in the FastWithdrawalAgent contract.

dong77 (approved):

dong77 (approved):

Issue #1696, opened 2020-09-04
[Loopring 3.6] Use weight as a gate to control AMM pool transfers on layer2.


dong77:
The idea is to disable incoming token transfers to an AMM pool account if the _weight_ of the token is non-zero. This is not the case yet in the current implementation.

Switching a token's weight between zero and non-zero will form a window such that the AMM managing contract can check the preconditions and postconditions regarding a sequence of account/balance update transactions.

PR #1705, opened 2020-09-05
[protocol 3.6] Disable transfers to active AMM tokens

Brechtpd:
#1696


dong77 (approved):

dong77:
Shall we also disable deposit and withdrawal as well? It seems to make sense to disable all non-trading balance change for tokens with non-zero weights.

dong77:
After a forced withdrawal, the token's weight should be set to 0. Not sure if this is already the case.

Brechtpd:
> Shall we also disable deposit and withdrawal, or even all transactions except trades/swaps for non-zero-weight tokens?

Unless there's a situation where something can be attacked, I would lean to just keep things with as little special cases as possible.

> After a forced withdrawal, the token's weight should be set to 0. Not sure if this is already the case.

This is already done.

dong77 (approved):

PR #1706, opened 2020-09-06
[Loopring 3.6] Fast Withdrawal Alternative Impl

dong77:
Based on https://github.com/Loopring/protocols/pull/1703 and https://github.com/Loopring/protocols/pull/1701


packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
```diff
@@ -52,112 +58,148 @@ contract FastWithdrawalAgent is ReentrancyGuard
...
-    // EIP712
-    bytes32 constant public FASTWITHDRAWAL_TYPEHASH = keccak256(
-        "FastWithdrawal(address exchange,address from,address to,address token,uint96 amount,address feeToken,uint96 fee,uint32 nonce,uint32 validUntil)"
+    bytes32 constant public FASTWITHDRAWAL_APPROVAL_TYPEHASH = keccak256(
+        "FastWithdrawalApproval(address exchange,address from,address to,address token,uint96 amount,uint32 storageID,int256 validUntil)"
     );
+
     bytes32 public DOMAIN_SEPARATOR;
 
+    mapping(bytes32 => bool) consumedApprovals;
```
Brechtpd, line 89: Doesn't seem necessary currently and makes things quite a bit more expensive.

My idea was to simply do this:
- Try/Catch `setWithdrawalRecipient`. In normal cases this will always succeed, only when the block already included the withdrawal does this fail. It also fails when trying to use the request more than once.
- Do the transfer if successful, don't do the transfer if not successful.

And that's it. Easy and efficient.

dong77: We need to handle a replay issue. If a fast withdrawal approval is used and this tx is successful, after it is consumed by the protocol (which deletes the redirect), this same approval can be used again.

dong77: @kongliangzhong is modifying this PR. redirect recipient will not be deleted any more, then we can remove consumedApprovals.

Brechtpd: > We need to handle a replay issue. If a fast withdrawal approval is used and this tx is successful, after it is consumed by the protocol (which deletes the redirect), this same approval can be used again.

That's why the protocol won't delete the redirect any more, so we can depend on it being there. 

> is modifying this PR. redirect recipient will not be deleted any more, then we can remove consumedApprovals.

Was already the case?

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -167,11 +168,6 @@ library WithdrawTransaction
             withdrawal.to = recipient;
             // Allow any amount of gas to be used on this withdrawal (which allows the transfer to be skipped)
             withdrawal.minGas = 0;
-
-            // Delete the recipient if we don't have to store it to get the gas refund
-            if (!auxData.storeRecipient) {
-                delete S.withdrawalRecipient[withdrawal.owner][withdrawal.to][withdrawal.tokenID][withdrawal.amount][withdrawal.storageID];
-            }
```
Brechtpd, line 174: Why was this deleted?

packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
```diff
@@ -12,152 +12,180 @@ import "../../lib/ReentrancyGuard.sol";
...
+
+        IExchangeV3 exchange = IExchangeV3(fwa.exchange);
+
+        address recipient = exchange.getWithdrawalRecipient(
+            fwa.from,
+            fwa.to,
+            fwa.token,
+            fwa.amount,
+            fwa.storageID
+        );
```
Brechtpd, line 165: Much more efficient to do try/catch on `setWithdrawalRecipient` instead. Saves several calls and storage reads.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
kongliangzhong: If receipient is deleted, then the FastWithdrawalApproval can be used again.

Brechtpd: That's why it's an option the relayer can set so we only lose efficiency when we actually need it. So the assumption here is that `withdrawalRecipient` will only be used in cases that need this permanent storage? `storeRecipient` will have to be `true` for fast withdrawals either way so no difference there, but okay I guess if we don't mind the extra cost in other use cases that don't need this.

dong77: Keeping the hash will demand more trust from the LPs which is not a good thing.

packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
kongliangzhong: Ok, fixed.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: If we can provide LPs with a withdrawal receipt I guess we can allow LPs not to trust the operator. But  because the operator has the have funds locked up for these receipts it's kind of just circular thing (lock up funds so no funds have to be used, so why not use the funds directly), so I'm not sure this is really something we need to support.

> If the relayer is dishonest/evil, he will delete the has, and since existing approvals are permanently onchian, they can be used to drain tokens from LPs.

I thought about making `storeRecipient` part of data signed by the user (as part of the data hashed to `onchainDataHash`) so it can't be chosen by the relayer and so this isn't a problem, but chose against it because I still think trusting the operator is inevitable and the only LP that wants to do fast withdrawals is the operator himself.

dong77: A temporary trust should be fine if the LP is not the operator, but if we delete a used recipient, then the LP must trust the relayer forever as any used approvals on the blockchain may be re-used, this will be a big surprise to the LP.

Brechtpd: - It will be impossible to ever do another withdrawal that will map to the same storage slot ever again. So if the the withdrawal has been included in a block, that's the end of it, the operator/user can never create another withdrawal like it, so if the storage hasn't been reset and the withdrawal has been processed, it can never be changed again.
- I'm sure that any signed approval data will always contain some `validUntil` that limits its use in time.

If we do think this is a use case we want to support, then making `storeRecipient` part of the withdrawal request like described above solves it without any assumptions.

kongliangzhong (dismissed):

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
dong77: > It will be impossible to ever do another withdrawal that will map to the same storage slot ever again. So if the the withdrawal has been included in a block, that's the end of it, the operator/user can never create another withdrawal like it, so if the storage hasn't been reset and the withdrawal has been processed, it can never be changed again.

My point is, the LP's money can still be transferred to the previous user again by arbitrary users with a used approval (before its expiry), the storage slot never matters as the exchange is not involved at all.

kongliangzhong (dismissed):

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
Brechtpd: Oh sure, approvals could be used twice when it's reset. But the fact still remains that the LP has no guarantee the operator will even use the withdrawal request for the user. The operator/user can easily create another withdrawal request on a different storage slot (or with slightly different values) and the withdrawal request the LP did the fast withdrawal for simply will never be used and the withdrawal for the user can still continue as usual, with the LP losing all his funds. So if the operator wants to steal the funds of the LP there's no need to depend on this.

So basically my point is that the LP needs to trust the operator 100% already, being able to reset the storage so the approval can be used again does not add additional trust requirements (at least without receipts).

PR #1707, opened 2020-09-08
[protocol36]FastWithdrawal impl

kongliangzhong:


packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
```diff
@@ -12,152 +12,168 @@ import "../../lib/ReentrancyGuard.sol";
...
+                fwas[i].to,
+                fwas[i].token,
+                fwas[i].amount,
+                fwas[i].provider,
+                msg.sender,
+                status
+            );
         }
-        // Return any ETH left into this contract
-        // (can happen when more ETH is sent than needed for the fast withdrawals)
```
Brechtpd, line 92: ETH needs to be sent into the contract by the LP itself, and ETH can be stored in this contract, so it seems we currently can't let users do the fast withdrawal themselves when withdrawing ETH.

I still like the idea of msg.sender still always be the liquidity provider in the agent contract like it was before.

> We require the fast withdrawals to be executed by the liquidity provider (as msg.sender) so that the liquidity provider can impose its own rules on how its funds are spent.

This solves 2 issues:
- The ETH problem
- Much more flexible fund security for LPs. In this PR LPs are are forced to depend on a single signature of their hot wallet. Obviously that's not great when storing lots of funds there.

This still fully supports letting people submit their own fast withdrawals, it's just something the LP contract calling into the agent contract needs to support. No signatures would be checked directly in the agent contract (like before this PR).

dong77: Brecht, could you show your implementation please?

Brechtpd: #1715

PR #1714, opened 2020-09-08
[hebao] Optimized memory copies

Brechtpd:
Optimized most frequent paths. In normal cases this only saves ~0.5%-1.0% of gas. Calls with a lot of data will save more gas, but probably quite infrequent.

I do feel like the compiler should be able to do this so we don't have to worry about this. Maybe I'll make an issue on the solidity github.

### DIFF
--- a/packages/hebao_v1/contracts/modules/base/BaseModule.sol
+++ b/packages/hebao_v1/contracts/modules/base/BaseModule.sol
@@ -129,7 +129,15 @@ abstract contract BaseModule is ReentrancyGuard, Module
         internal
         returns (bytes memory)
     {
-        return Wallet(wallet).transact(uint8(1), to, value, data);
+        bytes memory callData = abi.encodeWithSelector(
+            Wallet.transact.selector,
+            uint8(1),
+            to,
+            value,
+            data
+        );
+        bytes memory ret = wallet.fastCallAndVerify(gasleft(), 0, callData);
+        return abi.decode(ret, (bytes));
     }
 
     // Special case for transactCall to support transfers on "bad" ERC20 tokens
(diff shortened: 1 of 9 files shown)


packages/hebao_v1/contracts/modules/base/BaseModule.sol
```diff
@@ -129,7 +129,15 @@ abstract contract BaseModule is ReentrancyGuard, Module
         internal
         returns (bytes memory)
     {
-        return Wallet(wallet).transact(uint8(1), to, value, data);
+        bytes memory callData = abi.encodeWithSelector(
```
dong77, line 132: Shall we also use `fastCallAndVerify` in `transactDelegateCall` and `transactStaticCall`>

dong77: Or should we change `memory` to `calldata` as in `transactDelegateCall` without the function's body change?

dong77: Sometimes `transactCall` is used with a `calldata` data from the caller; sometimes the data is assembled in memory.

dong77:
If the saving of gas is only ~0.5%-1.0%, we probably want to skip it :)

Brechtpd:
> If the saving of gas is only ~0.5%-1.0%, we probably want to skip it 

I agree. We can keep the changes to `calldata` because why not, but the replacements for the calls isn't worth it for the smart wallet use cases where the data isn't very large.

dong77:
> > If the saving of gas is only ~0.5%-1.0%, we probably want to skip it
> 
> I agree. We can keep the changes to `calldata` because why not, but the replacements for the calls isn't worth it for the smart wallet use cases where the data isn't very large.

Great, let's do that.

Brechtpd:
Removed use of the new calls in the smart wallet, but left the functions in the library because they are actually useful in other scenario's.

dong77 (approved):

PR #1715, opened 2020-09-08
[protocol 3.6] Alternative fast withdrawal implementation

Brechtpd:
Mostly just moves things in #1707 around a bit.

- Signatures only required when the user does the fast withdrawal himself.
- ETH supported.
- Liquidity providers can implement their own security, but likely could even be a normal smart wallet (probably even for user initiated fast withdrawals because the LP would just provide a signed meta tx to the user that doesn't reimburse the gas cost). The agent contract makes sure the transfers are valid so the user doesn't need to care how it's implemented. Simple bare bone liquidity provider contract implemented with just a single owner.

### DIFF
--- a/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
+++ b/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
@@ -11,9 +11,8 @@ import "../../lib/MathUint.sol";
 import "../../lib/ReentrancyGuard.sol";
 import "../../lib/SignatureUtil.sol";
 
-
-/// @title Fast withdrawal agent implementation. The fast withdrawal request reduces to
-///        a normal onchain withdrawal request after a specified time limit has exceeded.
+/// @title Fast withdrawal agent implementation. With the help of liquidity providers (LPs),
+///        exchange operators can convert any normal withdrawals into fast withdrawals.
 ///
 ///        Fast withdrawals are a way for the owner to provide instant withdrawals for
 ///        users with the help of a liquidity provider and conditional transfers.
@@ -21,14 +20,9 @@ import "../../lib/SignatureUtil.sol";
 ///        A fast withdrawal requires the non-trustless cooperation of 2 parties:
 ///        - A liquidity provider which provides funds to users immediately onchain
 ///        - The operator which will make sure the user has sufficient funds offchain
-///          so that the liquidity provider can be paid back offchain using a conditional transfer.
-///          The operator also needs to process those conditional transfers so that the
-///          liquidity provider receives its funds back in its own account where it
-///          again has full custody over it.
-///
-///        However, there is a special case when the fast withdrawal reduces to a standard
-///        withdrawal and the fee is paid onchain. In this case the withdrawal can be
-///        done completely trustless, no cooperation with the owner is needed.
+///          so that the liquidity provider can be paid back.
+///          The operator also needs to process those withdrawals so that the
+///          liquidity provider receives its funds back.
 ///
 ///        We require the fast withdrawals to be executed by the liquidity provider (as msg.sender)
 ///        so that the liquidity provider can impose its own rules on how its funds are spent. This will
@@ -44,50 +38,58 @@ import "../../lib/SignatureUtil.sol";
 ///        authorize this contract as their agent.
 ///
 /// @author Brecht Devos - <brecht@loopring.org>
+/// @author Kongliang Zhong - <kongliang@loopring.org>
+/// @author Daniel Wang - <daniel@loopring.org>
 contract FastWithdrawalAgent is ReentrancyGuard
 {
     using AddressUtil       for address;
     using AddressUtil       for address payable;
     using ERC20SafeTransfer for address;
     using MathUint          for uint;
-    using SignatureUtil     for bytes32;
 
-    struct FastWithdrawal
+    event Processed(
+        address exchange,
+        address from,
+        address to,
+        address token,
+        uint96  amount,
+        address provider,
+        bool    success
+    );
+
+    struct Withdrawal
     {
         address exchange;
         address from;                   // The owner of the account
-        address to;                     // The address that will receive the tokens withdrawn
+        address to;                     // The `to` address of the withdrawal
         address token;
         uint96  amount;
-        address feeToken;
-        uint96  fee;
-        uint32  nonce;
-        uint32  validUntil;
-
-        bytes   signature;
-    }
-
-    // EIP712
-    bytes32 constant public FASTWITHDRAWAL_TYPEHASH = keccak256(
-        "FastWithdrawal(address exchange,address from,address to,address token,uint96 amount,address feeToken,uint96 fee,uint32 nonce,uint32 validUntil)"
-    );
-    bytes32 public DOMAIN_SEPARATOR;
-
-    constructor()
-    {
-        DOMAIN_SEPARATOR = EIP712.hash(EIP712.Domain("FastWithdrawalAgent", "1.0", address(this)));
+        uint32  storageID;
     }
 
     // This method needs to be called by any liquidity provider
-    function executeFastWithdrawals(FastWithdrawal[] memory fastWithdrawals)
+    function execute(Withdrawal[] calldata withdrawals)
         public
         nonReentrant
         payable
     {
         // Do all fast withdrawals
-        for (uint i = 0; i < fastWithdrawals.length; i++) {
-            executeFastWithdrawal(fastWithdrawals[i]);
+        for (uint i = 0; i < withdrawals.length; i++) {
+            executeInternal(withdrawals[i]);
         }
+        // Return any ETH left into this contract
+        // (can happen when more ETH is sent than needed for the fast withdrawals)
+        msg.sender.sendETHAndVerify(address(this).balance, gasleft());
+    }
+
+    // This method needs to be called by any liquidity provider
+    function execute(Withdrawal calldata withdrawal)
+        public
+        nonReentrant
+        payable
+    {
+        // Do the fast withdrawal
+        executeInternal(withdrawal);
         // Return any ETH left into this contract
         // (can happen when more ETH is sent than needed for the fast withdrawals)
         msg.sender.sendETHAndVerify(address(this).balance, gasleft());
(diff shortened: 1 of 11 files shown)


packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+            } else {
+                token.safeTransferAndVerify(to, amount);  // ERC20 token
+            }
+        }
+    }
+
+    // Allows the LP to enable the necessary ERC20 approvals
+    function approve(
+        address token,
+        address spender,
```
dong77, line 120: is the spender usually the exchange contract?

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+    FastWithdrawalAgent public agent;
+
+    constructor(FastWithdrawalAgent _agent)
+    {
+        agent = _agent;
+        DOMAIN_SEPARATOR = EIP712.hash(EIP712.Domain("BasicLiquidityProvider", "1.0", address(this)));
+    }
+
+    // User initiated fast withdrawal
+    function execute(FastWithdrawalApproval calldata approval)
```
dong77, line 53: I suggest we change this to `function execute1(FastWithdrawalApproval[] calldata approvals)` and comment that *this function can be used by LPs to perform fast withdrawals which can be accelerated by users with a higher gas price using the same function*.

Also, change the other function to something like `execute2` so tools can be made to monitor pending fast withrawal transactions simply using the function selector.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+        external
+        nonReentrant
+    {
+        // Compute the hash
+        bytes32 hash = EIP712.hashPacked(
```
dong77, line 58: extract a validateApproval function?

packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent.sol
```diff
@@ -44,119 +38,110 @@ import "../../lib/SignatureUtil.sol";
...
+            executeInternal(withdrawals[i]);
+        }
+        // Return any ETH left into this contract
+        // (can happen when more ETH is sent than needed for the fast withdrawals)
+        msg.sender.sendETHAndVerify(address(this).balance, gasleft());
     }
 
     // This method needs to be called by any liquidity provider
-    function executeFastWithdrawals(FastWithdrawal[] memory fastWithdrawals)
+    function execute(Withdrawal calldata withdrawal)
```
dong77, line 86: we may want to remove this method, as in FastWithdrawalProvider, only `execute(Withdrawal[] calldata withdrawals)` is used.

dong77: maybe also change the name to something other than `execute` to differentiate it from the function in FastWithdrawalProvider.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+import "../../lib/ERC20.sol";
+import "../../lib/ERC20SafeTransfer.sol";
+import "../../lib/MathUint.sol";
+import "../../lib/ReentrancyGuard.sol";
+import "../../lib/SignatureUtil.sol";
+
+
+/// @title Basic contract storing funds for a liquidity provider.
+/// @author Brecht Devos - <brecht@loopring.org>
+contract BasicLiquidityProvider is ReentrancyGuard, Claimable
```
dong77, line 18: I suggest we use `OwnerManagable` so we can add more addresses for signing.

dong77 (changes requested):

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
Brechtpd: Only the agent if just used for the fast withdrawals.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
Brechtpd: > this function can be used by LPs to perform fast withdrawals which can be accelerated by users with a higher gas price using the same function.

This function should only be called by the user, LPs shouldn't use this function because it will do an additional signature check that is unnecessary when the LP calls it himself.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
Brechtpd: Depends I guess on what we will use. More addresses that can authorize transactions (less secure I guess) vs needing multiple signatures to authorize a single withdrawal (owner could be smart wallet / multi sig contract).

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
dong77: I think my goal is that users or any third-parties can accelerate any pending fast withdrawal transfers. If LPs use `function execute(FastWithdrawalAgent.Withdrawal[] memory withdrawals)`, user-initiated acceleration will be impossible.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
dong77: Allowing more addresses will be more flexible, for example, if we really want to speed up an transaction, we don't have to wait for all existing pending transactions, we can simply use a different address.

packages/loopring_v3/contracts/aux/liquidity-providers/BasicLiquidityProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+    function execute(FastWithdrawalAgent.Withdrawal[] memory withdrawals)
+        public
+        nonReentrant
+        onlyOwner
+    {
+        executeInternal(withdrawals);
+    }
+
+    // Allows the LP to transfer funds back out of this contract.
+    function transferOut(
```
dong77, line 99: how about drain()?

Brechtpd:
One last question about the recipient storage (I promise ;).

https://github.com/Loopring/protocols/blob/20e7ba2eae464f2fd7e562b0ced05823de6c1bac/packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol#L160-L179

Let's say I remove lines 172-174, it'll still be an option for the relayer to set the recipient when it was not yet set (otherwise we increase gas costs of all withdrawals by 20,000 gas). So if the normal withdrawal is done before the fast withdrawal the relayer still has 100% control if the fast withdrawal can still be executed or not (and so the LP loses money).

So I would still propose to add `storeRecipient` to the withdrawal itself (as part of the data hashed to `onchainDataHash`, signed by the user), this way we can have fully trustless LPs doing fast withdrawals for users when we provide a receipt for the withdrawal to the LP. This would mean however that not all withdrawal requests are equally secure for fast withdrawals, but if `storeRecipient` if false for the withdrawal the security for LPs simply reduces to the one we currently have.

dong77 (approved):

dong77:
Brecht, I removed thse

> One last question about the recipient storage (I promise ;).
> 
> https://github.com/Loopring/protocols/blob/20e7ba2eae464f2fd7e562b0ced05823de6c1bac/packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol#L160-L179
> 
> Let's say I remove lines 172-174, it'll still be an option for the relayer to set the recipient when it was not yet set (otherwise we increase gas costs of all withdrawals by 20,000 gas). So if the normal withdrawal is done before the fast withdrawal the relayer still has 100% control if the fast withdrawal can still be executed or not (and so the LP loses money).
> 
> So I would still propose to add `storeRecipient` to the withdrawal itself (as part of the data hashed to `onchainDataHash`, signed by the user), this way we can have fully trustless LPs doing fast withdrawals for users when we provide a receipt for the withdrawal to the LP. This would mean however that not all withdrawal requests are equally secure for fast withdrawals, but if `storeRecipient` if false for the withdrawal the security for LPs simply reduces to the one we currently have.
> 
> ... or we don't care about this and I still just remove lines 172-174 even though it doesn't make that much of difference. :)

Brecht, I deleted these three lines, I feel that our design will work without any problem in most cases, and in all cases, LP and user's money are all secure. The worse case is the relayer submits blocks earlier than some fast withdrawal transfers, and relayer will have to pay extra 20K gas for the sstore.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -0,0 +1,166 @@
...
+    // withdrawal by calling , but the approvals are not validated when done by the
+    // owner or one of the managers.
+    function execute(FastWithdrawalApproval[] calldata approvals)
+        external
+        nonReentrant
+    {
+        // Prepare the data and validate the approvals when necessary
+        FastWithdrawalAgent.Withdrawal[] memory withdrawals = new FastWithdrawalAgent.Withdrawal[](approvals.length);
+        for (uint i = 0; i < approvals.length; i++) {
+            if (!isManager(msg.sender)) {
```
Brechtpd, line 64: This'll do an SLOAD at 800 gas/withdrawal instead of a single SLOAD for all withdrawals. Not a huge difference but caching it is easy enough.

Brechtpd:
> Brecht, I deleted these three lines, I feel that our design will work without any problem in most cases, and in all cases, LP and user's money are all secure. The worse case is the relayer submits blocks earlier than some fast withdrawal transfers, and relayer will have to pay extra 20K gas for the sstore.

The worst case is that the relayer submits the block earlier and doesn't set the recipient, in which case the fast withdrawal is still possible and the LP loses the funds for doing the fast withdrawal. But okay for me.

dong77:
> > Brecht, I deleted these three lines, I feel that our design will work without any problem in most cases, and in all cases, LP and user's money are all secure. The worse case is the relayer submits blocks earlier than some fast withdrawal transfers, and relayer will have to pay extra 20K gas for the sstore.
> 
> The worst case is that the relayer submits the block earlier and doesn't set the recipient, in which case the fast withdrawal is still possible and the LP loses the funds for doing the fast withdrawal. But okay for me.

Agreed. This is the part LPs need to trust the relayer.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
dong77: Good point, I'll restore it as you have implemented.

dong77: After a second thought, I want to make sure every approval is valid (with the right signature) so users wont' worry his own speeding-up the pending txs will be a waste of his own gas.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -0,0 +1,174 @@
...
+        external
+        nonReentrant
+    {
+        // Prepare the data and validate the approvals when necessary
+        FastWithdrawalAgent.Withdrawal[] memory withdrawals =
+            new FastWithdrawalAgent.Withdrawal[](approvals.length);
+
+        address prevSigner;
+        for (uint i = 0; i < approvals.length; i++) {
+            validateApproval(approvals[i]);
```
Brechtpd, line 67: Changed to always validate the signature even if `msg.sender` is the owner/manager? Will increase gas costs by ~5k gas/withdrawal in that case.

dong77: I will add more change to provide an isApprovalValid method then skip checking the signatures if msg.sender is the manager.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
Brechtpd: Oops missed this. But users can easily check if the signature is valid before even doing the transaction themselves, no?

dong77 (approved):

PR #1727, opened 2020-09-10
[protocol 3.6] Remove short storageID uses

Brechtpd:
https://github.com/Loopring/team/issues/124

Numbers in the design doc updated for the worst case scenario where all bytes are non-zero (like we do for all other data as well), cost will be lower in practice.

### DIFF
--- a/packages/loopring_v3/test/testExchangeUtil.ts
+++ b/packages/loopring_v3/test/testExchangeUtil.ts
@@ -47,9 +47,7 @@ import {
   WithdrawalRequest
 } from "./types";
 
-const LoopringIOExchangeOwner = artifacts.require(
-  "LoopringIOExchangeOwner"
-);
+const LoopringIOExchangeOwner = artifacts.require("LoopringIOExchangeOwner");
 
 type TxType =
   | Noop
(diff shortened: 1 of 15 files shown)


packages/loopring_v3/test/testExchangeUtil.ts
```diff
@@ -2252,14 +2223,14 @@ export class ExchangeTestUtil {
       // Transform DA
       const transformedDa = new Bitstream();
       const size = Constants.TX_DATA_AVAILABILITY_SIZE;
-      const sizeA = 25;
-      const sizeB = 43;
-      assert.equal(sizeA + sizeB, size, "invalid transform sizes");
+      const size1 = 29;
```
yueawang, line 2226: Per my understandings, TX_DATA_AVAILABILITY_SIZE(68) is the max bytes of DA in all types, what's the reason of this 29+39(before 25+43)? I think it is a separation best fits for 1 specific type (transfer??), is it?

yueawang (approved):

packages/loopring_v3/test/testExchangeUtil.ts
Brechtpd: Yep, it's the number of bytes necessary for a trade. And transfers are normally a couple of bytes less so it works well for transfers as well (though if extra addresses are put onchain a transfer can actually make use of almost all 68 bytes, but that should be uncommon).

hosschao (approved):
LGTM

Issue #1735, opened 2020-09-13
[Loopring 3.6] fail faster in block submission


dong77:
As of now, if we don't wait for long enough and a deposit transaction get re-orged out of a block, our block submission will fail. But since the failure is deep inside its corresponding process() function, there will be a lot of gas wasted.

I think for each block, we can add some extra checks to make sure a block-level precodition must be true before we process each transaction. One such precondition is the hash of the last known block that contains at least one onchain request.

dong77:
https://github.com/Loopring/protocols/pull/1736/files

Brechtpd:
> As of now, if we don't wait for long enough and a deposit transaction get re-orged out of a block, our block submission will fail. But since the failure is deep inside its corresponding process() function, there will be a lot of gas wasted.

In most cases the largest part of the gas cost is the calldata (especially when disregarding the proof verification, which is done after all the block checks), so if the invalid blocks gets onchain we will already paid most of the gas we will end up paying for the blocks (unless the block contains a lot of withdrawals or needs to do a lot of onchain verifications). The invalid block getting onchain in any capacity is already costly and should be avoided (also it's pretty bad on the UX level because it can potentially revert transactions for users).

I think this should be detected by the relayer well before we even create the Ethereum tx. There's also lots of other logic like `validUntil` which can fail blocks, so I think it's inevitability up to the relayer to keep track of all this, lots of things that can go wrong which aren't always easy to check early onchain. Even if we don't wait for any confirmations for the deposit, the deposit will still just have been included in a block waiting for more transactions to fill up the block (there's no benefit to the user to get the block containing the deposit onchain as fast as possible). And then the block still needs to be proven as well, which likely takes up at least another couple of minutes. So there should be enough time to catch these things offchain (of course something can slip through the cracks on the relayer level, but then that seems like something that should be fixed).

dong77:
Talked to relayer engineers, my proposal is unnecessary, same reason as Brecht mentioned above.

PR #1736, opened 2020-09-13
[Loopring 3.6] submitBlocks precondition check to fail fast

dong77:
see https://github.com/Loopring/protocols/issues/1735

### DIFF
--- a/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
+++ b/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
@@ -60,6 +60,8 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
     }
 
     function submitBlocksWithCallbacks(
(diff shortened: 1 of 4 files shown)


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -90,7 +92,10 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
         }
 
         // Finally submit the blocks
-        IExchangeV3(target).submitBlocks(blocks);
+        IExchangeV3(target).submitBlocks(
+            preconditionBlockNumber,
+            preconditionBlockHash,
+            blocks);
```
Brechtpd, line 98: Why pass this into the protocol and not just do the necessary checks here in the owner contract?

dong77: We can do that as well

PR #1737, opened 2020-09-13
[Loopring 3.6] Demo AMM pool

Brechtpd:
- L1/L2 joins + L1/L2 exits
- Transferable liquidity tokens on L1.


packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -105,6 +146,10 @@ contract AmmPool is IBlockReceiver {
     // A map from a token to the total balance owned directly by LPs (so NOT owned by the pool itself)
```
letsgoustc, line 146: Looks like the locked only used in totalBalance function? My question is that why we divide the add or remove liquidity into two APIs, eg, for add, we have deposit and joinPool. Why not use just only one API like depoistAndJoinPool?

Brechtpd: `depositAndJoinPool` of course also useful in some cases, so definitely something that should be exposed as well in an actual implementation. But the separation of deposit/withdraw and join/exit is to make things easier and cheaper. Not every join will be successful, and so the funds will already be deposited in the pool contract. If the user wants to join with updated parameters than he only needs to join, not deposit funds again. The same when exiting, funds are still stored in the pool contract (when exiting to L1), so joining again doesn't need (or only partially needs) any new deposited funds.

Join can also be used to just authorize the join, with funds coming from L2 (though it seems I haven't exposed that yet). So in that way it's similar to `approveOffchainTransfer` on the exchange contract so that joins/exits can also be approved by smart contracts. I made join/exit to just be operations that do a basic new thing, similar to how approving a transfer on the exchange doesn't deposit the funds either.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
-        bytes32 exchangeDomainSeparator = exchange.getDomainSeparator();
+        PoolTransaction[] memory poolTransactions = abi.decode(auxiliaryData, (PoolTransaction[]));
+
+        // Cache the domain seperator to save on SLOADs each time it is accessed.
+        Context memory ctx = Context({
+            _block: _block,
+            txIdx: txIdx,
+            DOMAIN_SEPARATOR: DOMAIN_SEPARATOR,
+            exchangeDomainSeparator: exchange.getDomainSeparator(),
+            ammBalancesInAccount: new uint[](tokens.length),
```
dong77, line 413: ammBalancesInAccount => ammLayer2Balances?
ammBalances => ammLayer1Balances?

I'm not sure I understand the meaning of these variables correctly.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
             update.validUntil = 0xffffffff;
-            bytes32 txHash = AmmUpdateTransaction.hashTx(exchangeDomainSeparator, update);
+            bytes32 txHash = AmmUpdateTransaction.hashTx(ctx.exchangeDomainSeparator, update);
             exchange.approveTransaction(address(this), txHash);
-            numTransactionsConsumed++;
-            // AMM account balance now available onchain
-            ammBalancesInAccount[i] = update.balance;
-            ammBalances[i] = ammBalancesInAccount[i];
+            ctx.numTransactionsConsumed++;
+            if (start) {
```
dong77, line 497: Remind me this: we don't check the layer2 balances here because the circuit will guarantee the balances are indeed correct so the operator cannot lie about them?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -46,43 +56,62 @@ contract AmmPool is IBlockReceiver {
...
 
-    enum WorkType
+    struct PoolTransaction
     {
-        NONE,
-        JOIN,
-        EXIT
+        PoolTransactionType txType;
+        bytes               data;
+        bytes               signature;
```
dong77, line 107: Seems in the current design, this signature field will always be empty. Shall we simply remove it?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
             // Check that the AMM update in the block matches the expected update
-            AmmUpdateTransaction.AmmUpdate memory update = _block.readAmmUpdate(txIdx++);
+            AmmUpdateTransaction.AmmUpdate memory update = ctx._block.readAmmUpdate(ctx.txIdx++);
             require(update.owner == address(this), "INVALID_TX_DATA");
             require(update.accountID == accountID, "INVALID_TX_DATA");
-            require(update.tokenID == tokens[i].tokenID, "INVALID_TX_DATA");
+            require(update.tokenID == ctx.tokens[i].tokenID, "INVALID_TX_DATA");
             require(update.feeBips == feeBips, "INVALID_TX_DATA");
-            require(update.tokenWeight == tokens[i].weight, "INVALID_TX_DATA");
+            require(update.tokenWeight == (start ? 0 : ctx.tokens[i].weight), "INVALID_TX_DATA");
```
letsgoustc, line 491: Here I think we should verify the weight adjustment onchain instead of just use them from relayer?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
+        PoolJoin memory join,
+        bytes    memory signature
+        )
+        internal
+    {
+        bytes32 poolTxHash = hashPoolJoin(ctx.DOMAIN_SEPARATOR, join);
+        if (signature.length == 0) {
+            require(queue[queuePos].txHash == poolTxHash, "NOT_APPROVED");
+            delete queue[queuePos];
+            queuePos++;
```
dong77, line 516: if we delete the element from the queue, should we still do `queuePos++`?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
+        } else {
+            require(poolTxHash.verifySignature(join.owner, signature), "INVALID_SIGNATURE");
```
dong77, line 518: Even if we authenticate the request using signature, should we also delete the queued request by doing `delete queue[queuePos]`

dong77: Maybe we can delete the `queuePos` variable and always use `queue[0]` and `delete queue[0]`

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -85,16 +85,14 @@ library WithdrawTransaction
...
-            )
+        bytes20 onchainDataHash = hashOnchainData(
+            auxData.minGas,
+            auxData.to,
+            auxData.extraData
         );
         // Only the 20 MSB are used, which is still 80-bit of security, which is more
         // than enough, especially when combined with validUntil.
-        require(withdrawal.onchainDataHash == bytes20(onchainDataHash), "INVALID_WITHDRAWAL_DATA");
+        require(withdrawal.onchainDataHash == onchainDataHash, "INVALID_WITHDRAWAL_DATA");
```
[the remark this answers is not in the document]
Brechtpd: Yes, using up less of the 68 bytes/DA that are available per transaction.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: These are the weights that are stored on-chain (cached in `ctx` for saving on SLOADs), the relayer cannot set them freely (unless I made a mistake somewhere, but looks okay).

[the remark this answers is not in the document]
Brechtpd: Not necessary because the nonce doesn't really matter for transactions authorized by the contract (can only be used once immediately). The relayer is free to choose the nonce/storageID (which needs to be valid, otherwise it can't be included in a block).

For transactions done by the user we I just made the storageIDs used by the transfers something that's signed by the user directly. This way we don't have to store anything onchain either, replay protection is done on layer 2.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
             exchange.approveTransaction(address(this), txHash);
-            numTransactionsConsumed++;
-            // AMM account balance now available onchain
-            ammBalancesInAccount[i] = update.balance;
-            ammBalances[i] = ammBalancesInAccount[i];
+            ctx.numTransactionsConsumed++;
+            if (start) {
+                // AMM account balance now available onchain
+                ctx.ammBalancesInAccount[i] = update.balance;
+                ctx.ammBalances[i] = ctx.ammBalancesInAccount[i];
```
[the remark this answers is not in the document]
Brechtpd: > the new weights are correct

The weights always come directly from the contract itself so can's see additional checks that can be added for this.

> ctx.ammBalancesInAccount[i] == ctx.ammBalances[i]?

Yep, already doing that now! (additional safety check because transfers can have small rounding errors).

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: `ammBalancesInAccount` is the balance in the account directly. Transfers update the balance of the account immediately, deposits/withdrawals don't update the balance immediately because those operations are batched and only done at the end. So `ammBalances` is the virtual balance if all operations would immediately update the balance in the account (which are the ones we need to do the calculations on), `ammBalancesInAccount` is the actual balance in the account so we can keep track of how much we need to deposit/withdraw at the end.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Correct, the operator has no control over the balance stored in the DA (if it puts in invalid data the block is invalid).

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
-                require(withdrawal.amount == amount, "INVALID_TX_DATA");
-                require(withdrawal.feeTokenID == withdrawal.tokenID, "INVALID_TX_DATA");
-                require(withdrawal.fee == 0, "INVALID_TX_DATA");
-                require(withdrawal.to == address(this), "INVALID_TX_DATA");
-                require(withdrawal.extraData.length == 0, "INVALID_TX_DATA");
-                // Now approve this withdrawal
-                withdrawal.validUntil = 0xffffffff;
-                bytes32 txHash = WithdrawTransaction.hashTx(exchangeDomainSeparator, withdrawal);
+        // Mint liquidity tokens
+        mint(join.owner, join.poolAmountOut);
```
dong77, line 562: Hmm... I thought this `join.poolAmountOut` should be calculated instead of provided by the user in the `joinPool` function... same for `poolAmountIn`

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: It most definitely is being used. Like with transactions on the exchange, it's cheaper to let users sign a request than have them do an onchain tx. Also more flexible for us, so seems win-win to use that by default. Only the onchain exit function is really needed for safety.

Only supporting joins from layer 2 + signature would makes things simpler because we don't have to worry about deposits/locking up funds.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
+                require(transfer.from == join.owner, "INVALID_TX_DATA");
+                require(transfer.toAccountID == accountID, "INVALID_TX_DATA");
+                require(transfer.tokenID == ctx.tokens[i].tokenID, "INVALID_TX_DATA");
+                require(isAlmostEqual(transfer.amount, amount), "INVALID_TX_DATA");
+                require(transfer.fee == 0, "INVALID_TX_DATA");
+                if (signature.length != 0) {
+                    // Now approve this transfer
+                    transfer.validUntil = 0xffffffff;
+                    bytes32 txHash = TransferTransaction.hashTx(ctx.exchangeDomainSeparator, transfer);
+                    exchange.approveTransaction(join.owner, txHash);
```
dong77, line 547: This seems to be a bug: the user who want to join the pool must approve the transfer using his own onchain tx.  Now the operator will be able to move user's fund into the AMM pool if the user authorized this contract as an agent.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Currently the queue needs to be handled FIFO, so it forces the operator to handle the pool transactions in the correct order. Though I'm not that big of a fan of the queue and would like to remove it.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: > Even if we authenticate the request using signature, should we also delete the queued request by doing delete queue[queuePos]

In most cases there won't be any queue item for the pool transaction. Just a signature. Like with the exchange, a pool transaction can either be authorized with a signature or with an onchain approval.

> Maybe we can delete the queuePos variable and always use queue[0] and delete queue[0]

I don't think `delete queue[0]` pushes the next queue item to `[0]`.

I don't really like the queue, it's there because joins from layer 1 need to be locked up but if they need to be locked we could also just deposit to the exchange directly (but this will cost more gas because things like batching the deposits won't be possible...).

packages/loopring_v3/contracts/test/AmmPool.sol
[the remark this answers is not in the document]
dong77: > The weights always come directly from the contract itself
Sorry, I thought the new weights are calculated by the operator and the contract just verify the new weights based on the new balances. 

> Yep, already doing that now!
Do you mean the code will be pushed to this PR, but not here yet?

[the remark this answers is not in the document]
Brechtpd: > Do you mean the code will be pushed to this PR, but not here yet?

Yes, will push shortly.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +400,340 @@ contract AmmPool is IBlockReceiver {
...
+                ctx.ammBalancesInAccount[i] = ctx.ammBalancesInAccount[i].sub(amount);
+            } else {
+                // Make the amount available for withdrawing
+                balance[exit.owner][ctx.tokens[i].addr] = balance[exit.owner][ctx.tokens[i].addr].add(amount);
             }
+            ctx.ammBalances[i] = ctx.ammBalances[i].sub(amount);
+        }
+
+        // Burn liquidity tokens
+        burn(exit.owner, exit.poolAmountIn);
```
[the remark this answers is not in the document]
Brechtpd: I lock them up currently, but not that much of a problem now because they can't be transferred. But this is indeed the most important thing we HAVE to do onchain (other operations not really needed onchain). The user needs to prove he has N liquidity tokens so that he can withdraw from the pool.

> the operator's transaction will fail

This should be valid (but predictable). Otherwise the the pool can be manipulated shortly and the user may withdraw unexpected amounts.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Yeah, but only if he approved the pool transaction with a signature or by calling `join`? This is why we have agents that can do stuff like this, with their own authorization method for doing transactions for users.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: I just stuck with how balancer does things here, not necessarily the best way to do things.

Brechtpd:
One thing to keep in mind if we would allow changing the weights is that we shouldn't do so just based on what LPs deposit. If previous liquidity LPs were okay with a 50/50 pool that does not mean they are fine with being in a pool that is suddenly a 90/10 pool (which changes the risks completely) just because some LPs only want to deposit one of the tokens. If an LP only wants to provide a single token they could do so, but the weights should remain the same (like balancer does), unless a more advanced curve is just probably using oracles.

Another thing is I think that we should really try to just use signature based joins/exits instead of needing Ethereum transactions by users (cheap and flexible). One of the difficulties here is with layer 1 balances and locking those. Currently this works kind of okay, but I think things can be made simpler when we by default lock all tokens up in the contract, and people need to unlock them (so the inverse what I do now for the amounts being deposited, having them unlocked by default and only locking them when they will be used). The other possibility here is that we only support joins from layer 2 + default locked behaviour of the liquidity tokens on layer 1 so they can be withdrawn only when specifically unlocked with some delay or with the help of a signature of the operator.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -233,48 +284,58 @@ contract AmmPool is IBlockReceiver {
...
+    /// @param poolAmountIn The number of liquidity tokens that will be burned.
+    /// @param minAmountsOut The minimum amounts that need to be withdrawn when burning
+    ///                      the specified amount of liquidity tokens.
+    function exitPool(uint poolAmountIn, uint96[] calldata minAmountsOut, bool toLayer2)
         external
         online
     {
         require(minAmountsOut.length == tokens.length, "INVALID_DATA");
 
+        // Lock liquidity token - Not needed now with non-transferable liquidity token
```
dong77, line 319: If poolAmountIn and poolAmountOut as requirements cannot be met, how we can fail these Join/Exit?  I don't see related logics.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
-                DepositTransaction.Deposit memory _deposit = _block.readDeposit(txIdx++);
-                require(_deposit.owner == address(this), "INVALID_TX_DATA");
-                require(_deposit.accountID == accountID, "INVALID_TX_DATA");
-                require(_deposit.tokenID == tokens[i].tokenID, "INVALID_TX_DATA");
-                require(_deposit.amount == amount, "INVALID_TX_DATA");
-                // Now do this deposit
-                uint ethValue = 0;
-                if (tokens[i].addr == address(0)) {
-                    ethValue = _deposit.amount;
+        if (!join.fromLayer2) {
```
dong77, line 607: should we skip this if `valid` is false?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
+            ratio = (join.poolAmountOut * BASE) / poolTotal;
+        } else {
+            // Important for accuracy
+            require(join.poolAmountOut == INITIAL_SUPPLY, "INITIAL_SUPPLY_UNEXPECTED");
+        }
+
+        // Check if the requirements are fulfilled
+        bool valid = true;
+        uint[] memory amounts = new uint[](ctx.tokens.length);
+        for (uint i = 0; i < ctx.tokens.length; i++) {
```
dong77, line 560: should we use:
```solidity
for (uint i = 0; valid && i < ctx.tokens.length; i++) {
```

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
+        uint poolTotal = totalSupply();
+        uint ratio = (exit.poolAmountIn * BASE) / poolTotal;
+
+        // Check if the requirements are fulfilled
+        bool valid = availableBalance(address(this), exit.owner) >= exit.poolAmountIn;
+        uint[] memory amounts = new uint[](ctx.tokens.length);
+        for (uint i = 0; i < ctx.tokens.length; i++) {
+            amounts[i] = ctx.ammBalances[i] * ratio / BASE;
+            if(amounts[i] < exit.minAmountsOut[i]) {
+                valid = false;
```
dong77, line 642: I suggest we extra a `validateExit` method and returns a boolean for 'valid'. Inside `validateExit`, we should also check there are enough liquidity tokens to burn then burn it there.

Inside `validateExit`, we can return false immediately if valid is false for the first time

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
+            }
+        }
```
dong77, line 644: I suggest we extra a `validateExit` method and returns a boolean for 'valid'. Inside `validateExit`, we should also check there are enough liquidity tokens to burn then burn it there.

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
-                    locked[item.owner][tokens[i].addr] = locked[item.owner][tokens[i].addr].sub(join.maxAmountsIn[i]);
+    function processJoin(
+        Context  memory ctx,
+        PoolJoin memory join,
+        bytes    memory signature
+        )
+        internal
+    {
+        bytes32 poolTxHash = hashPoolJoin(ctx.DOMAIN_SEPARATOR, join);
+        if (signature.length == 0) {
```
dong77, line 539: is this an ECDSA signature, if so, when a signature is provided, this can still be a layer-1 join just using a onchain signature verification. Asserting `require(join.fromLayer2, "INVALID_DATA")` will be incorrect.

We have:

- layer1-join with onchain hash approval
- layer1-join with ECDSA signature approval
- layer-2 join with transfers (EdDSA signature approval)

dong77:
Brecht, it will be ideal if the operator of the exchange can choose to skip some Join/Exit requests to avoid being Sybil attacked.

dong77:
Is the reason for introducing `poolAmountOut` and `poolAmountOut` to allow the user to quit joining/exiting the pool if the price has changed a lot?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -360,111 +424,411 @@ contract AmmPool is IBlockReceiver {
...
+            require(queue[queuePos].txHash == poolTxHash, "NOT_APPROVED");
+            delete queue[queuePos];
+            queuePos++;
+        } else {
+            require(poolTxHash.verifySignature(join.owner, signature), "INVALID_SIGNATURE");
+            require(join.fromLayer2, "INVALID_DATA");
```
dong77, line 545: There might be a bug here.

If the join is from layer2, it means they are L2 transfers, so we don't have to do signature verification or hash checking based on the queue.

If the join is from layer1, the request can be an offchain request with the user's signature or an onchain request with a PoolJoin item in the queue.

dong77:
Update: Basically we want to make the AmmPool contract getting closer to what we desire for production. The feedback from the relayer team is that they prefer using a map instead of a queue for the join/exit requests, such that the operator of the exchange can choose which join/exit requests to process and in which order.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: `poolAmountOut` can never fail because that amount will be minted for the user. `poolAmountIn` can fail if the balance for the user is insufficient, is checked here:
```
bool valid = availableBalance(address(this), exit.owner) >= exit.poolAmountIn;
```

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Currently the funds can only be unlocked when processed by the queue and are locked specifically to process the work in the queue. So even if it's invalid we want to unlock the funds, otherwise they would remain locked.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Possible, though that's strange loop condition for a `for`. :) But it's an inexpensive loop so I would just keep it a standard one.

packages/loopring_v3/contracts/test/AmmPool.sol
Brechtpd: Liquidity token balance checked with `bool valid = availableBalance(address(this), exit.owner) >= exit.poolAmountIn;`

Brechtpd:
> Brecht, it will be ideal if the operator of the exchange can choose to skip some Join/Exit requests to avoid being Sybil attacked.

Joins we can skip, but on-chain exit requests will all need to be processed to make sure people can exit the pool, same as with forced withdrawals on the exchange. Offchain requests > onchain requests in all normal cases though, so hopefully these onchain requests will rarely be used.

> Is the reason for introducing poolAmountOut and poolAmountOut to allow the user to quit joining/exiting the pool if the price has changed a lot?

Especially for withdrawing is makes a lot of sense to just let users specify how many of their liquidity tokens they want to redeem back to tokens. Otherwise we/they have to approximate how many tokens need to be withdrawn to burn the required amount of liquidity tokens, it would be almost impossible for people to burn all their liquidity tokens like that.

For joins I think not specifying it would be okay (maybe even better in some cases when the pool is small and amounts being deposited), still some other safety checks need though I think.

> What's the worst thing that may happen if we remove the two parameters and calculate the number of liquidity tokens to mint/burn automatically?

I think only some additional slippage losses when the pool is in an unbalanced state or is brought into an unbalanced state. But I think if we do things differently than uniswap/balancer we should have very good reasons to do so, seems needlessly risky otherwise.


> Update: Basically we want to make the AmmPool contract getting closer to what we desire for production. The feedback from the relayer team is that they prefer using a map instead of a queue for the join/exit requests, such that the operator of the exchange can choose which join/exit requests to process and in which order.

Offchain requests > onchain requests, I don't want us to actually use these onchain requests. :)

To support this we'll have to switch to lock all users funds by default. So do we want to handle this in the pool contract or do we just deposit funds directly to the exchange so we don't have to worry about that? This way we only need to really support joins from layer 2. Though for the liquidity tokens themselves, if we always want to keep them on L1, we'd still have to have some locking mechanism in the pool as well.

Brechtpd:
#1771

packages/loopring_v3/contracts/test/AmmPool.sol
dong77: For joins, if the PoolTransction's signature is non-empty, will `join.fromLayer2` must be true?
FOr exits, if the PoolTransaction's signature is non-empty, will `exit.toLayer2` must be true?

dong77 (approved):

PR #1738, opened 2020-09-14
[Loopring 3.6] add submit block precondition check

dong77:
https://github.com/Loopring/protocols/issues/1735

### DIFF
--- a/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
+++ b/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
@@ -7,11 +7,12 @@ import "../../aux/compression/ZeroDecompressor.sol";
 import "../../core/iface/IExchangeV3.sol";
 import "../../thirdparty/BytesUtil.sol";
 import "../../lib/MathUint.sol";
+import "../../lib/ReentrancyGuard.sol";
 import "./SelectorBasedAccessManager.sol";
 import "./IBlockReceiver.sol";
 
 
-contract LoopringIOExchangeOwner is SelectorBasedAccessManager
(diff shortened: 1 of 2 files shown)


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -60,16 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
     }
 
     function submitBlocksWithCallbacks(
-        ExchangeData.Block[] memory blocks,
-        BlockCallback[]      memory callbacks
+        uint    preconditionBlockNumber,
+        bytes32 preconditionBlockHash,
+        ExchangeData.Block[] calldata blocks,
+        BlockCallback[]      calldata callbacks
```
dong77, line 58: Brecht, these are supposed to be "memory" instead of "calldata", right?

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -60,16 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
...
 
+        if (preconditionBlockNumber == 0) {
+            require(preconditionBlockHash == bytes32(0), "INVALID_PRECONDITION_BLOCK_HASH");
+        } else if (block.number.sub(preconditionBlockNumber) <= 256) {
+            require(
+                blockhash(preconditionBlockNumber) == preconditionBlockHash,
+                "INVALID_PRECONDITION"
+            );
+        }
+
```
dong77, line 76: @Brechtpd is there a way to avoid data copy using assembly to replace `IExchangeV3(target).submitBlocks(blocks);`?

PR #1739, opened 2020-09-14
[Loopring 3.6] submit block precondtion check

dong77:
See https://github.com/Loopring/protocols/issues/1735

### DIFF
--- a/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
+++ b/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
@@ -7,11 +7,12 @@ import "../../aux/compression/ZeroDecompressor.sol";
 import "../../core/iface/IExchangeV3.sol";
 import "../../thirdparty/BytesUtil.sol";
 import "../../lib/MathUint.sol";
+import "../../lib/ReentrancyGuard.sol";
 import "./SelectorBasedAccessManager.sol";
 import "./IBlockReceiver.sol";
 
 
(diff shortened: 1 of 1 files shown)


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -60,16 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
     }
 
     function submitBlocksWithCallbacks(
+        uint    preconditionBlockNumber,
+        bytes32 preconditionBlockHash,
```
dong77, line 56: Shall we use 'calldata' instead of 'memory'?

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -60,16 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager
...
 
+        if (preconditionBlockNumber == 0) {
+            require(preconditionBlockHash == bytes32(0), "INVALID_PRECONDITION_BLOCK_HASH");
+        } else if (block.number.sub(preconditionBlockNumber) <= 256) {
+            require(
+                blockhash(preconditionBlockNumber) == preconditionBlockHash,
+                "INVALID_PRECONDITION"
+            );
+        }
+
```
dong77, line 76: @Brechtpd is there a way optimize `IExchangeV3(target).submitBlocks(blocks);` using assembly to avoid unnecessary data copying?

PR #1762, opened 2020-09-18
Amm approve erc20 on setup

dong77:


packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -65,9 +65,7 @@ contract AmmPool is IBlockReceiver {
         uint pos
     );
 
-    uint public constant MAX_UINT = ~uint(0);
```
Brechtpd, line 68: Why not use a constant?

Brechtpd (approved):

PR #1763, opened 2020-09-18
allow withdraw all tokens

dong77:


packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -231,21 +231,30 @@ contract AmmPool is IBlockReceiver {
...
 
         // Withdraw
+        uint96[] memory withdrawn = new uint96[](tokens.length);
         for (uint i = 0; i < tokens.length; i++) {
             address token = tokens[i].addr;
-            require(availableBalance(token, msg.sender) >= amounts[i], "INSUFFICIENT_BALANCE");
-            balance[token][msg.sender] = balance[token][msg.sender].sub(amounts[i]);
-            withdrawInternal(token, amounts[i], msg.sender);
+            uint available = availableBalance(token, msg.sender);
+            if (amounts[i] == 0 || amounts[i] > available) {
```
Brechtpd, line 246: Maybe remove `amounts[i] == 0` here because not withdrawing some tokens may be useful. Just being able to use a very large value to withdraw everything seems enough.

PR #1771, opened 2020-09-18
[AMM pool] Changed approvals + locking logic

Brechtpd:
- Removed queue, pool transactions can be done in any order (but exits need to be processed).
- All tokens deposited to the pool contract are locked by default. Users can withdraw immediately with the help of a signature of the operator. They can also do an Ethereum transaction themselves to unlock the funds with a delay.
- Onchain join/exit only to be used for censorship resistance/onchain composability, otherwise signatures are better.
- Transferring the liquidity tokens can now easily be supported (just have to make it an ERC20 token and everything will work as expected). Deposit/Withdraw can be used to transfer them in and out.

This is the basic idea, still room for improvement and probably some bugs but will wait on feedback if this setup is okay.


packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -204,17 +211,18 @@ contract AmmPool is IBlockReceiver, IAgent {
...
     {
-        require(amounts.length == tokens.length, "INVALID_DATA");
+        require(amounts.length == tokens.length + 1, "INVALID_DATA");
 
         // Deposit the max amounts to this contract so we are sure
         // the amounts are available when the actual deposit to the exchange is done.
-        for (uint i = 0; i < tokens.length; i++) {
-            address token = tokens[i].addr;
+        for (uint i = 0; i < tokens.length + 1; i++) {
+            address token = (i < tokens.length) ? tokens[i].addr : address(this);
```
dong77, line 225: if token == address(this), then the safeTransferFromAndVerify will fail, won't it?

packages/loopring_v3/contracts/test/AmmPool.sol
```diff
@@ -204,17 +211,18 @@ contract AmmPool is IBlockReceiver, IAgent {
         }
     }
 
+    /// @param amounts The amounts to deposit + the amount of pool tokens to deposit
```
dong77, line 214: Not sure I understand: why do we support depositing pool tokens (liquidity tokens)?

dong77 (commented):
> All tokens deposited to the pool contract are locked by default. Users can withdraw immediately with the help of a signature of the operator. They can also do an Ethereum transaction themselves to unlock the funds with a delay.

Can we make *join*  part of the *deposit* operation so users don't have to do it using another tx?

dong77:
> Deposit/Withdraw can be used to transfer them in and out.
I'd suggest we don't do it in this PR.

dong77:
Brecht, please go ahead with this implementation.

PR #1772, opened 2020-09-19
[AMM Pool] Rename variables and extract methods

dong77:
@Brechtpd After working on this PR, I finally figured out how your design works :) I do have the following suggestions regarding the pool token's mint/burn logic:

- Now we calculate the amount of tokens that will be accepted into the pool based on the Join's `poolAmountOut` parameter. If `poolAmountOut` is large enough, this join will likely to fail; otherwise, if it is small enough, the user ends up with a smaller amount of pool tokens than he deserves. I suggest we calculate the amount of pool tokens assuming his join will be accepted, then compare the new pool tokens to mint to him with a `minPoolTokenAmount` parameter that the user provides. If `minPoolTokenAmount` is smaller than or equal to the pool tokens to be minted, then this Join will be accepted, otherwise, the join will be rejected. By doing this, a user will feel really comfortable by providing a small value for `minPoolTokenAmount`, or even zero if he doesn't care about slippage.

- For exits, I suggest we only take one parameter - the amount of pool tokens the user would like to burn. Then calculate the amount of each token that will be released back to layer1 for withdrawal. There should be no slippage option for exits at all, as users collectively owns what's left in the pool anyway.  I believe this is also the case with uniswap.

### DIFF
--- a/packages/loopring_v3/contracts/test/AmmPool.sol
+++ b/packages/loopring_v3/contracts/test/AmmPool.sol
@@ -130,8 +130,8 @@ contract AmmPool is IBlockReceiver, IAgent {
         uint    txIdx;
         bytes32 DOMAIN_SEPARATOR;
         bytes32 exchangeDomainSeparator;
-        uint[]  ammBalancesInAccount;
-        uint[]  ammBalances;
+        uint[]  ammBalancesBefore;
+        uint[]  ammBalancesAfter;
         uint    numTransactionsConsumed;
         Token[] tokens;
     }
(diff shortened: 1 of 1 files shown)


Brechtpd:
It seems like uniswap does have very similar checks: https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/UniswapV2Router01.sol. So both uniswap and balancer check this (for a good reason I assume).

I do think this doesn't matter a lot in most cases (I think it only really matters when the pool is small, otherwise it's just small potential slippage risks, but pools can be small so we have to make sure that scenario also works).

Brechtpd (approved):

PR #1787, opened 2020-09-24
Update WithdrawTransaction.sol

yueawang:
Change WITHDRAWAL_TYPEHASH signature to fully comply with eip-712, so that external integrators could use 3rd party eip-712 libs to get correct hash, i.e. this hash:

```
return EIP712.hashPacked(
            DOMAIN_SEPARATOR,
            keccak256(
                abi.encode(
                    WITHDRAWAL_TYPEHASH,
                    withdrawal.owner,
                    withdrawal.accountID,
                    withdrawal.tokenID,
                    withdrawal.amount,
                    withdrawal.feeTokenID,
                    withdrawal.fee,
                    withdrawal.to,
                    keccak256(withdrawal.extraData),
                    withdrawal.minGas,
                    withdrawal.validUntil,
                    withdrawal.storageID
                )
            )
        );
```

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
+++ b/packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
@@ -33,7 +33,7 @@ library WithdrawTransaction
     using ExchangeWithdrawals  for ExchangeData.State;
 
     bytes32 constant public WITHDRAWAL_TYPEHASH = keccak256(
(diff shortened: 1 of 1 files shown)


dong77 (dismissed):

dong77 (approved):

Brechtpd:
Good catch. Btw, the tests use [eth-sig-util](https://github.com/MetaMask/eth-sig-util) for signing (the standard as far as I know, don't know if there's any other lib that's widely used) and sometimes already depends on things that are not always strictly EIP712. The lib is currently at V4 which goes beyond what was in the initial standard to support things like arrays. So not super sure if `uint256` vs `uint` actually matters much for implementations.

PR #1788, opened 2020-09-24
Will drop this one.

dong77:


packages/loopring_v3/contracts/amm/libamm/AmmExchange.sol
```diff
@@ -0,0 +1,76 @@
...
+        require(poolAmountIn <= S.totalSupply, "INVALID_POOL_AMOUNT");
+
+        S.burn(msg.sender, poolAmountIn);
+
+        // Currently commented out to make the contract size smaller...
+        uint size = S.tokens.length;
+        if (S.exchange.isInWithdrawalMode()) {
+            // Check if all tokens were withdrawn using Merkle proofs
+            for (uint i = 0; i < size; i++) {
+                // Question(Brecht): I removed a "!" isWithdrawnInWithdrawalMode, is it now correct?
```
Brechtpd, line 38: Yep!

packages/loopring_v3/contracts/amm/libamm/AmmExchange.sol
```diff
@@ -0,0 +1,76 @@
...
+            address token = S.tokens[i].addr;
+
+            // Calculate the balance inside the pool
+            uint balance = token == address(0) ?
+                address(this).balance :
+                ERC20(token).balanceOf(address(this));
+
+            // Withdraw the part owned
+            uint amount = balance.mul(poolAmountIn) / _totalSupply;
+            AmmUtil.tranferOut(token, amount, msg.sender);
```
Brechtpd, line 73: Need to subtract `totalLockedBalance` here from the contract balance because not all the funds are owned by the pool, LPs may have funds in `lockedBalance`, and those funds are also stored in this contract.

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
```diff
@@ -0,0 +1,95 @@
...
+        )
+        public
+    {
+        uint size = S.tokens.length;
+        require(amounts.length == size, "INVALID_DATA");
+
+        if (S.isExiting[msg.sender]) {
+            // Q: 这个标记的用途？
+            // This could suddenly change the amount of liquidity tokens available, which
+            // could change how the operator needs to process the exit.
```
Brechtpd, line 36: An exit depends on how many liquidity tokens the user has locked up, so we need this to keep the locked up balance the same while we are processing the exit. Otherwise the user may have done an exit with `poolAmount == 100`, but only locked up `50`. Then we are processing the exit, create a block + proof, but in the meantime the user could have deposited additional tokens which would make how we process the block invalid. We also cannot just partially process the exit when additional funds are available, because then we could censor exits and lock up user funds in the pool.

We also cannot check how many liquidity tokens the user has when calling `exitPool` and reserve them. This could also suddenly change how we need to process a block while we could already have been building a block. Scenario:
- A user requests an offchain exit with `poolAmount == 100
- We see that the user indeed has 100 liquidity tokens locked up so we start to process it.
- The user now calls `exitPool` to force the operator to do the exit.
- Onchain we see that the user has 100 liquidity tokens, but if we would reserve them for the onchain request the offchain request we are already processing will be rendered invalid.

So what we want to do here is still do the offchain request that we are already doing, and then process the onchain request with any remaining liquidity tokens that are still locked up. But the remaining locked amount needs to remain the same, otherwise the processing of the onchain request is not deterministic.

packages/loopring_v3/contracts/amm/libamm/AmmUtil.sol
```diff
@@ -0,0 +1,104 @@
...
+    using AddressUtil       for address;
+    using ERC20SafeTransfer for address;
+
+    bytes32 constant public POOLEXIT_TYPEHASH = keccak256(
+        "PoolExit(address owner,bool toLayer2,uint256 poolAmountIn,uint256[] minAmountsOut,uint32[] storageIDs,uint256 validUntil)"
+    );
+
+    bytes32 constant public POOLJOIN_TYPEHASH = keccak256(
+        "PoolJoin(address owner,bool fromLayer2,uint256 minPoolAmountOut,uint256[] maxAmountsIn,uint32[] storageIDs,uint256 validUntil)"
+    );
```
Brechtpd, line 24: Move to respective join/exit libs?

packages/loopring_v3/contracts/amm/libamm/AmmUtil.sol
```diff
@@ -0,0 +1,104 @@
...
+                    exit.toLayer2,
+                    exit.poolAmountIn,
+                    keccak256(abi.encodePacked(exit.minAmountsOut)),
+                    keccak256(abi.encodePacked(exit.storageIDs)),
+                    exit.validUntil
+                )
+            )
+        );
+    }
+}
```
Brechtpd, line 104: Move to respective join/exit libs?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -0,0 +1,208 @@
...
+        uint96[] calldata maxAmountsIn,
+        bool              fromLayer2,
+        uint              validUntil
+        )
+        external
+        onlyWhenOnline
+        nonReentrant
+    {
+        state.deposit(0, maxAmountsIn);
+        state.joinPool(minPoolAmountOut, maxAmountsIn, fromLayer2, validUntil);
```
Brechtpd, line 153: Probably best to move the events `emit Deposit(msg.sender, poolAmount, amounts);` and `emit PoolJoinRequested(join);` to the internal functions so they are always emitted.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -0,0 +1,127 @@
...
+            _block: _block,
+            txIdx: txIdx,
+            domainSeperator: S.domainSeperator,
+            exchangeDomainSeparator: S.exchange.getDomainSeparator(),
+            ammActualL2Balances: new uint96[](size),
+            ammExpectedL2Balances: new uint96[](size),
+            numTransactionsConsumed: 0,
+            tokens: S.tokens,
+            size: size,
+            poolTokenTotalSupply: poolTokenTotalSupply,
```
Brechtpd, line 48: While we can cache this, it needs to be updated for each mint/burn as they happen to do all calculations correctly. In theory the total supply state variable only needs to be read/updated once instead of for each mint/burn, but that makes things a bit more complicated (multiples updates to the same storage aren't cheap but also not super expensive, it's 1,600 gas/update for the SLOAD + SSTORE).

packages/loopring_v3/contracts/amm/libamm/AmmUtil.sol
dong77: Done.

packages/loopring_v3/contracts/amm/libamm/AmmUtil.sol
dong77: Done

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
dong77: done

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
dong77: remove this cache.

packages/loopring_v3/contracts/amm/libamm/AmmExchange.sol
dong77: done, please check

PR #1792, opened 2020-09-25
[Loopring V3] Refactor Loopring AMM contract (#1791)

dong77:
All issues found in https://github.com/Loopring/protocols/pull/1788 have been fixed in this PR.


packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -0,0 +1,129 @@
...
+    struct State {
+        // Pool token state variables
+        string name;
+        string symbol;
+        uint   totalSupply;
+
+        mapping(address => uint) balanceOf;
+        mapping(address => mapping(address => uint)) allowance;
+        // Question: should we use the same `nonces` variable?
+        mapping(address => uint) permitNonces; // ERC2612
```
Brechtpd, line 100: I think we can reuse it. In any case, the nonce used for the permits should be called `nonces` otherwise we don't follow the standard.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -0,0 +1,129 @@
...
+
+        // AMM pool state variables
+        IExchangeV3 exchange;
+        uint32      accountID;
+        bytes32     domainSeperator;
```
Brechtpd, line 105: I don't know why the previous PR was dropped, but I committed some small changes there that did some minor improvements. One of them was the type here -> separator: https://github.com/Loopring/protocols/pull/1788/commits/a47c18ef09956349f8993cb4379f2f5448a2c8be

packages/loopring_v3/contracts/amm/libamm/AmmExchange.sol
```diff
@@ -0,0 +1,68 @@
...
+
+        for (uint i = 0; i < size; i++) {
+            address token = S.tokens[i].addr;
+
+            // Question(Brecht): I removed a "!" isWithdrawnInWithdrawalMode, is it now correct?
+            require(
+                withdrawalMode && exchange.isWithdrawnInWithdrawalMode(accountID, token) ||
+                !withdrawalMode && !exchange.isForcedWithdrawalPending(accountID, token),
+                "PENDING_WITHDRAWAL"
+            );
```
Brechtpd, line 43: Looks correct, find this a bit more complicated to understand without the if on withdrawalMode.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+        WithdrawTransaction.Withdrawal memory withdrawal = ctx._block.readWithdrawal(ctx.txIdx++);
+        ctx.numTransactionsConsumed++;
+
+        // These fields are not read by readWithdrawal:
+        withdrawal.minGas = 0;
+        withdrawal.to = address(this);
+        withdrawal.extraData = new bytes(0);
+
+        require(
+            withdrawal.withdrawalType == 1 && // Question(brecht): should this be 1?
```
Brechtpd, line 50: It will be, but it also can't be anything else. Can't hurt to add this check.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+            withdrawal.owner == address(this) &&
+            withdrawal.accountID == S.accountID &&
+            withdrawal.tokenID == token.tokenID &&
+            // withdrawal.amount.isAlmostEqual(amount) && // Question(brecht):should we use isAlmostEqual?
+            withdrawal.amount == amount &&
```
Brechtpd, line 55: No, deposits/withdrawals don't have any rounding errors because we put in the complete uint96 in the DA.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+            withdrawal.feeTokenID == 0 &&
+            withdrawal.fee == 0 &&
+            withdrawal.storageID == 0 && // Question(brecht):should we check this ID?
```
Brechtpd, line 58: No, the relayer can freely set the storegeID to any value that makes the withdrawal valid. We don't know what value that is onchain (storageID == 0 can only be used once).

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+    {
+        S.validatePoolTransaction(
+            exit.owner,
+            AmmUtil.hashPoolExit(ctx.domainSeperator, exit),
+            signature
+        );
+
+        if (signature.length == 0) {
+            // This is an onchain exit, we're processing it now so stop tracking it.
+            S.isExiting[msg.sender] = false;
```
Brechtpd, line 92: `msg.sender` -> `exit.owner`: https://github.com/Loopring/protocols/pull/1788/commits/a47c18ef09956349f8993cb4379f2f5448a2c8be

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+
+                require(
+                    transfer.fromAccountID == S.accountID &&
+                    transfer.from == address(this) &&
+                    transfer.to == exit.owner &&
+                    transfer.tokenID == ctx.tokens[i].tokenID &&
+                    transfer.amount.isAlmostEqual(amount) &&
+                    transfer.feeTokenID == 0 &&
+                    transfer.fee == 0 &&
+                    transfer.storageID == (signature.length > 0 ? exit.storageIDs[i] : 0), // Question (brecht):is this right?
```
Brechtpd, line 113: No, shouldn't check storageIDs here when `signature.length == 0`.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -0,0 +1,174 @@
...
+        uint                  validUntil
+        )
+        private
+        returns (bool)
+    {
+
+        // Check if we can withdraw without unlocking with an approval
+        // from the operator.
+        if (signature.length == 0) {
+            require(validUntil == 0, "INVALID_VALUE");
```
Brechtpd, line 114: Does this matter?

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -0,0 +1,174 @@
...
+        bool                  approvedByOperator
+        )
+        private
+        returns (uint withdrawn)
+    {
+        uint available = approvedByOperator ?
+            S.lockedBalance[token][msg.sender] :
+            S.availableBalance(token, msg.sender);
+
+        withdrawn = (amount > available || amount == 0) ? available : amount;
```
Brechtpd, line 152: I believe `amount == 0` should be valid and not withdraw anything, otherwise we force people to always withdraw from all tokens which is strange? (and not something we actually want in some cases)

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -0,0 +1,211 @@
...
+                        transfer.storageID == 0, // Question (brecht):should we check this?
+                        "INVALID_OUTBOUND_TX_DATA"
+                    );
+
+                   // Now approve this transfer
+                    transfer.validUntil = 0xffffffff;
+                    txHash = TransferTransaction.hashTx(ctx.exchangeDomainSeparator, transfer);
+                    ctx.exchange.approveTransaction(address(this), txHash);
+                    amount = amount.sub(transfer.amount);
+                }
```
Brechtpd, line 146: I really don't think we should do this. Rounding errors are very small, if we are worried about these extremely small rounding errors (which the operator fully controls) we can always force that the users are a disadvantage instead of the pool (but the operator can already do this, the smart contract would just enforce this). These are the same rounding errors that happen when trading, so if we are okay with them there we should be okay with them here.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -0,0 +1,211 @@
...
+            if (join.fromLayer2) {
+                TransferTransaction.Transfer memory transfer = ctx._block.readTransfer(ctx.txIdx++);
+                ctx.numTransactionsConsumed++;
+
+                // We do not check these fields: fromAccountID, to, amount, feeTokenID, fee
+                require(
+                    transfer.toAccountID == S.accountID &&
+                    transfer.from == join.owner &&
+                    transfer.tokenID == ctx.tokens[i].tokenID &&
+                    transfer.storageID == (signature.length > 0 ? join.storageIDs[i] : 0), // Question (brecht):is this right?
```
Brechtpd, line 110: No, see above.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -0,0 +1,211 @@
...
+        if (!slippageRequirementMet) return;
+
+        for (uint i = 0; i < ctx.size; i++) {
+            uint96 amount = amounts[i];
+
+            if (join.fromLayer2) {
+                TransferTransaction.Transfer memory transfer = ctx._block.readTransfer(ctx.txIdx++);
+                ctx.numTransactionsConsumed++;
+
+                // We do not check these fields: fromAccountID, to, amount, feeTokenID, fee
```
Brechtpd, line 105: `transfer.fee` should be checked to be something that the user approves, now the operator can just fill in any fee and drain the user's account. So if we want these to be non-zero, the fee will have to be part of the join request that the user approves.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -0,0 +1,211 @@
...
+
+        // Calculate the amount of liquidity tokens that should be minted
+        for (uint i = 0; i < ctx.size; i++) {
+            if (ctx.ammExpectedL2Balances[i] > 0) {
+                uint amountOut = uint(join.maxAmountsIn[i])
+                    .mul(ctx.poolTokenTotalSupply) / uint(ctx.ammExpectedL2Balances[i]);
+
+                if (poolAmountOut == 0 || amountOut < poolAmountOut) {
+                    poolAmountOut = amountOut;
+                }
```
Brechtpd, line 193: Using `poolAmountOut == 0` like this is not necessarily correct here. `maxAmountsIn[0]` could be `0` which would correctly set `poolAmountOut` to `0`, but then it could still be incorrectly increased by following `maxAmountsIn`s.

This example is of course not very realistic because why would a user set `maxAmountsIn == 0`, but this is still a valid input and with rounding errors and tokens with low decimals I think there are cases this could end up being wrong.

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
```diff
@@ -0,0 +1,103 @@
...
+        AmmData.State storage S,
+        uint96[]     calldata amounts
+        )
+        public
+    {
+        uint size = S.tokens.length;
+        require(amounts.length == size + 1, "INVALID_DATA");
+
+        // Question(brecht): I don't understand this part, may be reasonable inside
+        // the withdrawlFromPool function, but I'm not sure it's necessary at all.
```
Brechtpd, line 33: > An exit depends on how many liquidity tokens the user has locked up, so we need this to keep the locked up balance the same while we are processing the exit. Otherwise the user may have done an exit with `poolAmount == 100`, but only locked up `50`. Then we are processing the exit, create a block + proof, but in the meantime the user could have deposited additional tokens which would make how we process the block invalid. We also cannot just partially process the exit when additional funds are available, because then we could censor exits and lock up user funds in the pool.
> 
> We also cannot check how many liquidity tokens the user has when calling `exitPool` and reserve them (we could check the balance, but that doesn't mean they will still be there when we actually process it). This could also suddenly change how we need to process a block while we could already have been building a block. Scenario:
> 
>  - A user requests an offchain exit with `poolAmount == 100
>  - We see that the user indeed has 100 liquidity tokens locked up so we start to process it.
>  - The user now calls `exitPool` to force the operator to do the exit.
>  - Onchain we see that the user has 100 liquidity tokens, but if we would reserve them for the onchain request the offchain request we are already processing will be rendered invalid.
> 
> 
> So what we want to do here is still do the offchain request that we are already doing, and then process the onchain request with any remaining liquidity tokens that are still locked up. But the remaining locked amount needs to remain the same, otherwise the processing of the onchain request is not deterministic.

So you don't think that changing the amount of liquidity tokens being available can invalidate exits already being processed by the operator?

`withdrawlFromPool` from pool should be fine because the operator knows how the balance will change for the user (either with the signature or because of the unlocking). For deposits this is not the case.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -0,0 +1,211 @@
...
+
+        // Check that the deposit in the block matches the expected deposit
+        DepositTransaction.Deposit memory deposit = ctx._block.readDeposit(ctx.txIdx++);
+        ctx.numTransactionsConsumed++;
+
+        require(
+            deposit.owner == address(this) &&
+            deposit.accountID == S.accountID &&
+            deposit.tokenID == token.tokenID &&
+            // deposit.amount.isAlmostEqual(amount),
```
Brechtpd, line 50: Not needed.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
Brechtpd: Oh I misread, I see you'll still fix the issues in previous PR.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
Brechtpd: Oh forgot the `nonces` are stored in a struct so seems like we'll have to create a public getter named `nonces` to conform with the standard any way.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,172 @@
...
+    {
+        amounts = new uint96[](ctx.size);
+
+        // Check if we can still use this exit
+        if (block.timestamp > exit.validUntil) {
+            return (false, amounts);
+        }
+
+        // Check if the user has enough pool tokens
+        if (S.lockedBalance[address(this)][exit.owner] < exit.poolAmountIn) {
```
Brechtpd, line 156: Mistake here by me: we should only look at the balance actually locked here:

```
function lockedBalance(
    AmmData.State storage S,
    address               token,
    address               owner
    )
    internal
    view
    returns (uint)
{
   return S.lockedBalance[address(this)][owner] - availableBalance(address(this), owner);
}
```

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
```diff
@@ -0,0 +1,103 @@
...
+        if (S.isExiting[msg.sender]) {
+            // This could suddenly change the amount of liquidity tokens available, which
+            // could change how the operator needs to process the exit.
+            require(amounts[0] == 0, "CANNOT_DEPOSIT_LIQUIDITY_TOKENS_WHILE_EXITING");
+        }
+
+        S.lockedUntil[msg.sender] = 0;
```
Brechtpd, line 40: I think users having the option to immediately lock the tokens again can also invalidate blocks:
- User unlocks funds, but keeps funds in the pool contract
- User does an onchain exit which the operator needs to process
- The funds are unlocked so the operator needs to process the exit as if no liquidity tokens are available. The operator start building the block that processes the exit
- The user now locks up all funds again, so the liquidity tokens are available again and need to be used in the exit
- The operator needs to re-do the exit and throw away any blocks/txs that depend on the wrong exit.

So I think any change from locked -> unlocked and unlocked -> locked needs to be done with a decent delay.

Brechtpd: Another way this could be solved is that for onchain exits `poolAmountIn` is locked up in yet another place, which can only be used by onchain exits (cannot even be used by exits that are approved with a signature) and need to either be transferred from the unlocked balance or be newly deposit directly to it. When processing the onchain exit it's either burned or moved to the unlocked funds (similar to how it previously worked with the queue).

This would partly make things easier onchain for all the locking/unlocking headaches, but we'd need to track yet another set of balances (though this should be farily easy because only used for the onchain exits, but need to be made available for withdrawing when the pool is shutdown etc...).

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
dong77: yes

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
dong77: fixed

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
dong77: https://github.com/Loopring/protocols/commit/a47c18ef09956349f8993cb4379f2f5448a2c8be merged

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: fixed

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
dong77: Not really, just want to make sure there are no random data.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
dong77: OK

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: This is not because of the rounding errors Brecht.

When a user wants to join on Layer2, he will first query from the relayer how much tokens he will transfer to the AMM pool, then sign the corresponding transfers. This takes time, maybe several seconds or even a minute.  The pool is constantly being used for swapping, so when the relayer process these layer-2 join transfers, the amount that can be accepted may have changed already. So we have to have a way to refund overpaid transfers.

The layer-1 join is different - the contract only deducts tokens that have really been accepted.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: Not sure I get it. Please help me to fix it using a small PR.

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
dong77: Please proceed with a fix of your preference.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: I'll enable non-zero fees later. Added two TODOs

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: fixed

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
dong77: I'll figure it out

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -0,0 +1,128 @@
...
+
+    struct PoolJoin
+    {
+        address  owner;
+        bool     fromLayer2;
+        uint     minPoolAmountOut;
+        uint96[] maxAmountsIn;
+        uint32[] storageIDs;
+        uint     validUntil;
+        uint96   fee;
```
Brechtpd, line 44: Seems like we should have an array of fees here, otherwise there's a fixed fee amount for all different tokens?

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -0,0 +1,169 @@
...
+                ctx.numTransactionsConsumed++;
+
+                require(
+                    transfer.fromAccountID == S.accountID &&
+                    transfer.from == address(this) &&
+                    transfer.to == exit.owner &&
+                    transfer.tokenID == ctx.tokens[i].tokenID &&
+                    transfer.amount.isAlmostEqual(amount) &&
+                    transfer.feeTokenID == ctx.tokens[i].tokenID &&
+                    transfer.fee.isAlmostEqual(exit.fee),
```
Brechtpd, line 110: For exits the fee should be subtracted from the amount withdrawn for the user, otherwise it's the pool account that pays the fee: `amount := transfer.fee + transfer.amount`.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -0,0 +1,212 @@
...
+
+        emit Withdrawal(msg.sender, amountOuts);
+    }
+
+    function exitPool(
+        AmmData.State storage S,
+        uint                  poolAmountIn,
+        uint96[]     calldata minAmountsOut,
+        bool                  toLayer2,
+        uint96                fee
```
Brechtpd, line 88: A non-zero fee is non-trivial because `fee <= amountWithdrawn` otherwise it's impossible so we'd have to check that.

But the main problem is that the operator is _required_ to process this onchain exit request regardless of the fee (so no reason to provide a non-zero fee). A fixed ETH fee like we do for forced withdrawals seems necessary while just setting `fee := 0`.

packages/loopring_v3/contracts/amm/libamm/AmmJoinRequest.sol
```diff
@@ -0,0 +1,167 @@
...
+/// @title AmmJoinRequest
+library AmmJoinRequest
+{
+    using ERC20SafeTransfer for address;
+    using MathUint          for uint;
+    using MathUint96        for uint96;
+    using SafeCast          for uint;
+
+    bytes32 constant public POOLJOIN_TYPEHASH = keccak256(
+        "PoolJoin(address owner,bool fromLayer2,uint256 minPoolAmountOut,uint256[] maxAmountsIn,uint32[] storageIDs,uint256 validUntil)"
```
Brechtpd, line 24: Probably still TODO, but the fee not part of signed message yet.

dong77: fixed

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
dong77: I removed exit fees.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: removed exit fees

Brechtpd (approved):

PR #1797, opened 2020-09-26
[AMM] Small improvements

Brechtpd:
Tried to keep things as simple as possible (because locking/unlocking is normally not needed) and so didn't add yet another place balances could be stored.

Locking/unlocking should also be fairly simply, lock/unlock requests cannot overlap so that a state change between lock/unlock always takes `LOCK_DELAY` seconds. I think this makes things easy for the relayer.

### DIFF
--- a/packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
+++ b/packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
@@ -68,7 +68,7 @@ library AmmExitProcess
         ctx.exchange.approveTransaction(address(this), txHash);
 
         // Total balance in this contract increases by the amount withdrawn
-        S.totalLockedBalance[token.addr] = S.totalLockedBalance[token.addr].add(amount);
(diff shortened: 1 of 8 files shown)


packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -111,6 +111,11 @@ library AmmExitProcess
                     "INVALID_TX_DATA"
                 );
 
+                // Replay protection when using a signature (otherwise the approved hash is cleared onchain)
+                if (signature.length > 0) {
+                    require(transfer.storageID == exit.storageIDs[i], "INVALID_TX_DATA");
+                }
```
Brechtpd, line 117: Seems like this was removed even when a signature is used. Unless some other onchain replay protection is implemented (which will be more expensive) there is no other way here to prevent replay protection (even EdDSA signed txs, which could in theory be used for the join transfers, are impossible because the pool account has no EdDSA public key).

dong77 (approved):

PR #1806, opened 2020-09-28
[Loopring AMM] refactored AMM pool

dong77:
Attempt to change the design as follows:

- Do not require the relayer to monitor the lock/unlock of tokens in LoopringAmmPool.sol, and only count on the PoolJoin/PoolExit events. If these events are emitted, it is guaranteed that the corresponding balances are locked. In other words, balances are locked per join/exit request. This not-monitoring the lock/unlock balances & timestamps will allow the relayer to be much simpler and there will be no race conditions.

- Joins using funds on layer1 and exits that burn pool tokens on layer1 must be approved onchain.


packages/loopring_v3/contracts/amm/LoopringAmmFactory.sol
```diff
@@ -22,28 +22,27 @@ contract LoopringAmmFactory is ReentrancyGuard
         poolImplementation = _poolImplementation;
     }
 
-    function createPool(AmmData.PoolConfig calldata config)
+    function createPool(
+        uint salt,
+        AmmData.PoolConfig calldata config
```
kongliangzhong, line 27: LPToken need to be registered in Exchange, so LPTokenId need to be added in PoolConfig.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -27,67 +29,97 @@ library AmmBlockReceiver
...
+    function afterBlockSubmission(
+        AmmData.State      storage S,
+        ExchangeData.Block memory  /* _block */
+        )
+        public
+    {
+        uint poolAmountToBurn = S.poolAmountToBurn;
+        if (poolAmountToBurn > 0) {
+            S.poolAmountToBurn = 0;
+            S.burn(address(this), poolAmountToBurn);
```
kongliangzhong, line 109: LPToken distribution may fail in Exchange, then maybe there will not be enough token to burn. So we should do the LPToken withdrawal from Exchange first.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -29,88 +29,39 @@ library AmmExitRequest
     );
 
     bytes32 constant public POOLEXIT_TYPEHASH = keccak256(
-        "PoolExit(address owner,bool toLayer2,uint256 poolAmountIn,uint256[] minAmountsOut,uint32[] storageIDs,uint256 validUntil)"
+        "PoolExit(address owner,uint256 poolAmountIn,uint256[] minAmountsOut,uint256 validUntil,bool exitToLayer2,burnFromLayer2)"
```
kongliangzhong, line 32: bool burnFromLayer2

PR #1812, opened 2020-09-30
[Loopring AMM] refactored AMM pool (#1806)

dong77:
Attempt to change the design as follows:

- Do not require the relayer to monitor the lock/unlock of tokens in LoopringAmmPool.sol, and only count on the PoolJoin/PoolExit events. If these events are emitted, it is guaranteed that the corresponding balances are locked. In other words, balances are locked per join/exit request. This not-monitoring the lock/unlock balances & timestamps will allow the relayer to be much simpler and there will be no race conditions.

- Joins using funds on layer1 and exits that burn pool tokens on layer1 must be approved onchain.

@Brechtpd @kongliangzhong please ignore AmmWithdrawal.sol for now and all TODOs.


packages/loopring_v3/contracts/amm/libamm/AmmUpdateProcess.sol
```diff
@@ -25,20 +25,20 @@ library AmmUpdateProcess
...
-    function processAmmUpdates(
+    function approveAmmUpdates(
         AmmData.State    storage S,
         AmmData.Context  memory  ctx,
         bool                     opening
         )
         internal
     {
-        for (uint i = 0; i < ctx.size; i++) {
+        for (uint i = 1; i < ctx.size; i++) {
```
Brechtpd, line 35: Should be `0` I think (LT added at the end).

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
-            uint            poolAmountOut,
+            bool            slippageOK,
+            uint96          mintAmount,
             uint96[] memory amounts
         )
     {
         // Check if we can still use this join
-        amounts = new uint96[](ctx.size);
-        uint _totalSupply = S.totalSupply;
+        amounts = new uint96[](ctx.size - 1);
```
Brechtpd, line 174: `ctx.size`?

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
-            return(true, ctx.poolTokenInitialSupply, join.maxAmountsIn);
+        if (ctx.totalSupply == 0) {
+            return(true, ctx.poolTokenInitialSupply.toUint96(), join.joinAmounts);
         }
 
-        // Calculate the amount of liquidity tokens that should be minted
-        for (uint i = 0; i < ctx.size; i++) {
+        // Calculate the amount of pool tokens that should be minted
+        bool initialized = false;
+        for (uint i = 1; i < ctx.size; i++) {
```
Brechtpd, line 186: `1` -> `0`?

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
+            require(join.nonce == 0, "LAYER2_JOIN_WITH_NONCE_DISALLOWED");
+        }
+
         // Check if the requirements are fulfilled
-        (bool slippageRequirementMet, uint poolAmountOut, uint96[] memory amounts) = _calculateJoinAmounts(S, ctx, join);
+        (bool slippageOK, uint96 mintAmount, uint96[] memory amounts) = _calculateJoinAmounts(ctx, join);
 
-        if (!slippageRequirementMet) return;
+        if (!slippageOK) {
+            if (joinFromLayer1) {
```
Brechtpd, line 71: If `joinAmounts[i] != amounts[i]` we don't refund and leave the funds of the user locked inside this contract? (locked up forever until perhaps the pool is shutdown and then divided by the remaining LPs). But if that is the goal (not sure if that's a good idea) we may as well deposit the full amount to the pool when joining from L1.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
         }
 
         // Calculate the amounts to deposit
-        uint ratio = poolAmountOut.mul(ctx.poolTokenBase) / _totalSupply;
+        uint ratio = ctx.poolTokenBase.mul(mintAmount) / ctx.totalSupply;
 
-        for (uint i = 0; i < ctx.size; i++) {
-            amounts[i] = ratio.mul(ctx.ammExpectedL2Balances[i] / ctx.poolTokenBase).toUint96();
+        for (uint i = 1; i < ctx.size; i++) {
+            amounts[i - 1] = ratio.mul(ctx.ammExpectedL2Balances[i] / ctx.poolTokenBase).toUint96();
```
Brechtpd, line 208: Seems like the liquidity token was the first token and this worked differently in previous versions. :)

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
-            }
+        bool mintToLayer1 = (
+            join.direction == AmmData.Direction.L1_TO_L1 ||
+            join.direction == AmmData.Direction.L2_TO_L1);
 
-            ctx.ammExpectedL2Balances[i] = ctx.ammExpectedL2Balances[i].add(amount);
+        if (mintToLayer1) {
+            S.mint(join.owner, mintAmount);
+        } else {
+            _approvePoolTokenDeposit(ctx, mintAmount, join.owner);
```
Brechtpd, line 123: Seems like we should still mint the tokens here to `address(this)` so they can be deposited to the user.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
+        AmmData.Token memory poolToken = ctx.tokens[ctx.size];
+
+        // Question(brecht):can we read the `to` address?
+        require(
+            deposit.owner == address(this) &&
+            deposit.accountID== ctx.accountID &&
+            deposit.tokenID == poolToken.tokenID &&
+            deposit.amount == amount,
+            "INVALID_TX_DATA"
+        );
```
Brechtpd, line 150:
```
require(
    deposit.owner == to &&
    deposit.tokenID == poolToken.tokenID &&
    deposit.amount == amount,
    "INVALID_TX_DATA"
);
```

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -74,11 +81,35 @@ library AmmBlockReceiver
...
+        S.approveAmmUpdates(ctx, false);
+
+        // Save the pool tokens to burn
+        uint totalSupply = S.totalSupply;
+
+        if (ctx.totalSupply > totalSupply) {
+            S.mint(address(this), ctx.totalSupply - totalSupply);
+        } else {
+            S.poolTokenToBurn = totalSupply - ctx.totalSupply;
+        }
```
Brechtpd, line 93: Have to do the pool token distribution after this I guess, though maybe have to watch out for joins/exits from the same user.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -79,93 +38,152 @@ library AmmExitProcess
...
-
         // Calculate how much will be withdrawn
-        uint ratio = exit.poolAmountIn.mul(ctx.poolTokenBase) / S.totalSupply;
+        uint ratio = ctx.poolTokenBase.mul(exit.burnAmount) / ctx.totalSupply;
 
-        for (uint i = 0; i < ctx.size; i++) {
-            amounts[i] = (ratio.mul(ctx.ammExpectedL2Balances[i]) / ctx.poolTokenBase).toUint96();
-            if (amounts[i] < exit.minAmountsOut[i]) {
+        for (uint i = 0; i < ctx.size - 1; i++) {
+            amounts[i] = (ratio.mul(ctx.ammExpectedL2Balances[i + 1]) / ctx.poolTokenBase).toUint96();
```
Brechtpd, line 185: The usual. :)

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -24,97 +24,71 @@ library AmmExitRequest
...
         );
 
-        amountOuts = new uint[](size + 1);
-        amountOuts[0] = _withdrawToken(
-            S, address(this), amounts[0], approvedByOperator
-        );
+        uint32 lockIdx;
+        if (fromLayer1) {
+            require(burnStorageID == 0, "INVALID_STORAGE_ID");
+            require(S.exitLockIdx[msg.sender] == 0, "ONLY_ONE_LAYER1_EXIT_PER_USER_ALLOWED");
```
Brechtpd, line 55: Why is this done?

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -79,93 +38,152 @@ library AmmExitProcess
...
+            exit.direction == AmmData.Direction.L1_TO_L1 ||
+            exit.direction == AmmData.Direction.L1_TO_L2
+        );
+
+        if (burnFromLayer1) {
+            require(signature.length == 0, "LAYER1_EXIT_WITH_OFFCHAIN_APPROVAL_DISALLOWED");
+
+            delete S.exitLockIdx[msg.sender];
+            delete S.exitLocks[S.exitLocksIndex];
+            S.exitLocksIndex++;
```
Brechtpd, line 57: `msg.sender` -> `exit.owner`?

This data is deleted but there doesn't seem to be any link with the actual exit being processed? Also don't understand what the purpose it of this data. Maybe for manual withdrawals when the pool is shutdown is possible just be posting the request data onchain again. Still not as user friendly as just a single balance like it was where people could in theory do it themselves by interacting with the contract with a simple `withdraw`, but still very doable because it's just reposting data from events (seems easier than extra onchain data).

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -79,93 +38,152 @@ library AmmExitProcess
...
+
+
+        if (signature.length > 0) {
+            require(withdrawal.storageID == burnStorageID, "INVALID_STORAGE_ID_OR_REPLAY");
+        }
+
+        require(
+            withdrawal.withdrawalType == 1 &&
+            withdrawal.owner == from &&
+            withdrawal.accountID == 0 &&
```
Brechtpd, line 149: Will have to be ignored (which is fine), `accountID == 0`won't be valid (`0` is not a special value for any account). I guess it could also come from the exit request, but I don't much reason for that.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -79,93 +38,152 @@ library AmmExitProcess
...
+            withdrawal.tokenID == ctx.tokens[0].tokenID &&
```
Brechtpd, line 150: `withdrawal.tokenID == ctx.tokens[ctx.size].tokenID`

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -74,11 +81,35 @@ library AmmBlockReceiver
...
 
         return ctx.numTransactionsConsumed;
     }
 
+    function afterBlockSubmission(
```
Brechtpd, line 98: Maybe easier/more efficient to just modify `totalSupply()` to return `totalSupply.sub(totalBurned)` (with maybe a function to actually burn the tokens and resetting `totalBurned` back to 0, though shouldn't really be needed because uint256 is plenty)?

packages/loopring_v3/contracts/amm/libamm/AmmExchange.sol
```diff
@@ -3,68 +3,128 @@
...
+        AmmData.State storage S
+        )
+        internal
+    {
+        uint size = S.tokens.length;
+        address[] memory owners = new address[](size);
+        address[] memory tokenAddresses = new address[](size);
+
+        for (uint i = 0; i < size; i++) {
+            owners[i] = msg.sender;
```
Brechtpd, line 125: I think this should be `owners[i] = address(this)`.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
+
+            if (join.nonce == S.joinLocks[msg.sender].length) {
+                S.joinLocks[msg.sender].pop();
+            } else {
+                delete S.joinLocks[msg.sender][join.nonce - 1];
+
+                if (S.joinLockIdx[msg.sender] == join.nonce - 1) {
+                    S.joinLockIdx[msg.sender]++;
+                }
+            }
```
Brechtpd, line 62: `msg.sender` -> `join.owner`.

Seems like a risk currently is that `joinLocks` can grow without limit for users, and for withdrawing we have to run over this list, so could potentially lock up user funds if the list is too big.

Could you explain `joinLockIdx`? Or maybe the use will explain itself when withdrawals are done.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -105,25 +123,21 @@ library AmmData
...
 
-        // A map from an owner to the timestamp from which all funds of the user funds will be locked
-        mapping (address => uint) lockedSince;
-        // A map from an owner to the timestamp until all funds of the user are locked
-        // A zero value == locked indefinitely.
-        mapping (address => uint) lockedUntil;
+        mapping (address => mapping (address => uint96)) balance;
 
-        // A map from a token to the total balance owned directly by LPs (so NOT owned by the pool itself)
-        mapping (address => uint) totalUserBalance;
```
Brechtpd, line 121: Will have to come back unless user balances and pool balances (which when shutdown are force withdrawn to the owner, so also the pool contract) are kept separately.

Brechtpd:
Another way to allow withdrawing from joins/exits is to just let users post the onchain join/exit data again, check if it was indeed approved and allow them to withdraw their funds when expired/shutdown. Seems like that would simplify things a bit onchain, still possible for users to do themselves (though sure, more complicated then just calling `withdraw`). Though I assume because you did things like you did you probably think otherwise. :)

Joins from layer 1 (with token transfers) at this point maybe not exactly a better experience (or even cheaper) for users compared to just depositing to the exchange and doing an L2 join? (I mean, we could simply deposit to the exchange in the background instead if transferring the funds in the pool contract). Especially if the join fails, because then it means waiting for the join to expire/fail (we'd have to spend gas to do the failed join so it can be withdrawn faster, but that's a useless operation that costs quite some gas to us without any fee, withdrawing the funds again and redoing the join request, all costly in time and gas. Maybe it makes even more sense to keep things simple? Don't have to worry about special fund management in the pool with special cased withdrawals, if the join fails cheap/easy to retry because no need to withdraw/deposit again, funds no longer locked up for some fixed time, fast withdrawals available like normal L2 balances....

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
Brechtpd: I think exits can be simplified while keeping a similar mechanism (exits can be done out-of-order):
- Just keep a counter to limit the number of exits (increment when creating the request, decrement when processing it)
- Store the amount like `S.lockedToBurn[exit.owner] = exit.burnAmount`, if `S.lockedToBurn[exit.owner]  > 0` user is already exiting.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
Brechtpd: The `delete S.exitLocks[S.exitLocksIndex];` doesn't seem to be directly related to the exit being processed. I think for the current system `S.exitLockIdx[exit.owner] == S.exitLocks[S.exitLocksIndex]` needs to be true, otherwise it's possible an unrelated `exitLocks` is deleted when the exits are done out-of-order.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
dong77: I want to make exits similar to layer AMMs, they process exits by order.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: fixed

dong77:
Brecht, please take another look. I know, there are a lot of changes... I'm exhausted.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
             return (false, 0, amounts);
         }
 
         // Calculate the amounts to deposit
-        uint ratio = poolAmountOut.mul(ctx.poolTokenBase) / _totalSupply;
+        uint ratio = ctx.poolTokenBase.mul(mintAmount) / ctx.totalSupply;
 
-        for (uint i = 0; i < ctx.size; i++) {
-            amounts[i] = ratio.mul(ctx.ammExpectedL2Balances[i] / ctx.poolTokenBase).toUint96();
+        for (uint i = 1; i < ctx.size; i++) {
```
Brechtpd, line 207: `1` -> `0`

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
-
             } else {
-                // Make the amount unavailable for withdrawing
-                address token = ctx.tokens[i].addr;
-                S.userBalance[token][join.owner] = S.userBalance[token][join.owner].sub(amount);
+                for (uint j = 0; j < ctx.size; j++) {
+                    address token =  ctx.tokens[j].addr;
+                    S.addUserBalance(ctx.exchange.owner(), token, join.joinFees[j]);
+                    S.addUserBalance(join.owner, token, join.joinAmounts[j].sub(amounts[j]));
+                }
```
Brechtpd, line 99: I don't think anything should be here.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
+        // Handle pool tokens
+        if (join.mintToLayer2) {
+            S.mint(address(this), mintAmount);
+            _approvePoolTokenDeposit(ctx, mintAmount, join.owner);
+        } else {
+            S.mint(join.owner, mintAmount);
         }
+    }
+
+    function _approvePoolTokenDeposit(
```
Brechtpd, line 112: Seems like this could re-use `approveTokenDeposit` with some small changes.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
-        S.mint(address(this), poolAmountOut);
-        S.addUserBalance(address(this), join.owner, poolAmountOut);
+        require(
+            deposit.to == to &&
+            deposit.tokenID == poolToken.tokenID &&
+            deposit.amount == amount,
+            "INVALID_TX_DATA"
+        );
+
+        ctx.exchange.deposit{value: 0}(
```
Brechtpd, line 134: Liquidity token currently not approved anywhere (using `approveTokenDeposit` would fix that).

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
         }
 
         // Calculate the amounts to deposit
-        uint ratio = poolAmountOut.mul(ctx.poolTokenBase) / _totalSupply;
+        uint ratio = ctx.poolTokenBase.mul(mintAmount) / ctx.totalSupply;
 
-        for (uint i = 0; i < ctx.size; i++) {
-            amounts[i] = ratio.mul(ctx.ammExpectedL2Balances[i] / ctx.poolTokenBase).toUint96();
+        for (uint i = 1; i < ctx.size; i++) {
+            amounts[i - 1] = ratio.mul(ctx.ammExpectedL2Balances[i] / ctx.poolTokenBase).toUint96();
```
Brechtpd, line 208: `i - 1` -> `i`

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
Brechtpd: So `afterAllBlocksSubmitted` can now be removed because it's not necessary any more?

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -4,124 +4,135 @@ pragma solidity ^0.7.0;
...
-            );
-        }
+        appproveDepositOrWithdrawal(ctx);
 
-        // The closing AMM updates
-        S.processAmmUpdates(ctx, false);
+        S.approveAmmUpdates(ctx, false);
 
         return ctx.numTransactionsConsumed;
     }
```
Brechtpd, line 53: Small improvement/simplication I didn't think of before: We don't have to track `numTransactionsConsumed`, we can just calculate it here as `ctx.txIdx - txIdx`.

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
```diff
@@ -65,22 +70,34 @@ library AmmStatus
...
+
+        uint size = S.tokens.length;
+
+        // remember the part owned by users collectively
+        for (uint i = 0; i < size; i++) {
+            address token = S.tokens[i].addr;
+            S.withdrawableBeforeShutdown[token] = (token == address(0)) ?
+                address(this).balance :
+                ERC20(token).balanceOf(address(this));
+        }
```
Brechtpd, line 98: Balances withdrawn from `withdrawable` and joints/exits, which we also need to support when shutdown, will need to be subtracted from `withdrawableBeforeShutdown` to keep it correct

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -83,109 +37,178 @@ library AmmJoinProcess
...
+        AmmData.TokenLock[] storage joinLocks = S.joinLocks[join.owner];
+
+        if (join.nonce == joinLocks.length) {
+            joinLocks.pop();
+        } else {
+            delete joinLocks[join.nonce - 1];
+            if (S.joinLocksStartIdx[join.owner] == join.nonce - 1) {
+                S.joinLocksStartIdx[join.owner]++;
+            }
+        }
```
Brechtpd, line 158: `joinLocksStartIdx` never decreases here (or even in `withdraw` as far as I can tell), so I think it's possible to increase `joinLocksStartIdx`, then the operator processes the latest joins each time. In that case `joinLocks.length == 0` and `joinLocksStartIdx > 0`, and so when the user does another join that join will be missed in `withdraw`.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -79,93 +37,153 @@ library AmmExitProcess
...
+                ctx.ammActualL2Balances[i] = ctx.ammActualL2Balances[i].sub(amount);
+
                 TransferTransaction.Transfer memory transfer = ctx._block.readTransfer(ctx.txIdx++);
                 ctx.numTransactionsConsumed++;
 
                 require(
-                    transfer.fromAccountID == S.accountID &&
+                    transfer.fromAccountID== ctx.accountID &&
+                    // transfer.toAccountID == UNKNOWN &&
+                    // transfer.storageID == UNKNOWN &&
```
Brechtpd, line 86: How is replay protection handled when `exit.burnFromLayer2 == false`?

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -0,0 +1,169 @@
...
+                    amounts[size - 1] = amounts[size - 1].add(lock.amounts[0]);
+                } else if (lock.amounts.length == size - 1) {
+                    for (uint j = 0; j < size - 1; j++) {
+                        amounts[j] = amounts[j].add(lock.amounts[j]);
+                    }
+                }
+            } else {
+                return (newJoinLocksStartIdx, amounts);
+            }
+        }
```
Brechtpd, line 137: I assume this needs to also check the exitLocks, because the case `lock.amounts.length == 1` doesn't happen for joins.

Brechtpd:
I think my main impression now is that things are quite a bit too complex:
- The amount of options is immense and sometimes they have an effect one another. It's hard to keep track of everything. One of the things I think we greatly improved on in 3.6 compared to 3.1 is to keep things very simple and don't implement things nobody really cares about which add a lot of complexity. For example: do people really care about that exits are forced to be processed in order when doing an onchain request (and onchain exit requests aren't the type of request we actually want people to use)? We removed a similar thing in 3.1with deposits/withdrawals in 3.6 because people (even we) don't care and it makes things cheaper and easier.
- Complex (at least for smart contracts) data structures (like for locking now), hard to test.
- Even for the relayer/users/UI funds can be locked up in different pool contracts, under different constraints, for different reasons and for different times. The relayer may not need to directly monitor locking/unlocking, but it seems the end result is pretty much the same because all these things need to tracked for users.

The number of scenario's to support and the number of test cases necessary to test everything is very large.

And I think the main source of this complexity is by trying to do too much on L1. So (like yesterday I guess :) my question here is: why? Is there a benefit? It seems to only make things more complex for the smart contracts, users and UI (balances available/locked up until some date in one of many pool contracts etc...). I know I implemented those L1 joins/exits as well, but the more I think about it the less sense it makes. 

How easy and user friendly wouldn't things get when we just do everything on L2 (including mint/burning, only special case needed for force withdraw on the pool contract). Why would we do our best to *not* let people go through L2 directly? 
- Seems to be *less* user friendly: 
  - Even if someone wants to exit to L1 they can just do a withdrawal after the exit (largely the same cost), the time the user gets his funds is exactly the same with the benefit of *the funds actually arriving in his wallet* without any additional Ethereum tx (and having more options like a custom `to` address). And in the best case the user has immediate access to his funds with fast withdrawals.
  - And, again, easy tracking of balances which is kind of a big thing I think. Same single `balance` and `available balance` that we have now can be used, easy for users to understand and easy to make UI, and a system that the relayer already supports, though I don't know of course what the details here are.
- Uses *more* gas:
  - Total gas costs will be higher for L1 joins because of all the work we have to do onchain (and especially if the rejoin fails).
  - Total gas costs for L1 exits will also be higher for the user and for us (for the user because he needs to do an additional Ethereum tx to actually have access to his funds, and for us because we need to update the onchain state so the balance is withdrawals. Both are unnecessary for L2 exits).
  - Minting/Burning the liquidity tokens (which it seems we do want to do on L2) also cheaper of course than doing that onchain.

Just my opinion. I certainly don't know the difficulties with actually implementing the necessary services needed for all of this so I could be very wrong.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -4,124 +4,135 @@ pragma solidity ^0.7.0;
...
-            ammExpectedL2Balances: new uint96[](size),
-            numTransactionsConsumed: 0,
-            tokens: S.tokens,
-            size: size,
-            poolTokenBase: AmmData.LP_TOKEN_BASE(),
-            poolTokenInitialSupply: AmmData.LP_TOKEN_INITIAL_SUPPLY()
-        });
+        AmmData.Context memory ctx = getContext(S, _block, txIdx);
+
+        // Question(brecht): is it OK not to check the following for every poolTx?
```
Brechtpd, line 38: In theory the exchange owner can be the owner of multiple exchanges, and we have to make sure the block is for the correct exchange. But yeah, seems a bit far fetched, but not impossible. :)

Reading the header is pretty cheap so I don't think there's a problem keeping it.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
Brechtpd: Ah maybe there was some misunderstanding. I was just trying to avoid having to do another callback afterwards, I was fine with how the caching worked before. But now that only a single transaction can be processed it probably doesn't matter that much any more either way.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
Brechtpd: I unresolved a couple of these, but I'm actually not sure how you want this to be stored in the arrays now. I think there's definitely some inconsistency because in some places the index can go out of bounds.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -4,124 +4,135 @@ pragma solidity ^0.7.0;
...
-        // This also pulls the AMM balances onchain.
-        S.processAmmUpdates(ctx, true);
+        // The openning AMM updates will pull the latest AMM layer2 balances onchain.
+        S.approveAmmUpdates(ctx, true);
 
-        // Process all pool transactions
-        for (uint i = 0; i < poolTransactions.length; i++) {
-            _processPoolTransaction(S, ctx, poolTransactions[i]);
-        }
+        processPoolTx(S, ctx, poolTxData);
```
Brechtpd, line 46: Single transaction makes things easier? Relayer may of course only support 1 transaction (for now), but didn't seem to make much difference onchain to be able to support more.

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -0,0 +1,169 @@
...
+
+            // Check that nothing is withdrawable anymore.
+            require(
+                exchange.getAmountWithdrawable(address(this), token) == 0,
+                "MORE_TO_WITHDRAWAL"
+            );
+        }
+
+        // Withdraw proportionally to the liquidity owned, including the pool token itself
+        for (uint i = 0; i < size; i++) {
```
Brechtpd, line 85: Skip for the liquidity token.

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -0,0 +1,169 @@
...
+            address token = S.tokens[i].addr;
+
+            uint balance = token == address(0) ?
+                address(this).balance :
+                ERC20(token).balanceOf(address(this));
+
+            // Withdraw the part owned by the pool
+            uint amount = balance
+                .sub(S.withdrawableBeforeShutdown[token])
+                .mul(burnAmount) / S.totalSupply;
```
Brechtpd, line 95: Currently needs to call `totalSupply()`, but depends on if we always burn all liquidity tokens or not (`poolSupplyToBurn`).

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
dong77: Will start from simple and we probably add batch support if needed.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: The last element of the array is the pool token.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -17,105 +18,84 @@ import "./AmmUtil.sol";
...
     using SafeCast          for uint;
     using SignatureUtil     for bytes32;
 
-    bytes32 constant public WITHDRAW_TYPEHASH = keccak256(
-        "Withdraw(address owner,uint256[] amounts,uint256 validUntil,uint256 nonce)"
-    );
-
     bytes32 constant public POOLEXIT_TYPEHASH = keccak256(
-        "PoolExit(address owner,bool toLayer2,uint256 poolAmountIn,uint256[] minAmountsOut,uint32[] storageIDs,uint256 validUntil)"
+        "PoolExit(address owner,bool burnFromLayer2,uint96 burnAmount,bool burnFromLayer2,uint32 burnStorageID,bool exitToLayer2,uint96[] exitMinAmounts,uint256 validUntil,uint32 nonce)"
```
yueawang, line 30: duplicate "bool burnFromLayer2"

PR #1816, opened 2020-10-02
[Loopring AMM] a simplified version using only transfers on L2

dong77:
Based on Brecht and Kongliang's feedback, I greatly simplified the AMM pool implementation by:

- Normal Joins and Exits can only be approved with the user's ECDSA signatures, not even onchain hashes (we can add this later if desired). Users can only join with layer2 funds and the pool tokens will be deposited to the user on layer2; Users can also only exit with layer2 pool tokens, and their funds will be returned to them on layer2 as well.
- Forced Exits are possible with an Ethereum transaction, and they can be processed out of order.
- All pool tokens are minted and deposited to L2 upon setup.


packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -6,186 +6,157 @@ pragma experimental ABIEncoderV2;
...
-            uint            poolAmountOut,
+            bool            slippageOK,
+            uint96          mintAmount,
             uint96[] memory amounts
         )
     {
         // Check if we can still use this join
-        amounts = new uint96[](ctx.size);
-        uint _totalSupply = S.totalSupply;
+        amounts = new uint96[](ctx.size - 1);
```
Brechtpd, line 122: `ctx.size - 1` -> `ctx.size`

kongliangzhong (approved):

PR #1825, opened 2020-10-04
Amm uinttest

kongliangzhong:


Brechtpd:
Don't forget the previous version was tested here: https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/test/testAMMPool.ts. Will still contain useful code and should be a fairly good starting point! Just making sure you know. :)

kongliangzhong:
> Don't forget the previous version was tested here: https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/test/testAMMPool.ts. Will still contain useful code and should be a fairly good starting point! Just making sure you know. :)

Got it, thanks!

PR #1827, opened 2020-10-04
[draft] Initiate one Context object per token per submitBlocksWithCallbacks

dong77:


Brechtpd:
This'll definitely work, but I'm not going to make any predictions on how much gas this'll save though. Setting up the context object each time will cost 5,000 - 10,000 gas probably, this will have also have some overhead which will probably be pretty low but unsure how low. Things like having the block inside the context (and so serializing/deserializing that as well) we probably shouldn't do.

Some other things that we should look into optimizing eventually: 
- `beforeBlockSubmission` will do expensive memory copies, but don't want to do any premature optimizations here because should be carefully tested.
- Because we always depend on ECDSA signatures (or similar onchain approvals) we don't really need to do the ending AMM updates which should make things quite a bit cheaper if we process each pool tx separately, if we don't plan on using the EdDSA transfers to join feature (which definitely has some drawbacks which doesn't make it that attractive) then maybe we should remove the restriction that transfers to non-zero weight tokens are disallowed.
- Support batched `approveTransaction`s on the exchange which will remove call + proxy + nonReentrant (~2500 gas) overheads for each approval we need to do. But would want to see if this actually makes a lot of difference in testing.

dong77:
> This'll definitely work, but I'm not going to make any predictions on how much gas this'll save though. Setting up the context object each time will cost 5,000 - 10,000 gas probably, this will have also have some overhead which will probably be pretty low but unsure how low. Things like having the block inside the context (and so serializing/deserializing that as well) we probably shouldn't do.
> 
> Some other things that we should look into optimizing eventually:
> 
> * `beforeBlockSubmission` will do expensive memory copies, but don't want to do any premature optimizations here because should be carefully tested.
> * Because we always depend on ECDSA signatures (or similar onchain approvals) we don't really need to do the ending AMM updates which should make things quite a bit cheaper if we process each pool tx separately, if we don't plan on using the EdDSA transfers to join feature (which definitely has some drawbacks which doesn't make it that attractive) then maybe we should remove the restriction that transfers to non-zero weight tokens are disallowed.
> * Support batched `approveTransaction`s on the exchange which will remove call + proxy + nonReentrant (~2500 gas) overheads for each approval we need to do. But would want to see if this actually makes a lot of difference in testing.

I got 1 and 3 done, but not the second one.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -21,73 +21,88 @@ library AmmBlockReceiver
...
             poolTokenID: S.poolTokenID,
-            poolTokenBurnedSupply: S.poolTokenBurnedSupply,
-            size: size,
             tokens: S.tokens,
-            tokenBalancesL2: new uint96[](size)
+            tokenBalancesL2: new uint96[](size),
+            poolTokenBurnedSupply: S.poolTokenBurnedSupply,
+            pendingTxIdx: 0,
+            pendingTxOwners: new address[](size * 3),
+            pendingTxHashes: new bytes32[](size * 3)
```
Brechtpd, line 42: Seems like the code currently depends on these values to remain updated across calls to `onAmmTransaction`, until `afterAllBlocks` is called once/context object, but the context object is passed into an external/public call, so any updates to the struct in `onAmmTransaction` will be lost after the call ends. The updated context object would need to be returned by  `onAmmTransaction` if it also stores these dynamic variables (or could be stored in a different object, not sure about gas costs).

`pendingTxOwners` and `pendingTxHashes` are only big enough for a single call to `onAmmTransaction`. We should calculate those offchain and pass them onchain (would expensive to calculate onchain).

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -21,73 +21,88 @@ library AmmBlockReceiver
...
+    function beforeEachBlock(
+        AmmData.State      storage /* S */,
+        ExchangeData.Block memory  _block,
+        AmmData.Context    memory  ctx
+        )
+        internal
+        pure
+    {
+        BlockReader.BlockHeader memory header = _block.readHeader();
+        require(header.exchange == address(ctx.exchange), "INVALID_EXCHANGE");
```
Brechtpd, line 65: Seems unlikely it's worth doing an extra call for these, especially if the functions are specifically written for AMM. Doing an external call (+ passing through the complete block) will be expensive. Will be cheaper/easier to just run over the contexts in `LoopringIOExchangeOwner` (which contains the exchange address the pool expects) and check them against the exchange addresses in the blocks there.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -142,32 +147,49 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+        for (uint i = 0; i < config.receivers.length; i++) {
+            config.receivers[i].afterAllBlocks(ctxs[i]);
         }
     }
 
-    function _processTxCallbacks(
-        ExchangeData.Block memory _block,
-        TxCallback[]       calldata txCallbacks,
-        address[]          calldata receivers
+    // Question(brecht): is `calldata` used correctly here?
```
Brechtpd, line 166: The big problem is the block here. Even in the most efficient scenario (which we're not in here, we still do costly memory->memory copies here generated by solidity) it's still expensive to pass through the complete block for each update (for a block with 512 transactions each call including the block will cost `(3+3)*(68/32)*512 = 6,528` gas just for the block data). And it's also pretty useless, we only use a couple of transactions, copying all the other transactions is useless work and costs gas.

To have lowest cost here we can just extract the transactions that are needed in `onAmmTransaction`. We can make a dummy `Block` object if we want just containing those couple of transactions (this way nothing needs to be changed in the pool contract) or just pass in the raw transaction data (that's the only thing we actually use currently I think). One extra input that we need to be able so that we can do that is the number of transactions that are expected to be processed in the callback to `TxCallback` (instead of just calculating that on the fly like we do now).

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -613,6 +619,23 @@ contract ExchangeV3 is IExchangeV3
...
     }
 
+    function approveTransactions(
+        address[] calldata owners,
+        bytes32[] calldata transactionHashes
+        )
+        external
+        override
+        nonReentrant
+        onlyFromUserOrAgent(owner)
```
Brechtpd, line 629: remove (unfortunately a bit dangerous that the owner of the contract is also just `owner`, which we often use as the name of our function inputs).

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -613,6 +619,23 @@ contract ExchangeV3 is IExchangeV3
...
+    {
+        require(owners.length == transactionHashes.length, "INVALID_DATA");
+        for (uint i = 0; i < owners.length; i++) {
+            require(isUserOrAgent(owners[i]), "UNAUTHORIZED");
```
Brechtpd, line 633: Possible easy optimization here to add the ability to `IAgentRegistry` to batch check an agent against multiple owners:
```
function isAgent(
    address[] calldata users,
    address agent
    )
    external
    override
    view
    returns (bool)
{
    if (isUniversalAgent(agent)) {
        return true;
    }
    for (uint i = 0; i < users.length; i++) {
        if (!isUserAgent(users[i], agent)) {
            return false;
        }
    }
    return true;
}
```

The AMM pool will be a universal agent so instead of a CALL + SREAD / transaction, with this function we have a fixed cost of a single CALL + SREAD.

Brechtpd:
@dong77 Want me to continue work on this?

dong77:
> @dong77 Want me to continue work on this?

Yes please, I feel like the current code is still costly and there are room for optimizing the numbers of SSTORE and SLOAD.

PR #1828, opened 2020-10-04
[Loopring AMM 1.1] reduce gas by not mainining special address's approval and balance data

dong77:
///        This contract do not maintain the following data to save gas:
///        S.balanceOf[address(this)], and
///        S.allowance[*][address(this)]

### DIFF
--- a/packages/loopring_v3/contracts/amm/libamm/AmmPoolToken.sol
+++ b/packages/loopring_v3/contracts/amm/libamm/AmmPoolToken.sol
@@ -42,9 +42,7 @@ library AmmPoolToken
         view
         returns (uint balance)
     {
-        return owner == address(this) ?
-          uint96(-1) :
-          S.balanceOf[owner];
+        return owner == address(this) ? uint96(-1) : S.balanceOf[owner];
     }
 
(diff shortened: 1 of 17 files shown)


packages/loopring_v3/contracts/amm/libamm/AmmPoolToken.sol
```diff
@@ -10,6 +10,9 @@ import "../../lib/SignatureUtil.sol";
 
 
 /// @title AmmPoolToken
+///        This contract do not maintain the following data to save gas:
+///        S.balanceOf[address(this)], and
```
Brechtpd, line 14: i think this balance is only used when people do a forced exit/withdraw when shutdown. So in normal cases it doesn't impact gas costs, so perhaps still useful to have the default behavior where the real balance is returned for tools like etherscan (vs returning uint96(-1) which may be confusing for people/tools).

dong77: This PR is created when we support layer1 joins and exits. Now it can indeed be dropped.

PR #1832, opened 2020-10-05
use abi.encode instead o abi.encodePacked.

yueawang:
Refer to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md

> The array values are encoded as the keccak256 hash of the concatenated encodeData of their contents (i.e. the encoding of SomeType[5] is identical to that of a struct containing five members of type SomeType).

I think the spec may not force us using abi.encode, and encodePacked can be another proposal, but so far gateway and 3rd-party python lib(for MM) can easily support the type-encoded data array rather then encodePacked. So switch back to abi.encode just like Hebao does for address[] can save some many dev efforts.


yueawang:
[] in abi.encode is dynamic type, it does not fit here.
Still need to support encodePacked.

PR #1844, opened 2020-10-07
[Loopring AMM 1.1] Withdraw from deposit requests

dong77:


packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -16,6 +16,8 @@ library AmmWithdrawal
     using AmmPoolToken      for AmmData.State;
     using MathUint          for uint;
 
+    // Question(brecht): since all joins/exits are using transfers now, shall we
+    // delete this function?
```
Brechtpd, line 20: Okay with me. The only scenario where it can still be useful is when shutdown and all funds need to withdrawn back to the contract using forced withdrawals/Merkle tree, and so funds could still be stuck in the exchange. But anyone can call `withdrawFromApprovedWithdrawals` for this contract so it's basically just a utility function that could make things a bit easier.

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -32,6 +34,30 @@ library AmmWithdrawal
         S.exchange.withdrawFromApprovedWithdrawals(owners, tokens);
     }
 
+    // This is to claim other people depositing into this AMM pool.
```
Brechtpd, line 37: I guess same as with `withdrawFromApprovedWithdrawals`, but with the difference that people shouldn't deposit to the pool directly so should never happen. :)

`withdrawFromDepositRequest` can also be called by anyone for this smart contract, so maybe just make this contract `Drainable` for all tokens except the pool token so all funds owned by this smart contract can be withdrawn? Someone else could in theory call `withdrawFromDepositRequest` before this function is used and funds would be stuck in this contract.

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -119,4 +130,22 @@ contract LoopringAmmPool is
...
+    {
+        if (drainer != state.exchange.owner()) {
+            return false;
+        }
+
+        for (uint i = 0; i < state.tokens.length; i++) {
+            if (token == state.tokens[i].addr) {
+                return false;
+            }
+        }
```
Brechtpd, line 148: I think this needs to be like this:
```
// Never allow draining the pool token, these can be stored in this contract for forced exits at all times
if (token == address(this)) {
    return false;
}

// Allow draining the tokens of this pool only when the pool is online,
// when offline those tokens can be stored in this contract for use in withdrawWhenOffline
if (!S.isOnline()) {
    for (uint i = 0; i < state.tokens.length; i++) {
        if (token == state.tokens[i].addr) {
            return false;
        }
    }
}
```

dong77: Based on your description, it should be like this?

```solidity
    {
        if (drainer != state.exchange.owner() || !S.isOnline()) {
            return false;
        }

        // Allow draining the tokens of this pool only when the pool is online,
        // when offline those tokens can be stored in this contract for use in withdrawWhenOffline
        for (uint i = 0; i < state.tokens.length; i++) {
            if (token == state.tokens[i].addr) {
                return false;
            }
        }

        return true;
    }
```

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -119,4 +130,29 @@ contract LoopringAmmPool is
...
+        public
+        override
+        view
+        returns (bool)
+    {
+        if (drainer != state.exchange.owner()) {
+            return false;
+        }
+
+        // When online, owner can claim all tokens, as we do not allow layer-1 joins.
```
Brechtpd, line 144: We can never allow draining the pool token itself here! Even when online, if people do a forced exit, they transfer their pool tokens, the amount they want to burn, to this contract. So still needs this part here in the code above this point:
```
// Never allow draining the pool token, these can be stored in this contract for forced exits at all times
if (token == address(this)) {
    return false;
}
```

Brechtpd (approved):

Issue #1847, opened 2020-10-08
[Loopring 3.6] Bug in AccountUpdate regarding fee and maxFee


dong77:
In our smart contract, there is only this "fee" field in AccountUpdateTransaction. It is supposed to be interpreted as the max fee the user would like to pay. In the circuit 'AccountUpdateCircuit', there is  `fee` and `maxFee` fields, but at the end of the file, we pushed the actual fee we charged onchain as part of DA. this makes the tx hash different and failures of signature verification.

```cpp

    const VariableArrayT getPublicData() const
    {
        return flattenReverse(
          {type.bits,
           owner.bits,
           accountID.bits,
           feeTokenID.bits,
           fFee.bits(),
           compressPublicKey.result(),
           nonce.bits});
    }
```

dong77:
The same issue exists for Tranfers and Withdrawals as well.

dong77:
I suggest that we change "fee" to "maxFee" in contracts so users will notice the nuance when signing the transactions.

Brechtpd:
Explained in https://github.com/Loopring/protocols/issues/1688#issuecomment-686481044 and https://github.com/Loopring/protocols/pull/1689.

We always need at least the actual fee being paid in the DA, because that's the one needed for reconstructing the Merkle tree.

Brechtpd:
It's actually pretty easy to support maxFee, we just put it in the auxiliary data for those transactions, which we already use for `validUntil`. Circuits remain the same.

dong77:
I understand that. Is it possible that we put `maxFee`s in `auxiliaryData` before we calculate transaction hashes in the smart contracts?

Brechtpd:
Yep https://github.com/Loopring/protocols/issues/1847#issuecomment-705600453

dong77:
Great, could you please make that change so the new relayer can make the according changes shortly?

Brechtpd:
Okay, will do this today.

dong77:
@Brechtpd in the current implementation, we keep the `maxFee` value in the transactions auxiliary data. Do you think we should use something called `maxFeeDiff` (= maxFee - fee) so we have a bigger chance of having zeros in calldata?

dong77:
<img width="730" alt="Screen Shot 2020-10-09 at 10 47 16" src="https://user-images.githubusercontent.com/1710087/95536153-d2630980-0a1c-11eb-955c-7155217a7033.png">

[...]

PR #1850, opened 2020-10-08
[protocol 3.6] Support max fee in all authorization methods

Brechtpd:
#1847 

Also removed `approveOffchainTransfer` because we're not even using this ourselves, and easy to use libraries for this are now available.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
+++ b/packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
@@ -21,7 +21,7 @@ library TransferTransaction
     using ExchangeSignatures   for ExchangeData.State;
 
     bytes32 constant public TRANSFER_TYPEHASH = keccak256(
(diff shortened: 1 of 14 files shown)


dong77 (approved):

packages/loopring_v3/contracts/core/impl/libtransactions/TransferTransaction.sol
```diff
@@ -65,9 +67,12 @@ library TransferTransaction
         Transfer memory transfer = readTx(data, offset);
         TransferAuxiliaryData memory auxData = abi.decode(auxiliaryData, (TransferAuxiliaryData));
 
-        // Check validUntil
-        require(ctx.timestamp < auxData.validUntil, "TRANSFER_EXPIRED");
+        // Fill in withdrawal data missing from DA
```
dong77, line 70: I some cases certain fields in TransferAuxiliaryData are not required. For example. The `maxFee` is not required if the transfer is signed off-chain using EdDSA signature. By using `abi.decode(auxiliaryData, (TransferAuxiliaryData));` we have to use 96bits anyway, correct?

PR #1852, opened 2020-10-09
[draft] optimize maxFee in aux data to reduce gas cost

dong77:


packages/loopring_v3/contracts/core/impl/libtransactions/AccountUpdateTransaction.sol
```diff
@@ -43,7 +45,7 @@ library AccountUpdateTransaction
     struct AccountUpdateAuxiliaryData
     {
         bytes  signature;
-        uint96 maxFee;
+        uint16 feeDiscount;
```
dong77, line 48: I'm not sure if we should move `signature` to the end of the struct, will that reduce gas? The same question also apply to the other AuxiliaryData definition.

Brechtpd:
You were too fast with the PR, was still collecting data. :)
 
So the worst case scenario is that all values are non-zero: `16*12 = 192` gas. In most cases however, the max fee will (or we can make it so) be pretty "nice" (as in containing a lot of zeros) so the actual cost will probably be maybe half, but likely even lower. 

Another thing is that `uint16` or `uint96` doesn't matter here, all simple value types will take up 32 bytes regardless of the type. So zeros will already be plentifull, but that's okay, the compression can take care of that. In most cases the number of zeros won't really matter.

Maybe another thing I learned with the whole compression thing is that it's hard to win against plain data. Any math operation that needs to be done costs gas, and more than you would expect. A safe math `add` costs ~75 gas. That's already ~5 non-zero bytes! And around the expected cost in calldata for `maxFee`.

>  accountUpdate.maxFee = accountUpdate.fee.add(uint(auxData.feeDiscount).decodeFloat(16));

The gas cost of this is 475 gas, so ~2.5x the *max* amount calldata cost! But even just a simple add with overflow check already gets us to ~half the max data cost.

So maybe a simple thing like `maxFee = (maxFee == 0) ? fee : maxFee` could save gas in some some select cases, but not sure if even that's worth it.

packages/loopring_v3/contracts/core/impl/libtransactions/WithdrawTransaction.sol
```diff
@@ -64,7 +66,7 @@ library WithdrawTransaction
         uint    minGas;
         address to;
         bytes   extraData;
-        uint96  maxFee;
+        uint16  feeDiscount;
```
dong77, line 69: I think `minGas` and `gasLimit` can also use uint24, the ethereum block's max gas limit is ~10,000,000. uint24's max value is 16777216.

dong77:
> You were too fast with the PR, was still collecting data. :)
> 
> So the worst case scenario is that all values are non-zero: `16*12 = 192` gas. In most cases however, the max fee will (or we can make it so) be pretty "nice" (as in containing a lot of zeros) so the actual cost will probably be maybe half, but likely even lower.
> 
> Another thing is that `uint16` or `uint96` doesn't matter here, all simple value types will take up 32 bytes regardless of the type. So zeros will already be plentifull, but that's okay, the compression can take care of that. In most cases the number of zeros won't really matter.
> 
> Maybe another thing I learned with the whole compression thing is that it's hard to win against plain data. Any math operation that needs to be done costs gas, and more than you would expect. A safe math `add` costs ~75 gas. That's already ~5 non-zero bytes! And around the expected cost in calldata for `maxFee`.
> 
> > accountUpdate.maxFee = accountUpdate.fee.add(uint(auxData.feeDiscount).decodeFloat(16));
> 
> The gas cost of this is 475 gas, so ~2.5x the _max_ amount calldata cost! But even just a simple add with overflow check already gets us to ~half the max data cost.
> 
> So maybe a simple thing like `maxFee = (maxFee == 0) ? fee : maxFee` could save gas in some some select cases, but not sure if even that's worth it.

Good to know this: **it's hard to win against plain data**

dong77:
This is nice: maxFee = (maxFee == 0) ? fee : maxFee

PR #1853, opened 2020-10-09
[Loopring 3.6] minor optimize for maxFee aux data

dong77:


Brechtpd:
So surprisingly this also costs ~50 gas... so around ~3 non-zero bytes. But still seems reasonable to do.

Brechtpd (approved):

PR #1856, opened 2020-10-09
[AMM] onchain join/exit requests + multiple small changes that have been reviewed already.

dong77:


packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
```diff
@@ -32,6 +34,30 @@ library AmmWithdrawal
...
+        for (uint i = 0; i < tokens.length; i++) {
+            address token = tokens[i];
+            S.exchange.withdrawFromDepositRequest(address(this), token);
+            if (token != address(this)) {
+                uint balance = token == address(0) ?
+                    address(this).balance :
+                    ERC20(token).balanceOf(address(this));
+
+                AmmUtil.transferOut(token, balance, recipient);
+            }
```
Brechtpd, line 57: Seems like I missed this previously, but because this will transfer out the full balance (not just what was withdrawn from the deposit) we have to implement the same logic here as in `canDrain`, we cannot allow this for tokens of the pool when the pool is offline.

So perhaps we should just call `canDrain` here, or similar to `withdrawFromApprovedWithdrawals` don't actually transfer any tokens out here, and leave that to `drain`.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -41,7 +41,14 @@ library AmmJoinProcess
         require(join.validUntil >= block.timestamp, "EXPIRED");
 
         bytes32 txHash = AmmJoinRequest.hash(ctx.domainSeparator, join);
-        require(txHash.verifySignature(join.owner, signature), "INVALID_JOIN_APPROVAL");
+
+        if (signature.length == 0) {
+            require(join.validUntil >= block.timestamp, "JOIN_NOT_FOUND_OR_EXPIRED");
```
Brechtpd, line 46: Same check already done above on line 41, so will never trigger here.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -112,5 +112,6 @@ library AmmData
 
         // A map from a user to the forced exit.
         mapping (address => PoolExit) forcedExit;
+        mapping (bytes32 => uint) approvedTx;
```
Brechtpd, line 115: Not that it matters I think, but I guess this could just be `=> bool` because we don't actually use the timestamp now.

Brechtpd:
@dong77 Updated tests for new authorization methods and fixed some small issues.

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
dong77: changed to use bool

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: fixed

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawal.sol
dong77: Brecht, I think we can delete this function and `withdrawFromApprovedWithdrawals` and the Drainable interface. Because in all our expected use cases, there will be no onchain deposit or withdrawals. What do you think?

Brechtpd: Okay for me, they would indeed only be useful in case people would use the pool incorrectly. And `withdrawFromApprovedWithdrawals` is just a helper function.

Brechtpd (approved):

PR #1857, opened 2020-10-09
[AMM] Optimizations

Brechtpd:
#1827, but re-doing optimizations step by step, measuring gas for each step along the way and tweaking/changing/adding things where necessary.

Joins/exits now cost ~175,000 gas (2 token pools). Only did minimal changes and reverted some smaller optimizations I did. Did make sure that larger blocks don't increase the gas cost for joins/exits, which was my main concern because of data copying.

Tested reading all the static data once and reuse it between calls:
```
struct StaticContext
{
    // Exchange state variables
    IExchangeV3 exchange;
    bytes32     exchangeDomainSeparator;

    // AMM pool state variables
    bytes32 domainSeparator;
    uint32  accountID;

    uint16  poolTokenID;

    Token[]  tokens;
}
```

This costs ~5,000 gas *more* for 2 joins, no difference for 4 joins, and saves ~14,000 gas for 8 joins (these are *total* gas consumption differences, not for each join. So for 8 joins it's `14,000/8 = 1,750` cheaper per join). I don't think that's really worth the extra complexity, but I can push this if this is deemed a useful optimization.

Maybe one thing we should keep in mind is that we shouldn't use `AutoUpgradabilityProxy`, it's just too expensive for something that probably doesn't get used. This makes each call to the exchange contract cost more than 10,000 gas because of all the overhead. For `ManualUpgradabilityProxy` each call costs around ~4,000 gas, which is much more acceptable.

But even when using `ManualUpgradabilityProxy` it may still be worth it to cache the exchange owner in the pool contract (so we don't have to do `state.exchange.owner()` each time, we need to check this for each call because we can't trust any of the data otherwise). This saves ~5,000 gas per join/exit, but then we do have to worry about updating each pool when the exchange owner changes.


dong77:
1. I think we can cache the exchange owner and provide a permissionless function to allow anyone to update the cache.
2. Regarding the proxy, I agree, maybe we should get rid of IUniversalRegistry entirely (I'll propose a PR)

packages/loopring_v3/contracts/core/iface/IAgentRegistry.sol
```diff
@@ -17,4 +17,16 @@ interface IAgentRegistry
         external
         view
         returns (bool);
+
+    /// @dev Returns whether an agent address is an agent of an account owner
```
dong77, line 21: Returns whether an agent address is an agent of all accounts

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -56,37 +58,43 @@ library AmmBlockReceiver
...
             accountID: S.accountID,
             poolTokenID: S.poolTokenID,
             totalSupply: S._totalSupply,
-            size: size,
             tokens: S.tokens,
-            tokenBalancesL2: new uint96[](size)
+            tokenBalancesL2: new uint96[](size),
+            transactionBuffer: AmmData.TransactionBuffer({
+                size: 0,
+                owners: new address[](size*2 + 1),
```
dong77, line 72: size  + 2?
For each token, we need one transfer (+size), and we also need a pool token transfer (+1) and an AmmUpdate (+1)

Brechtpd: We need an AmmUpdate for each token (have to get the balance for each token separately). So I think `size * (transfer + update) + burn/mint`.

dong77: I see.

dong77 (approved):

dong77 (approved):

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -30,6 +33,7 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
     struct TxCallback
     {
         uint16 txIdx;
+        uint16 numTxs;
```
dong77, line 36: 中继请注意这个修改。

PR #1873, opened 2020-10-12
[Loopring AMM] enable exit fees and remove join fees

dong77:
Charge a fixed amount of exit fee using the last token.

### DIFF
--- a/packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
+++ b/packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
@@ -82,7 +82,7 @@ library AmmExitProcess
                 transfer.from == address(this) &&
                 transfer.to == exit.owner &&
                 transfer.tokenID == ctx.tokens[i].tokenID &&
-                transfer.amount.isAlmostEqualAmount(amounts[i]),
+                transfer.amount.add(transfer.fee).isAlmostEqualAmount(amounts[i]),
                 "INVALID_TX_DATA"
             );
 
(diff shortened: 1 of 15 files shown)


packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -82,12 +82,19 @@ library AmmExitProcess
...
             );
 
+            if (transfer.fee > 0) {
+                require(
+                    i == ctx.size - 1 &&
+                    transfer.fee.isAlmostEqualAmount(exit.fee) &&
+                    transfer.feeTokenID == ctx.tokens[i].tokenID,
+                    "INVALID_FEES"
+                );
+            }
```
Brechtpd, line 96: - Should do `(transfer.fee + transfer.amount) == amounts[i]`, otherwise it's the pool paying the exit fee. 
- `isAlmostEqualAmount` -> `isAlmostEqualFee`.

I also think this is still a very good option as well:
- Charge a fee on the burn transfer, so we can really just charge a fee to the user without having to subtract it here from the amount (but then we should move the burn transfer after the user received his extra funds so the fee can always be paid from those funds).
- Add a `feeTokenId` to the exit so a fee can be paid in any of the tokens being exited. I guess we can always make it so that the last token is the one we want the fee to be paid in like a stable coin/ETH, but maybe still a good option to have.

dong77: Please check if the new code is correct.

dong77: I want to save gas without an extra feeTokenID field. We'll probably not going to use the extra options anyway...

dong77: Regarding `(transfer.fee + transfer.amount) == amounts[i]`, I think this really depends on what `exitMinAmounts` really means. In this PR, I added a comment so `exitMinAmounts` means the min amounts BEFORE paying the fee. But even with this, I think it's still fine to use `transfer.amount.add(transfer.fee).isAlmostEqualAmount(amounts[i])` instead of the strict `==` sign.

Brechtpd (approved):

PR #1887, opened 2020-10-15
[loopring36]update migration script for a dev docker image

kongliangzhong:


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -105,34 +105,6 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
         );
     }
 
-    function cloneExchange(
```
Brechtpd, line 108: So we're going to put the exchange behind a proxy somewhere else?

kongliangzhong: I will deploy the proxy contract manually, there is no need to put the code in the contract.

dong77 (approved):

dong77 (approved):

PR #1891, opened 2020-10-16
[hebao] Optimizations

Brechtpd:
Probably not all changes we want to do (modifications to stores), but I'm just getting it as low as possible for now with reasonable changes.


dong77:
I suggest using ETH transfer and LRC transfer's meta-tx gas consumptions as the bench mark.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -26,15 +26,14 @@ abstract contract ForwarderModule is BaseModule
...
 
+    mapping(address => uint) public nonces;
+
     event MetaTxExecuted(
         address relayer,
         address from,
         uint    nonce,
         bytes32 txAwareHash,
         bool    success,
-        address gasToken,
```
dong77, line 35: @lyddream @dyfvicture 这几个数据项被移除会有什么影响吗？

packages/hebao_v1/contracts/modules/transfers/TransferModule.sol
```diff
@@ -103,8 +103,8 @@ abstract contract TransferModule is BaseTransferModule
         bytes calldata logdata
         )
         external
-        nonReentrant
-        txAwareHashNotAllowed()
```
dong77, line 107: this will cause a security issue I believe.

packages/hebao_v1/contracts/modules/base/BaseModule.sol
```diff
@@ -216,9 +244,10 @@ abstract contract BaseModule is ReentrancyGuard, Module
         uint gasCost = gasAmount.mul(gasPrice);
 
         if (!skipQuota) {
-            uint value = controller().priceOracle().tokenValue(gasToken, gasCost);
+            ControllerImpl _controller = controller();
+            uint value = _controller.priceOracle().tokenValue(gasToken, gasCost);
```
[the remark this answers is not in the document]
Brechtpd: Yep! Currently not part of the benchmark and did not yet replace all cases where it should be used elsewhere.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -209,13 +205,8 @@ abstract contract BaseWallet is ReentrancyGuard, Wallet
...
         returns (bytes memory returnData)
     {
-        require(
-            !controller.moduleRegistry().isModuleRegistered(to),
-            "TRANSACT_ON_MODULE_DISALLOWED"
-        );
-
         bool success;
-        (success, returnData) = nonReentrantCall(mode, to, value, data);
+        (success, returnData) = _call(mode, to, value, data);
```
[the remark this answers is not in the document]
Brechtpd: I don't know how reentrancy can be a problem here. No state is modified (or even read except authorization of the module) in this function so I don't know how it could be abused.

packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -209,13 +205,8 @@ abstract contract BaseWallet is ReentrancyGuard, Wallet
         onlyFromFactoryOrModule
         returns (bytes memory returnData)
     {
-        require(
```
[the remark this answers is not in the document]
Brechtpd: This check is quite expensive to do for each call while mostly we are just calling contracts in modules that are fixed or are checked before they even get here (which makes sense).

I think only an arbitrary `to` is allowed for contract calls and token transfers/approvals:
- For contract calls we do extra checks on `to` in `callContractInternal`
- For token transfers/approvals we don't do any check, but only a specific function can be called. Unless modules/wallets implement those ERC20 functions (which is not impossible, just not really realistic) or don't do anything in the `fallback` function which could be abused I think we should be fine.

But for sure should be looked into for closely if we remove this check. Unfortunately having additional checks for things are good but certainly not free...

packages/hebao_v1/contracts/modules/transfers/TransferModule.sol
Brechtpd: - Removing `txAwareHashNotAllowed` should be fine everywhere: https://github.com/Loopring/protocols/pull/1444#issuecomment-661828691 . It's not like removing this suddenly saves 10,000 gas or something, but removing a lot of these "not really necessary" things does add up.
- Removing `nonReentrant` here should also be fine here I believe. This function doesn't really do much, the only state change that's possible is the quota that's being updated, which is done before the external call is made so I don't think it's necessary.

packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
```diff
@@ -30,7 +30,7 @@ abstract contract MetaTxAware
 
     modifier txAwareHashNotAllowed()
     {
-        require(txAwareHash() == 0, "INVALID_TX_AWARE_HASH");
+        //require(txAwareHash() == 0, "INVALID_TX_AWARE_HASH");
```
dong77, line 33: I think this cannot be removed.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dyfvicture: 我们只要更新下abi 这几个值在系统里没有用

packages/hebao_v1/contracts/base/BaseWallet.sol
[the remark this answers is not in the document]
dong77: It seems you are right, there will be reentrancy, but they will not cause any issues.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -103,27 +101,53 @@ abstract contract ForwarderModule is BaseModule
...
 
         bytes32 metaTxHash = EIP712.hashPacked(FORWARDER_DOMAIN_SEPARATOR, encoded);
-        require(metaTxHash.verifySignature(from, signature), "INVALID_SIGNATURE");
+
+        // Instead of always taking the expensive path through ER1271,
+        // skip directly to the wallet owner here (which could still be another contract).
+        //require(metaTxHash.verifySignature(from, signature), "INVALID_SIGNATURE");
+        (uint _lock,) = controllerCache.securityStore.getLock(from);
+        require(_lock <= block.timestamp, "WALLET_LOCKED");
+        require(metaTxHash.verifySignature(Wallet(from).owner(), signature), "INVALID_SIGNATURE");
```
letsgoustc, line 110: In ERC1271 module, verify signature is also to check whether it's signed by owner or not. Looks only some extra function call is included. How many gas will it influence with or without the optimization?

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -137,17 +161,12 @@ abstract contract ForwarderModule is BaseModule
             metaTx.gasToken,
             metaTx.gasPrice,
             metaTx.gasLimit,
-            metaTx.data,
+            data,
             signature
         );
 
-        // Nonce update must come after the real transaction in case of new wallet creation.
-        if (metaTx.nonce != 0) {
```
letsgoustc, line 145: The original comment points out the verify must come after the real transaction. Is it safe to move it ahead?

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -26,15 +26,14 @@ abstract contract ForwarderModule is BaseModule
     uint    public constant MAX_REIMBURSTMENT_OVERHEAD = 165000;
     bytes32 public FORWARDER_DOMAIN_SEPARATOR;
 
+    mapping(address => uint) public nonces;
```
letsgoustc, line 29: I think all the data should be stored in some DataStore for future upgrading?

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: This saves 11,000 gas. Calls on the wallet are extra expensive because they have to go through the proxy contract, so there are additional CALL and SLOAD costs. There's also some other savings like not having to check twice in the signature code that the signer isn't an address etc...

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: No interaction necessary with the store any more which needed a valid wallet, so it shouldn't matter any more where we update the nonce for that.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: Yes, but nonces aren't especially something that need to be kept between upgrades. If the user updates the forwarder module the nonces restart from `0` (with a different `FORWARDER_DOMAIN_SEPARATOR` value) which should be okay. The user doesn't care about the nonce value, unlike things like guardians, quota, etc...

packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
Brechtpd: > I knew this was possible, but I don't see it as a security issue. It's the user that signs the wrong data. We can't really guard against the user signing the wrong things.

> That's why we sign the function selector for the meta tx so it's the user that remains in control and the relayer can't mess things up agains the users. If we check this everywhere (quite error prone because easy to forget) I don't think we even have to sign the function selector like we do now.


If the function that was called actually uses the `txAwareHash` than replay protection is provided `verifyAndUpdate`. 

But if the user signed a meta transaction with `nonce == 0 && txAwareHash != 0` for a function call that doesn't support `txAwareHash` to begin with than you're right that the meta tx can be reused without the check. But why did the user sign an invalid meta tx to begin with, the meta tx would never be valid with the check, not even the first time?

But if you think we need to guard the user from signing an invalid meta tx we can of course keep it, but if the user only signs valid meta txs than I think there is no problem.

Brechtpd: The extra cost is only 1,200 gas, mainly because of the SLOAD of `trustedForwarder`, which is actually already read by `onlyFromWalletOrOwnerWhenUnlocked` so could be reduced to lower the cost to 400 gas, but it's a bit awkward to cache `trustedForwarder` between the modifiers.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: Still 8,000 gas cheaper after #1905.

packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
dong77: Lets keep it please.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
kongliangzhong: OK, so we should set a different value to FORWARDER_DOMAIN_SEPARATOR when we upgrading the ForwarderModule. Otherwise a replay-attach becomes possible.

packages/hebao_v1/contracts/modules/base/MetaTxAware.sol
dong77: I reverted the change to txAwareHashNotAllowed

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: Done automatically because of EIP712, the contract address is always part of the data hashed to FORWARDER_DOMAIN_SEPARATOR.

packages/hebao_v1/contracts/modules/security/SecurityModule.sol
```diff
@@ -41,17 +41,14 @@ abstract contract SecurityModule is MetaTxModule
             (_logicalSender == Wallet(wallet).owner() && !isWalletLocked(wallet)),
              "NOT_FROM_WALLET_OR_OWNER_OR_WALLET_LOCKED"
         );
-        SecurityStore ss = controller().securityStore();
-        if (block.timestamp > ss.lastActive(wallet) + MIN_TOUCH_INTERVAL) {
```
dong77, line 45: Brecht, we should try to keep using the same SecurityStore as it is the most important one (WhitelistStore and QuotaStore can be reset for all test users). How does this change impact gas usage, I am thinking to revert this change (line45-47) if possible.

dong77: @Brechtpd

Brechtpd: Okay, the only difference is that when the last active actually changes it costs 2,000 gas more. If the timestamp doesn't need to be updated the gas cost remains the same.

dong77: Let us keep it as is :) A simple and effective change is to use 2 weeks as the MIN_TOUCH_INTERVAL value. What do you think?

PR #1896, opened 2020-10-20
Improve reentrance check

dong77:


packages/hebao_v1/contracts/base/BaseWallet.sol
```diff
@@ -16,7 +15,7 @@ import "./Controller.sol";
 ///
 /// The design of this contract is inspired by Argent's contract codebase:
 /// https://github.com/argentlabs/argent-contracts
-abstract contract BaseWallet is ReentrancyGuard, Wallet
+abstract contract BaseWallet is Wallet
```
Brechtpd, line 18: I think removing `ReentrancyGuard` here will change the storage layout of the contract (`_guardValue` storage variable in `ReentrancyGuard`). Not sure if that's still a problem, but there's a warning below so probably not compatible with the first wallets that were upgradeable.

dong77: Good catch, I'll revert related changes.

Brechtpd (approved):

PR #1899, opened 2020-10-20
Fix a bug?

dong77:
@kongliangzhong @Brechtpd I think this is a bug, please help confirm.


kongliangzhong:
onlyFromWalletOrOwnerWhenUnlocked(wallet), 大部分的函数都有这个约束， 所以这不算一个bug。

Brechtpd:
I also think it's fine, like the comment above it says, I think the wallet cannot be the sender without having already gone through a check that the wallet isn't locked before checking the signature (`ERC1271Module` or `ForwarderModule`).

PR #1905, opened 2020-10-20
[hebao] Deploy optimizations

Brechtpd:
- Uses the optimized proxy code from https://eips.ethereum.org/EIPS/eip-1167. Like SimpleProxy it doesn't allow the wallet to be upgraded. Etherscan should also detect this as a proxy contract and should show the proxy interface, though I did not test this.
- Wallet `init` function now also can set modules directly instead of needing multiple calls on the wallet

When running the "anyone should be able to create a group of blank wallets" test which creates 10 wallets, with the 3 common modules initialized, the deployment cost is reduced from 3,649,674 gas to 2,103,030 gas, a 42% reduction. 35% of that 42% reduction is because of the much smaller deployment cost of the proxy contract.

This also reduces the cost of using the wallet because the implementation address is baked into the proxy code, instead of having to SLOAD the implementation address each time for each call. Because of that, transfers are now also 3,000 to 4,000 gas cheaper (~3%).

### DIFF
--- a/packages/hebao_v1/contracts/modules/core/WalletFactory.sol
+++ b/packages/hebao_v1/contracts/modules/core/WalletFactory.sol
@@ -80,8 +80,9 @@ contract WalletFactory is MetaTxAware
         external
         txAwareHashNotAllowed()
     {
+        address _walletImplementation = walletImplementation;
         for (uint i = 0; i < salts.length; i++) {
-            createBlank_(walletImplementation, modules, salts[i]);
+            createBlank_(_walletImplementation, modules, salts[i]);
         }
     }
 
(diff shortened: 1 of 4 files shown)


packages/hebao_v1/contracts/modules/core/WalletFactory.sol
```diff
@@ -80,8 +80,9 @@ contract WalletFactory is MetaTxAware
         external
         txAwareHashNotAllowed()
     {
+        address _walletImplementation = walletImplementation;
```
dong77, line 83: I tested it and I am surprised the optimizer doesn't even optimize this one.

dong77 (approved):

packages/hebao_v1/contracts/modules/core/WalletFactory.sol
Brechtpd: The optimizer certainly isn't as good as you may expect from optimizers in other languages.

PR #1906, opened 2020-10-20
[protocol 3.6] Better optimizer settings

Brechtpd:
Just increases the `runs` parameter of the optimizer so it optimizes more for runtime efficiency than contract size. Saves ~1% gas on the operations I checked (deposits, AMM joins).

@kongliangzhong Is there a reason there is a `LoopringAmmPoolCopy`? Not possible to just call `LoopringAmmPool.new()` to create another copy?

### DIFF
--- a/packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
+++ b/packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
@@ -36,7 +36,7 @@ library AmmStatus
         AmmData.State      storage  S,
         AmmData.PoolConfig calldata config
         )
-        internal
+        public
     {
         require(
             bytes(config.poolName).length > 0 && bytes(config.tokenSymbol).length > 0,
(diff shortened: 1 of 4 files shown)


dong77:
I was going to do this as well because I checked that Argent uses 999 runs. I wonder if 1000000 is too large. What about 1000 or 10000?

dong77 (approved):

dong77:
> @kongliangzhong Is there a reason there is a LoopringAmmPoolCopy? Not possible to just call LoopringAmmPool.new() to create another copy?

I forward this question to Kongliang

dong77 (approved):

Brechtpd:
> I was going to do this as well because I checked that Argent uses 999 runs. I wonder if 1000000 is too large. What about 1000 or 10000?

Tested it with a couple of different values with the wallet. Changes are small, but more is always better, as you would expect. I'm not sure if there is any internal max value, but 1,000,000 is better than 100,000, did not see any improvements with 10,000,000 so kept it at 1,000,000, but I think it just needs to be a very large value.

PR #1911, opened 2020-10-21
[hebao] Small optimizations wallet creation

Brechtpd:


packages/hebao_v1/contracts/modules/core/WalletFactory.sol
```diff
@@ -298,10 +319,6 @@ contract WalletFactory is MetaTxAware
         );
 
         bytes32 signHash = EIP712.hashPacked(DOMAIN_SEPERATOR, encodedRequest);
-
-        bytes32 txAwareHash_ = txAwareHash();
-        require(txAwareHash_ == 0 || txAwareHash_ == signHash, "INVALID_TX_AWARE_HASH");
```
Brechtpd, line 303: I don't think we need to check this. In the worst case where the user would sign a meta tx with a txAwareHash (though that wouldn't make much sense to begin with) the call will fail when replayed any way because a wallet can only be created once.

Brechtpd:
https://github.com/Loopring/protocols/blob/1c192dfa12664a362827c40b32a1cb78e4b87663/packages/hebao_v1/contracts/modules/core/WalletFactory.sol#L47

Do we still want to always use ENS? This is now ~50% of the cost of creating a wallet.

dong77:
> https://github.com/Loopring/protocols/blob/1c192dfa12664a362827c40b32a1cb78e4b87663/packages/hebao_v1/contracts/modules/core/WalletFactory.sol#L47
> 
> Do we still want to always use ENS? This is now ~50% of the cost of creating a wallet.

Yes, that's one of the main features ...

packages/hebao_v1/contracts/modules/core/WalletFactory.sol
dong77: Ok, but please comment on this `require` and add a comment so later we don't forget why we removed this logic.

dong77 (approved):

PR #1912, opened 2020-10-21
[hebao] Remove wallet registry?

Brechtpd:
Doesn't seem to be very useful, and increases the cost of creating a wallet by 30,000 gas (5%-10% of total cost).

Only used in 2 places, added comments there why I think the use can be removed there.


packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -74,8 +74,7 @@ abstract contract ForwarderModule is BaseModule
 
             // We only allow the wallet to call itself to addModule
             (to == from) &&
-            data.toBytes4(0) == Wallet.addModule.selector &&
-            controllerCache.walletRegistry.isWalletRegistered(from) ||
+            data.toBytes4(0) == Wallet.addModule.selector ||
```
Brechtpd, line 77: Unnecessary check I think. If `from` isn't a wallet than that should be okay, `addModule` will be called on this arbitrary (but similar to a Wallet) contract which is fine.

packages/hebao_v1/contracts/modules/security/InheritanceModule.sol
```diff
@@ -74,7 +74,6 @@ abstract contract InheritanceModule is SecurityModule
         txAwareHashNotAllowed()
         onlyFromWalletOrOwnerWhenUnlocked(wallet)
     {
-        require(controllerCache.walletRegistry.isWalletRegistered(_inheritor), "NOT_A_WALLET");
```
Brechtpd, line 77: Not really necessary to enforce this onchain. We can enforce rules for inheritors in the app. If someone manually wants to set an arbitrary address as an inheritor that's also fine, but they'll have to do that by directly interacting with the contracts, and so if they can do that they can also do all other operations they want to do manually themselves as well.

dong77 (approved):

PR #1913, opened 2020-10-22
[hebao] A few behavior changes and important bug fixes

dong77:
- Change to use constant values instead of state variables for many configurations
- Change min guardian count to 1.
- Change lock period from 1 day to 3 days, change guardian addition/removal pending period from 1 day to 3 days. If a wallet owner's address is leaked, the owner now has more time to react.
- Fix a bug that we didn't check the number of guardians who signed transactions
- Allow guardian-approved transactions even if the wallet is locked.
- add `addGuardianImmediately` , `removeGuardianImmediately` and `removeFromWhitelistImmediately` (not tested yet)


packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -87,8 +84,54 @@ abstract contract GuardianModule is SecurityModule
...
+        controller().verifyRequest(
+            GUARDIAN_DOMAIN_SEPERATOR,
+            txAwareHash(),
+            GuardianUtils.SigRequirement.OwnerRequired,
+            request,
+            abi.encode(
+                ADD_GUARDIAN_IMMEDIATELY_TYPEHASH,
+                request.wallet,
+                request.validUntil,
+                guardian
```
Brechtpd, line 107: Add `group`

dong77: done

Brechtpd (approved):

Brechtpd:
> Fix a bug that we didn't check the number of guardians who signed transactions

Maybe not a bug, but a feature. :) If a wallet doesn't have any guardians than I don't think any of the delayed operations have any effect so doesn't really matter if operations can be done immediately or not. And with 1 guardian I think both the both owner and guardian are necessary depending on the guardian group, so probably a small change in behaviour here.

PR #1931, opened 2020-10-22
[hebao 1.2] add multiple breaking changes 

dong77:
A very important change of behavior is that once a wallet is locked, it should be as secure as possible within 7 days unless the majority of its guardians collude. (Currently, the lock will be automatically removed after 1 day)

- We remove the `onlyHaveEnoughGuardians` modifier; make sure `requireMajority` checks the number of active guardians is > 0
- The first two guardians still don't have to wait a pending period
- If a function `doSomething` has a version that supports multi-sig authorization, I name it `doSomethingWA` where *WA* means *WithApproval*. I also make sure these two functions are next to each other. If a function only supports multisig, *WA* will be dropped from the function name
- Add default impl for `bindableMethods` in BaseModule
- Allow owner to lock his own wallet (regardless of the current lock status)
- Disable automatic unlock after 1 day. Do not track who locked the wallet.
- Lock and unlock does not check the previous status and will always succeed.
- Change the max number of guardians from 20 to 10, change guardian waiting period from 1 day to 7 days
- Disable unlock by one single guardian and only allow unlock with a majority (owner signature required)
- Event `WalletLock` now includes which guardian locked a wallet and which module unlocked a wallet.
- Allow wallet owner to specify the inheritance waiting period (2 months to 10 years)
- Remove `cancelGuardianAddition` and `cancelGuardianRemoval`. These will be part of the `removeGuardian` and `addGuardian` behavior
- Rename event 'WalletLock' to 'WalletLocked'
- Add 'lockWA' that doesn't use nonce
- Daily quota is by default disabled and can be set to non-zero to enable quoto checking/update.
- Last touch timestamp will not be checked/updated if no inheritor is set.

##TODO
- [ ] Test new InheritanceModule
- [ ] Test new lock/unlock logics
- [ ] Test the new  `removeGuardian` and `addGuardian` behavior.
- [ ] Test the new  majority logic


packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -26,6 +26,10 @@ abstract contract ForwarderModule is BaseModule
     uint    public constant MAX_REIMBURSTMENT_OVERHEAD = 165000;
     bytes32 public FORWARDER_DOMAIN_SEPARATOR;
 
+    bytes32 constant public META_TX_TYPEHASH = keccak256(
+        "MetaTx(address from,address to,uint256 nonce,bytes32 txAwareHash,address gasToken,uint256 gasPrice,uint256 gasLimit,bytes data)"
```
dong77, line 30: nothing changed, moved code position

Brechtpd:
> Add `addGuardiansWithTheirApproval` this is to get around nonce-issue when we need a signature from new guardians before adding them.

Not sure I understand this. Why would a guardian need to sign something when being added to a wallet? Or do you mean a nonce issue for the wallet/wallet owner? Not sure why adding a guardian would have that problem because the meta tx can always be done immediately like any other meta tx. But if it would be a problem maybe a good reason to add back replay protection with hashes on the meta tx level instead trying to use the `txAwareHash` system for that which isn't really made for that.

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -134,33 +157,35 @@ abstract contract GuardianModule is SecurityModule
...
-        txAwareHashNotAllowed()
-        onlyFromGuardian(wallet)
     {
-        unlockWallet(wallet, false);
+        controller().verifyRequest(
+            GUARDIAN_DOMAIN_SEPERATOR,
+            txAwareHash(),
+            GuardianUtils.SigRequirement.OwnerRequired,
+            request,
+            abi.encode(UNLOCK_TYPEHASH)
```
Brechtpd, line 185: Add `request.wallet` and `request.validUntil`.

packages/hebao_v1/contracts/modules/security/GuardianModule.sol
```diff
@@ -62,45 +58,61 @@ abstract contract GuardianModule is SecurityModule
...
         address guardian,
         uint    group
         )
         external
-        onlyHaveEnoughGuardians(request.wallet)
     {
         controller().verifyRequest(
             GUARDIAN_DOMAIN_SEPERATOR,
             txAwareHash(),
             GuardianUtils.SigRequirement.OwnerAllowed,
```
Brechtpd, line 112: I still find it a bit strange that guardians can add/remove guardians without the owner, but I don't think there's anything wrong with it. Only that maybe guardians can sneakily add additional guardians without the owner knowing (when e.g. funds in the wallet are still low, so no reason to immediately take over the wallet), so if the owner isn't careful he can get into state where the guardians are differently than he expected in the future.

kongliangzhong: I think it is OK. the majority of guardians can recover the wallet, so they can do whatever else.

dong77: One improvement can be that before guardians collectively remove/add another guardian, the wallet must be locked first. How do you like the idea?

dong77:
> > Add `addGuardiansWithTheirApproval` this is to get around nonce-issue when we need a signature from new guardians before adding them.
> 
> Not sure I understand this. Why would a guardian need to sign something when being added to a wallet? Or do you mean a nonce issue for the wallet/wallet owner? Not sure why adding a guardian would have that problem because the meta tx can always be done immediately like any other meta tx. But if it would be a problem may be a good reason to add back replay protection with hashes on the meta tx level instead of trying to use the `txAwareHash` system for that which isn't really made for that.

Good challenge. We ended up deciding to remove the signing process.

packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
```diff
@@ -47,8 +47,7 @@ abstract contract ERC1271Module is ERC1271, BaseModule
         returns (bytes4 magicValue)
     {
         address wallet = msg.sender;
-        (uint _lock,) = controllerCache.securityStore.getLock(wallet);
-        if (_lock > block.timestamp) { // wallet locked
+        if (controllerCache.securityStore.isLocked(wallet)) {
```
letsgoustc, line 50: Do we need to add back the old Api isValidSignature(bytes hash, bytes signature) in case of interoperability issue with old app.

dong77: Lets not. It's too eary to priority backward compatibility.

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -35,105 +35,71 @@ library GuardianUtils
         }
 
         // Calculate total group sizes
-        Data.Guardian[] memory allGuardians = securityStore.guardians(wallet);
-        uint[MAX_NUM_GROUPS] memory total = countGuardians(allGuardians);
+        Data.Guardian[] memory allGuardians = securityStore.guardians(wallet, false);
+        require(allGuardians.length > 0, "NO_GUARDIANS");
```
adamchen0510, line 39: There has no limit when remove guardian and all guardians can be removed, if then, the exceed quota transfer would be fail here.

Issue #1924, opened 2020-10-22
[hebao 1.2] remove guardian group support & improve majority calculation


dong77:

dong77:
Let's suppose M is the total number of active guardians, M can be an even number (`2N`) or an odd number (`2N+1`). Now let's calculate how many guardians (excluding the owner) need to sign a transaction to achieve a majority approval.

Owner Signature Required  | Note | Guardians Signature Required (when M=2N) | Guardians Signature Required (when M=2N+1)
------------- | ------------- | ------------- | -------------
Yes  | Owner will be part of the **extended guardian group**, the count is M+1 | **(M+1)/2+1** *or N+1* | **(M+1)/2+1** *or N+2*
Content Cell  | Content Cell | Content Celll | Content Cell

dong77:
https://github.com/Loopring/protocols/pull/1938

PR #1938, opened 2020-10-23
[Hebao1.2]Remove guardian group

kongliangzhong:


packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -53,66 +48,43 @@ library GuardianUtils
...
                 walletOwnerSigned = true;
             } else {
-                require(securityStore.isGuardian(wallet, signers[i]), "SIGNER_NOT_GUARDIAN");
-                signingGuardians[numGuardians++] = securityStore.getGuardian(wallet, signers[i]);
+                require(isWalletGuardian(allGuardians, signers[i]), "SIGNER_NOT_GUARDIAN");
             }
         }
 
+        if (walletOwnerSigned) {
+            totalNumVotes += 1;
```
Brechtpd, line 56: I think we need to do this instead for `totalNumVotes`: 
```
if (requirement != SigRequirement.OwnerNotAllowed) {
    totalNumVotes += 1;
}
```

kongliangzhong: I think owner should not included in totalNumVotes if it not in the signers. In this way we need less guardian signers in some case.

Brechtpd: Certainly not a big difference, only a difference when `OwnerAllowed` is used. With the current code, for example if a wallet only has a single guardian and `OwnerAllowed` is enabled, only that singly guardian can sign (and the wallet owner can also sign, but then that single guardian also still needs to sign because then suddenly 2 signatures are required so no reason to do that). Maybe a bit unexpected.

kongliangzhong: If a wallet only has a single guardian, then that guardian can do whatever he want to the wallet, so nothing unexpected. :)

Brechtpd: Oh sure, but similar cases with a different number of guardians.

I guess what I find unexpected is that there's no fixed number of signatures required and the signature of the wallet owner is somehow worth less than a signature of a guardian. I'm unsure if that makes sense or not.

On the other hand, we don't really use `OwnerAllowed` any more I think so it doesn't really matter now. :)

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -65,76 +57,40 @@ library GuardianUtils
...
+        uint numExtendedSigners = allGuardians.length;
+        uint numGuardianApprovals = signers.length;
 
-        // Count the number of votes
-        uint totalNumVotes = 0;
-        uint numVotes = 0;
-        if (requirement != SigRequirement.OwnerNotAllowed) {
-            totalNumVotes += 1;
-            numVotes += walletOwnerSigned ? 1 : 0;
+        if (walletOwnerSigned) {
```
dong77, line 61: @Brechtpd please review and verify the logic is correct : https://github.com/Loopring/protocols/issues/1924

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
dong77: `OwnerAllowed` is not really used but I feel like have it for future usage :)

dong77: @Brechtpd I have improved the majority calculation code based on the tables in https://github.com/Loopring/protocols/issues/1924, please take another look at this PR.

kongliangzhong:
Ok for me

PR #1939, opened 2020-10-23
[loopring36] Add type in TransactionApproved event

kongliangzhong:


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -548,30 +548,36 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
...
+        ExchangeData.ApprovalType approvalType
         )
         external
         override
         nonReentrant
         onlyFromUserOrAgent(owner)
     {
         state.approvedTx[owner][transactionHash] = true;
-        emit TransactionApproved(owner, transactionHash);
+        emit TransactionApproved(owner, transactionHash, approvalType);
```
Brechtpd, line 560: There's no way to check if the `approvalType` value specified is actually the expected value, so I don't really see how it can be used in a reliable way like this.

Brechtpd:
Is this to filter out requests made by the pool contracts in the relayer? Maybe using the transaction hash is better? I think that value is easily available in the event, so if the event was done by an Ethereum tx by the relayer it can be removed or somethink like that?

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
```diff
@@ -28,6 +28,12 @@ library ExchangeData
         AMM_UPDATE
     }
 
+    enum ApprovalType
+    {
+     USER,
+     AMM_POOL
```
dong77, line 34: Exchange should not know anything about AMM, what if we have another application, shall we modify the exchange contract with a new ApprovalType? No!

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
```diff
@@ -85,7 +85,8 @@ abstract contract IExchangeV3 is Claimable
 
     event TransactionApproved(
         address owner,
-        bytes32 transactionHash
+        bytes32 transactionHash,
+        ExchangeData.ApprovalType approvalType
```
dong77, line 89: Maybe we can emit msg.sender here and use it to check where the approval came from.

kongliangzhong:
Ok, @lyddream will filter the event by the methodId of the tx. A event is processed only if the tx methodId is approveTransaction or approveTransactions.

kongliangzhong:
I will close this PR.

PR #1940, opened 2020-10-23
Improve security store & remove all pending guardians after recovery

dong77:
Remove `cancelGuardianAddition` and `cancelGuardianRemoval`.

Please carefully review `addGuardian` and `removeGuardian` functions, I am not so confident.


kongliangzhong (approved):

packages/hebao_v1/contracts/stores/GuardianStore.sol
```diff
@@ -0,0 +1,262 @@
...
+            _cleanExpiredGuardians(wallet);
+        } else {
+            Data.Guardian memory g = w.guardians[pos - 1];
+            bool isGuardianActive = _isActivated(g) && !_isExpired(g);
+
+            // If this wallet is active, do nothing.
+            if (!isGuardianActive) {
+                w.guardians[pos - 1].validSince = validSince.toUint40();
+                w.guardians[pos - 1].validUntil = 0;
+            }
```
Brechtpd, line 148: If the guardian is pending for removal `isGuardianActive` will be true, but `validUntil` is not reset to 0 in that case, so the pending removal is not cancelled, which I think is the desired outcome.

`w.guardians[pos - 1].validUntil = 0;` I think should always be done.

dong77: please see the new impl.

packages/hebao_v1/contracts/stores/GuardianStore.sol
```diff
@@ -0,0 +1,307 @@
...
+
+        if (_isAdded(g)) {
+            w.guardians[pos - 1].status = uint8(Data.GuardianStatus.REMOVE);
+            w.guardians[pos - 1].timestamp = validUntil.toUint64();
+            return validUntil;
+        }
+
+        if (_isPendingAddition(g)) {
+            w.guardians[pos - 1].status = uint8(Data.GuardianStatus.REMOVE);
+            w.guardians[pos - 1].timestamp = 0;
```
kongliangzhong, line 199: delete w.guardianIdx[addr] here.

packages/hebao_v1/contracts/stores/GuardianStore.sol
```diff
@@ -0,0 +1,307 @@
...
+        Data.Guardian memory g = w.guardians[pos - 1];
+
+        if (_isAdded(g)) {
+            w.guardians[pos - 1].status = uint8(Data.GuardianStatus.REMOVE);
+            w.guardians[pos - 1].timestamp = validUntil.toUint64();
+            return validUntil;
+        }
+
+        if (_isPendingAddition(g)) {
+            w.guardians[pos - 1].status = uint8(Data.GuardianStatus.REMOVE);
```
kongliangzhong, line 198: remove pending guardian immedately here.

Brechtpd: Isn't that what happens here? It's just not removed from the list which I think is fine.

Brechtpd (approved):

PR #1952, opened 2020-10-25
Shall we require at least one guardian in majority?

dong77:


Brechtpd:
I guess the question here is if a wallet has no guardians do we want to give the wallet owner full control of the wallet. I think we can because if the wallet owner keys are lost/compromised there is no way to recover the wallet any way so any protection mechanism we have is useless. Unless we think a scenario where both the actual owner and the attacker have access to the owner key is realistic. In that case it may still be useful to have some delays, but I'm not sure because both will work against each other so maybe still a lost cause, but probably still possible to partially get funds out in some scenario's.

dong77:
> I guess the question here is if a wallet has no guardians do we want to give the wallet owner full control of the wallet.

I would suggest disabling some features if there is no guardian at all.

PR #1955, opened 2020-10-25
[hebao] batch call

Brechtpd:
I'm back with batched call support. :)

Allow meta-txs or the wallet owner to do batched calls. I think also very important for efficiency (add multiple guardians, do multiple token transfers, whitelist multiple address,... or a combination of all of the above in multiple modules which isn't even possible with a module specific batch implementation)  with only the overhead of a single Ethereum tx/meta tx, which can be very significant (especially with reimbursement).

I think technically possible to support `txAwareHash` for batched calls, but I don't really see much use for that so kept things as simple as possible.

**Untested!**

### DIFF
--- a/packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
+++ b/packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
@@ -8,14 +8,14 @@ import "../../lib/AddressUtil.sol";
 import "../../lib/ERC1271.sol";
 import "../../lib/SignatureUtil.sol";
 import "../../thirdparty/BytesUtil.sol";
-import "../base/BaseModule.sol";
+import "../security/SecurityModule.sol";
 
 
 /// @title ERC1271Module
 /// @dev This module enables our smart wallets to message signers.
 /// @author Brecht Devos - <brecht@loopring.org>
 /// @author Daniel Wang - <daniel@loopring.org>
-abstract contract ERC1271Module is ERC1271, BaseModule
+abstract contract ERC1271Module is ERC1271, SecurityModule
 {
     using SignatureUtil for bytes;
     using SignatureUtil for bytes32;
--- a/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
+++ b/packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
@@ -71,6 +71,11 @@ abstract contract ForwarderModule is SecurityModule
         view
     {
         verifyTo(to, from, data);
+        require(
+            msg.sender != address(this) ||
+            data.toBytes4(0) == ForwarderModule.batchCall.selector,
+            "INVALID_TARGET"
+        );
 
         require(
             nonce == 0 && txAwareHash != 0 ||
@@ -227,6 +232,7 @@ abstract contract ForwarderModule is SecurityModule
         require(to.length == data.length, "INVALID_DATA");
 
         for (uint i = 0; i < to.length; i++) {
+            require(to[i] != address(this), "INVALID_TARGET");
             verifyTo(to[i], wallet, data[i]);
             // The trick is to append the really logical message sender and the
             // transaction-aware hash to the end of the call data.
@@ -256,7 +262,7 @@ abstract contract ForwarderModule is SecurityModule
     function verifyTo(
         address to,
         address wallet,
-        bytes memory data
+        bytes   memory data
         )
         internal
         view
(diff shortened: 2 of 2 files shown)


packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
```diff
@@ -8,14 +8,14 @@ import "../../lib/AddressUtil.sol";
...
-import "../base/BaseModule.sol";
+import "../security/SecurityModule.sol";
 
 
 /// @title ERC1271Module
 /// @dev This module enables our smart wallets to message signers.
 /// @author Brecht Devos - <brecht@loopring.org>
 /// @author Daniel Wang - <daniel@loopring.org>
-abstract contract ERC1271Module is ERC1271, BaseModule
+abstract contract ERC1271Module is ERC1271, SecurityModule
```
dong77, line 18: It's a bit wierd that this module is also a SecurityModule...

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -222,6 +215,28 @@ abstract contract ForwarderModule is BaseModule
         );
     }
 
+    function batchCall(
+        address   wallet,
+        address[] calldata to,
+        bytes[]   calldata data
+        )
+        external
+        txAwareHashNotAllowed()
```
dong77, line 224: I think we can skip checking the txAwareHash here, as it is not used in the function body at all. In each `to[i].call`, the calldata is managed using `abi.encodePacked(data[i], wallet, bytes32(0))` and it makes sure each internal transaction will have `bytes32(0)` as the txAwareHash.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -65,20 +70,8 @@ abstract contract ForwarderModule is BaseModule
...
-        // Stores via this module. Therefore, we must carefully check the 'to' address as follows,
-        // so no Store can be used as 'to'.
-        require(
-            (to != address(this)) &&
-            controllerCache.moduleRegistry.isModuleRegistered(to) ||
-
-            // We only allow the wallet to call itself to addModule
-            (to == from) &&
-            data.toBytes4(0) == Wallet.addModule.selector ||
+        verifyTo(to, from, data);
```
dong77, line 73: if `to` is `address(this)`, shall we check the function must be `batchCall`?
Inside `batchCall`, shall we check `to` can never be `address(this)`?

I assume the expected use case is to use `executeMetaTx` to invoke `batchCall`, not the other way around.

packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
Brechtpd: Was only done to fix compile problems. Otherwise `FinalSecurityModule` inherits from both `BaseModule` here and `SecurityModule` in `ForwarderModule` which needs reimplementation from certain virtual functions (multiple inheritance problems, solidity specific).

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
Brechtpd: Can do. I don't think there can be any problems not doing those checks, but can't hurt to limit how it can be used to make it easier to think about.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
```diff
@@ -222,6 +215,28 @@ abstract contract ForwarderModule is BaseModule
...
+        onlyFromWalletOrOwnerWhenUnlocked(wallet)
+    {
+        require(to.length == data.length, "INVALID_DATA");
+
+        for (uint i = 0; i < to.length; i++) {
+            verifyTo(to[i], wallet, data[i]);
+            // The trick is to append the really logical message sender and the
+            // transaction-aware hash to the end of the call data.
+            (bool success, ) = to[i].call(
+                abi.encodePacked(data[i], wallet, bytes32(0))
```
dong77, line 234: Why not use `Wallet.transact(...)`

Brechtpd: No real reason. But this will be quite a bit cheaper because this'll call modules, and if we have to pass through the wallet that's an additional 2 contract calls we have to pass through (proxy + wallet) to get to the same destination.

How could this simplify things? Now it just does the same thing we do in normal meta-txs.

dong77: I figured out, was going to say the same thing.

packages/hebao_v1/contracts/modules/core/ForwarderModule.sol
dong77: Lets be cautious by adding these extra checks.

PR #1956, opened 2020-10-26
[hebao] add `lockWA` that does not use nonce

dong77:
Locking a wallet should be a really urgent operation and shall not wait for any previous transactions to finish due to nonce dependency.

### DIFF
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -46,7 +46,7 @@ library GuardianUtils
             require(signers[i] > lastSigner, "INVALID_SIGNERS_ORDER");
             lastSigner = signers[i];
 
-            if (signers[i] == owner) {
+            if (signers[i] == owner || signers[i] == wallet) {
                 walletOwnerSigned = true;
             } else {
                 require(isWalletGuardian(allGuardians, signers[i]), "SIGNER_NOT_GUARDIAN");
--- a/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
+++ b/packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
@@ -12,9 +12,10 @@ library GuardianUtils
 {
     enum SigRequirement
     {
-        OwnerNotAllowed,
-        OwnerAllowed,
-        OwnerRequired
+        MAJORITY_OWNER_NOT_ALLOWED,
+        MAJORITY_OWNER_ALLOWED,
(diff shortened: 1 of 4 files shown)


packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -44,17 +46,24 @@ library GuardianUtils
             require(signers[i] > lastSigner, "INVALID_SIGNERS_ORDER");
             lastSigner = signers[i];
 
-            if (signers[i] == owner) {
+            if (signers[i] == owner || signers[i] == wallet) {
```
dong77, line 49: @Brechtpd I'm not sure this is correct or necessary.

Brechtpd:
> Locking a wallet should be a really urgent operation and shall not wait for any previous transactions to finish due to nonce dependency.

Can we allow overwriting meta txs like Ethereum txs in the relayer? If the user signs a new meta tx with the same nonce but higher gas price the previous one is replaced. These meta txs for the wallet will probably be even more centralized so seems even less of a problem I'd think. Existing meta txs will probably be useless after the lock as well.

dong77:
> > Locking a wallet should be a really urgent operation and shall not wait for any previous transactions to finish due to nonce dependency.
> 
> Can we allow overwriting meta txs like Ethereum txs in the relayer? If the user signs a new meta tx with the same nonce but higher gas price the previous one is replaced. These meta txs for the wallet will probably be even more centralized so seems even less of a problem I'd think. Existing meta txs will probably be useless after the lock as well.

Its certainly possible, the downside is if the transaction we try to replace with the lock operation get included first, the locking will fail. We need to make sure once a user signs the lock tx, it will succeed regardless.

adamchen0510 (approved):

packages/hebao_v1/contracts/modules/security/GuardianUtils.sol
```diff
@@ -44,17 +46,24 @@ library GuardianUtils
...
                 walletOwnerSigned = true;
```
Brechtpd, line 50: I think this needs a `require(!walletOwnerSigned, "WALLET_CANNOT_SIGN_TWICE")` here. Now the owner of the wallet can take up 2 slots in `signers.length`, and so the majority calculation will be wrong. So easiest I think to just limit it here so only one of the 2 is used.

dong77: This is really a nice catch!

Brechtpd (approved):

PR #1959, opened 2020-10-26
[loopring 3.6] remove duplicate events

dong77:


kongliangzhong (approved):

Brechtpd:
Does this work now with events emitted in libraries? It doesn't work in the tests...

PR #1960, opened 2020-10-27
Add checks when processing forced exit and slippageOK is false

kongliangzhong:


packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -65,7 +65,14 @@ library AmmExitProcess
         if (isForcedExit) {
             if (!slippageOK) {
                 AmmUtil.transferOut(address(this), exit.burnAmount, exit.owner);
+                
+                // check no pool tokens are transfered on L2:
+                for (uint i = 0; i < ctx.tokens.length; i++) {
+                    TransferTransaction.Transfer memory transfer = _block.readTransfer(ctx.txIdx++);
+                    require(transfer.amount == 0, "INVALID_AMM_EXIT_AMOUNT");
```
Brechtpd, line 72: Probably best to check some additional values, otherwise some other transfers that don't have anything to do with AMM can also be used instead:
```
require(
    transfer.fromAccountID == ctx.accountID &&
    transfer.from == address(this) &&
    transfer.to == exit.owner &&
    transfer.tokenID == ctx.tokens[i].tokenID &&
    transfer.amount == 0,
    "INVALID_EXIT_TRANSFER_TX_DATA"
);
```

Is the idea to always do the same number of transfers in all cases? If so, does it make sense to also add a `amount == 0` L2 transfer even for forced exits?

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -65,7 +65,14 @@ library AmmExitProcess
         if (isForcedExit) {
             if (!slippageOK) {
                 AmmUtil.transferOut(address(this), exit.burnAmount, exit.owner);
+                
+                // check no pool tokens are transfered on L2:
+                for (uint i = 0; i < ctx.tokens.length; i++) {
```
dong77, line 70: 是不是只检查第一个transfer是0就可以了。

dong77: 如果中继认为滑点不足应该失败，只要放一个0值的tranfer就可以了。

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: The idea is to prevent the following scenario from happening:
The relayer based on its offchain logic (which could be wrong due to a bug) believes the forced exit should be processed successfully (with good slippage), but the AMM contract believes the slippage requirement is not met. Now the AMM contract will transfer the pool token back to the user, but the relayer may also be transferred the two tokens to the user on layer2. 

The new change makes sure if the AMM contract believes the slippage is bad for a forced withdrawal, there are one or more zero-valued layer2 transfers to confirm this (consensus with the relayer). If not, it must be the relayer who makes a mistake and this entire transaction should fail. But we don't lose funds.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: @kongliangzhong This 0-value transfers also have to be approved onchain.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: I propose this:

- If the relayer believes the forced withdrawal will fail due to bad slippage, it will ALWAYS append one single 0 value transfer of ETH (ID = 0). The AMM contract will check and have to approve it. We also need to check some additional fields for sure.
- We make sure for all successful withdrawals, the transfer amount shall never be 0.

dong77: Wait, inside 
```solidity
     if (isForcedExit) {
            if (!slippageOK) {
```

As long as we didn't approve any transfer, then there should be no problem at all, so the transaction will fail! @Brecht?

Brechtpd: That's correct, the transfers aren't approved so the block isn't valid either way. But it will fail because some transfer wasn't authorized instead of giving a nicer error message here, which may be helpful for the relayer?

PR #1966, opened 2020-10-28
[hebao1.2] Allow unlimited quota

dong77:
Now a wallet's quota by default is 0, and it means the quota is unlimited, consuming quota will not update the quota value in QuotaStore to reduce gas. Values other than 0 will behave as usual.

@Brechtpd I'm not sure if `uint128(-1)` is the right way to calculate the max number for uint128.

### DIFF
--- a/packages/hebao_v1/contracts/stores/QuotaStore.sol
+++ b/packages/hebao_v1/contracts/stores/QuotaStore.sol
@@ -112,7 +112,7 @@ contract QuotaStore is DataStore
     }
 
     function spentQuota(address wallet)
-        external
+        public
         view
         returns (uint)
     {
@@ -141,7 +141,7 @@ contract QuotaStore is DataStore
     // Internal
 
     function _currentQuota(Quota memory q)
-        internal
+        private
         view
         returns (uint)
     {
@@ -149,7 +149,7 @@ contract QuotaStore is DataStore
     }
 
     function _pendingQuota(Quota memory q)
-        internal
+        private
         view
         returns (
             uint __pendingQuota,
@@ -163,7 +163,7 @@ contract QuotaStore is DataStore
     }
 
     function _spentQuota(Quota memory q)
-        public
+        private
         view
         returns (uint)
     {
@@ -176,7 +176,7 @@ contract QuotaStore is DataStore
     }
 
     function _availableQuota(Quota memory q)
-        public
+        private
         view
         returns (uint)
     {
@@ -192,7 +192,7 @@ contract QuotaStore is DataStore
         Quota   memory q,
         uint    requiredAmount
         )
-        public
+        private
         view
         returns (bool)
     {
@@ -204,7 +204,7 @@ contract QuotaStore is DataStore
         Quota   memory q,
         uint    amount
         )
-        internal
+        private
     {
         Quota storage s = quotas[wallet];
         s.spentAmount = _spentQuota(q).add(amount).toUint128();
--- a/packages/hebao_v1/contracts/stores/QuotaStore.sol
+++ b/packages/hebao_v1/contracts/stores/QuotaStore.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2017 Loopring Technology Limited.
 pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
 
 import "../base/DataStore.sol";
 import "../lib/MathUint.sol";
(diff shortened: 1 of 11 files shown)


packages/hebao_v1/contracts/stores/QuotaStore.sol
```diff
@@ -87,9 +71,11 @@ contract QuotaStore is DataStore, Claimable
         public
         onlyWalletModule(wallet)
     {
-        Quota memory q = quotas[wallet];
-        require(_hasEnoughQuota(q, amount), "QUOTA_EXCEEDED");
-        _addToSpent(wallet, q, amount);
+        uint available = availableQuota(wallet);
+        if (available != MAX_QUOTA) {
+            require(available >= amount, "QUOTA_EXCEEDED");
+            addToSpent(wallet, amount);
```
Brechtpd, line 77: This won't work, even when using `requireSelfOrWalletModule` on `addToSpent`. Changing this to `this.addToSpent(wallet, amount)` will work, but this will use ~1,700 gas more gas because it needs to do an external call.

packages/hebao_v1/contracts/stores/QuotaStore.sol
```diff
@@ -1,27 +1,25 @@
...
 import "../base/DataStore.sol";
 import "../lib/MathUint.sol";
-import "../lib/Claimable.sol";
 import "../thirdparty/SafeCast.sol";
 
 /// @title QuotaStore
 /// @dev This store maintains daily spending quota for each wallet.
 ///      A rolling daily limit is used.
-contract QuotaStore is DataStore, Claimable
+contract QuotaStore is DataStore
```
Brechtpd, line 12: No more SLOAD optimizations by caching `Quota` once in memory and reusing it? :(

Or do we see the quota now as something that will rarely get used so no need to optimize it as much as possible?

dong77: We should definitely optimize it, but according to Vitalik, if we write twice for the same variable, the second one only costs 200 gas, not sure if it is indeed the case. *I asked him about this in my morning*.

packages/hebao_v1/contracts/stores/QuotaStore.sol
Brechtpd: Will do!

packages/hebao_v1/contracts/stores/QuotaStore.sol
Brechtpd: Writing isn't so much the concern here, it's the reading. And unfortunately each read costs 800 gas regardless of how many times it has been read (Vitalik made a EIP to change this, but it currently doesn't have wide support so it's very uncertain if it will ever be included in a future hard fork).

Writing to the same storage slot is indeed quite cheap (well, at least cheaper than 20,000 of 5,000 gas), but the cost for this is still the same as the SLOAD cost. This was 200 gas previously, but this is now also 800 gas because of the SLOAD gas cost increase.
- https://github.com/sorpaas/EIPs/blob/c88b73889e66669db0e7094dda7907661f54d873/EIPS/eip-2200.md

packages/hebao_v1/contracts/stores/QuotaStore.sol
Brechtpd: This actually does work *because* `msg.sender` remains the same... But we still don't want to do that because it's expensive to do the module check twice.

packages/hebao_v1/contracts/stores/SecurityStore.sol
```diff
@@ -55,6 +39,7 @@ contract SecurityStore is GuardianStore
         uint    minInternval
         )
         external
+        onlyWalletModule(wallet)
```
Brechtpd, line 42: This one can actually be removed, `requireWalletModule(wallet);` is called only when `lastActive` is updated.

The reason I did this is to optimize for the most common case  `lastActive` doesn't need to updated, in which case there's no need to check `onlyWalletModule` (which is pretty expensive because it needs a CALL + SLOAD) because it acts just like a `view` function. And first calling a `view` function and then an actual `touch()` function is of course also more expensive because of the multiple calls.

So definitely a non-standard way to do things to save gas...

dong77: If we remove `requireWalletModule(wallet)`, any address will be touch the wallet to prevent inheritance from happening, won't it?

dong77: I noticed the requireWalletModule inside the funciton body.

Brechtpd (approved):

PR #1967, opened 2020-10-28
[hebao 1.2] improve DataStore sender check

dong77:


Brechtpd:
How will this be used? (I don't think anywhere the sender is `this`?). If you call a public function in a smart contract `msg.sender` is still the original value, not the address itself. To make the contract itself by the `msg.sender` you have to force it to do a real CALL doing something like `this.publicFunction()` but this will be more expensive than doing an internal call.

dong77:
> How will this be used? (I don't think anywhere the sender is `this`?). If you call a public function in a smart contract `msg.sender` is still the original value, not the address itself. To make the contract itself by the `msg.sender` you have to force it to do a real CALL doing something like `this.publicFunction()` but this will be more expensive than doing an internal call.

You are right ... somehow I ...

PR #1968, opened 2020-10-28
[loopring36] update migration scripts and add a getter method for VK

kongliangzhong:


packages/loopring_v3/contracts/core/impl/BlockVerifier.sol
```diff
@@ -134,4 +134,16 @@ contract BlockVerifier is ReentrancyGuard, IBlockVerifier
     {
         return circuits[blockType][blockSize][blockVersion].enabled;
     }
+
+    function getVerificationKey(
+        uint8  blockType,
+        uint16 blockSize,
+        uint8  blockVersion
+        )
+        external
```
dong77, line 143: change to public

dong77 (approved):

PR #1970, opened 2020-10-28
[hebao] QuotaStore optimizations

Brechtpd:
Reverted back to internal functions + caching `Quota`. A bit more functions and bloat, but caching `Quota` saves a nice 4,000 gas so seems worth it.

There's a ~3,000 gas overhead by having to call `currentQuota` to check if the quota is enabled, and so if the quota is enabled this is just extra cost. (in some cases `currentQuota` only needs to do 1 SLOAD in some cases when `pendingUntil` and either `currentQuota` or `pendingQuota` or stared in the same slot, but this would only work sometimes and I haven't found a clean way to do that yet).

But the if the quota is disabled there are of course savings, and this ends up being ~25,000 gas in the benchmark (~5,000 gas saved by being able to skip the whitelist check, ~5,000 gas saved by not having to get the token value from the price oracle, and the remaining ~15,000 gas by not having to call `checkAndAddToSpent`). The benchmark now also uses `PriceCacheStore` to read cached prices from an oracle. So when using e.g. the uniswap price oracle direclty without caching the savings would be a bit larger.

Benchmark now tests both with and without quota enabled.

### DIFF
--- a/packages/hebao_v1/contracts/stores/QuotaStore.sol
+++ b/packages/hebao_v1/contracts/stores/QuotaStore.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2017 Loopring Technology Limited.
 pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
 
 import "../base/DataStore.sol";
 import "../lib/MathUint.sol";
@@ -71,10 +72,11 @@ contract QuotaStore is DataStore
         public
         onlyWalletModule(wallet)
     {
-        uint available = availableQuota(wallet);
+        Quota memory q = quotas[wallet];
+        uint available = _availableQuota(q);
         if (available != MAX_QUOTA) {
             require(available >= amount, "QUOTA_EXCEEDED");
-            addToSpent(wallet, amount);
+            _addToSpent(wallet, q, amount);
         }
     }
 
@@ -85,9 +87,7 @@ contract QuotaStore is DataStore
         public
         onlyWalletModule(wallet)
     {
-        Quota storage q = quotas[wallet];
-        q.spentAmount = spentQuota(wallet).add(amount).toUint128();
-        q.spentTimestamp = uint64(block.timestamp);
+        _addToSpent(wallet, quotas[wallet], amount);
     }
 
     // Returns 0 to indiciate unlimited quota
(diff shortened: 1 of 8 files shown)


packages/hebao_v1/contracts/stores/QuotaStore.sol
```diff
@@ -85,9 +87,7 @@ contract QuotaStore is DataStore
         public
         onlyWalletModule(wallet)
     {
-        Quota storage q = quotas[wallet];
-        q.spentAmount = spentQuota(wallet).add(amount).toUint128();
-        q.spentTimestamp = uint64(block.timestamp);
+        _addToSpent(wallet, quotas[wallet], amount);
```
dong77, line 90: shall we change `function addToSpent`'s visibility to `external`?

packages/hebao_v1/contracts/test/PriceCacheStore.sol
```diff
@@ -4,16 +4,18 @@ pragma solidity ^0.7.0;
...
-
 
 /// @title PriceCacheStore
-contract PriceCacheStore is DataStore, PriceOracle, OwnerManagable
+contract PriceCacheStore is PriceOracle, OwnerManagable
 {
     using MathUint for uint;
+    using SafeCast for uint;
+
+    uint public constant EXPIRY_PERIOD = 14 days;
```
dong77, line 18: suggest to change it to 1 week instead.

packages/hebao_v1/contracts/test/PriceCacheStore.sol
```diff
@@ -77,14 +74,14 @@ contract PriceCacheStore is DataStore, PriceOracle, OwnerManagable
...
-        onlyManager
+        onlyOwnerOrManager
     {
         cacheTokenPrice(token, amount, value);
     }
 
     function setOracle(PriceOracle _oracle)
         external
-        onlyManager
+        onlyOwnerOrManager
```
dong77, line 84: I still prefer manager-only to force the owner to add and use a manager so that the owner address will not be used in any hot wallet at all.

Brechtpd: Okay, will revert.

packages/hebao_v1/contracts/stores/QuotaStore.sol
Brechtpd: I think making at least all functions that are `onlyWalletModule` `external` makes a lot of sense. I think in some select cases `public` may be marginally more efficient than `external` but it's so small it doesn't really matter.

dong77 (approved):

PR #1971, opened 2020-10-29
[hebao 1.2] adjust function modifier

dong77:


Brechtpd (approved):

Brechtpd:
So for non `internal`/`private` functions the rule is always `external`, unless for `view` interface functions when needed?

PR #1979, opened 2020-10-29
[hebao 1.2] Start using immutable

Brechtpd:
So I found `immutable` exists since solidity 0.7 while reading the solidity github:
- https://solidity.readthedocs.io/en/v0.7.4/contracts.html?highlight=immutable#constant-and-immutable-state-variables

Allows immutable simple data to be initialized in the constructor without using storage, so as cheap to read as constants most times. So we could pass in constant values in the constructor with this without loss of efficiency (after we stopped doing that until recently...).

Maybe some other cases it could be used as well. 

Does it make sense to just simply cache the store addresses in `immutable` vars in modules? Doesn't seem likely we can just update the stores used in the modules without any upgrading. @dong77 

Same for `WalletFactory`, easy to replace the contract when needed so not that important that those ENS contracts would be hard-coded I think.

### DIFF
--- a/packages/hebao_v1/contracts/base/Controller.sol
+++ b/packages/hebao_v1/contracts/base/Controller.sol
@@ -10,6 +10,15 @@ import "../iface/ModuleRegistry.sol";
 /// @author Daniel Wang - <daniel@loopring.org>
 abstract contract Controller
 {
-    ModuleRegistry public moduleRegistry;
-    address        public walletFactory;
+    function moduleRegistry()
+        external
+        view
+        virtual
+        returns (ModuleRegistry);
+
+    function walletFactory()
+        external
+        view
+        virtual
+        returns (address);
 }
--- a/packages/hebao_v1/contracts/modules/ControllerImpl.sol
+++ b/packages/hebao_v1/contracts/modules/ControllerImpl.sol
@@ -18,13 +18,15 @@ import "../thirdparty/ens/BaseENSManager.sol";
 /// @author Daniel Wang - <daniel@loopring.org>
 contract ControllerImpl is Claimable, Controller
 {
+    HashStore           public immutable hashStore;
+    QuotaStore          public immutable quotaStore;
+    SecurityStore       public immutable securityStore;
+    WhitelistStore      public immutable whitelistStore;
+    ModuleRegistry      public immutable override moduleRegistry;
+    address             public override  walletFactory;
     address             public collectTo;
     BaseENSManager      public immutable ensManager;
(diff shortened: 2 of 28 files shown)


Brechtpd:
Oh seems like stores are already not updatable in `ControllerImpl`, so we could just cache them in the modules without any change in behavior! Will make a separate PR for that.

kongliangzhong (approved):

dong77 (approved):

dong77:
> So I found `immutable` exists since solidity 0.7 while reading the solidity github:
> 
> * https://solidity.readthedocs.io/en/v0.7.4/contracts.html?highlight=immutable#constant-and-immutable-state-variables
> 
> Allows immutable simple data to be initialized in the constructor without using storage, so as cheap to read as constants most times. So we could pass in constant values in the constructor with this without loss of efficiency (after we stopped doing that until recently...).
> 
> Maybe some other cases it could be used as well.
> 
> Does it make sense to just simply cache the store addresses in `immutable` vars in modules? Doesn't seem likely we can just update the stores used in the modules without any upgrading. @dong77

I think we SHOULD cache stores inside modules, there is no case that we only upgrade stores without upgrading modules. Maybe we should cache all contract addresses to avoid calling for another address to interact with.

> 
> Same for `WalletFactory`, easy to replace the contract when needed so not that important that those ENS contracts would be hard-coded I think.
I don't understand this question, could you please clarify?

Brechtpd:
> I think we SHOULD cache stores inside modules, there is no case that we only upgrade stores without upgrading modules. Maybe we should cache all contract addresses to avoid calling for another address to interact with.

Only upgradeable ones are `collectTo` and `priceOracle`, which seems reasonable. Small risk with making `priceOracle` updatable, but I think being able to update that one at low cost makes sense because it's likely that may have to tweaked more often.

> > Same for `WalletFactory`, easy to replace the contract when needed so not that important that those ENS contracts would be hard-coded I think.
> I don't understand this question, could you please clarify?

If we can easily just switch to a new `WalletFactory` we may as well just hard code all addresses to save some guess instead of making it possible to upgrade them in the existing contract.

But I thought `WalletFactory` could also be easily updated to a new version, but it seems that's not completely true because `walletFactory` is normally not updatable in the controller. On the other hand, the address is only used for creating the wallet and never again so doesn't really matter much I guess.

Currently `ensManager` is immutable because it cannot be changed in the controller, but `ensResolver` and `ensReverseRegistrar` are not immutable because they can currently be updated. I think this is fine.

PR #1984, opened 2020-10-30
[hebao 1.2] More optimizations

Brechtpd:
- Kind of changed my mind on making `priceOracle` and `collectTo` immutable. I think it makes sense to do so. The price oracle in `PriceCacheStore` can still be updated, so the price oracle address that is static in the modules is actually just some price caching contract, which is unlikely to need any updating.
- Added `isWhitelisted` to function calls that do whitelist checking. The caller knows if the address is whitelisted or not, so it's not necessary to check if it's actually whitelisted or not if the address isn't whitelisted. So if `isWhitelisted == false` we can skip the check.
- Transfers can now only be done to whitelisted addresses of the user, not to whitelisted dApps. This was potentially dangerous because an attacker could still transfer funds to a dApp smart contract so funds could still be lost completely (though the hacker couldn't transfer to his own address).
- Changed `checkAndAddToSpent` to also query the value on the price oracle to save on calls. This way we don't have to do an initial `_needCheckQuota` call. The downside here is that more work is done in the store itself.

With quota enabled these optimizations save 8,000 - 10,000 gas, without quota these optimizations save 0 - 4,000 gas depending on the context.


packages/hebao_v1/contracts/modules/base/BaseModule.sol
```diff
@@ -25,20 +25,14 @@ abstract contract BaseModule is Module
...
-    ControllerCache public controllerCache;
-
     ModuleRegistry public immutable moduleRegistry;
     SecurityStore  public immutable securityStore;
     WhitelistStore public immutable whitelistStore;
     QuotaStore     public immutable quotaStore;
     HashStore      public immutable hashStore;
     address        public immutable walletFactory;
+    PriceOracle    public immutable priceOracle;
+    address        public immutable collectTo;
```
dong77, line 35: In ControllerImpl, there is a `setCollectTo` function, if collectTo is immutable in modules, then the modules won't be able to use any new value. We have to consider the scenario that we sometimes have to change the fee recipient address.

packages/hebao_v1/contracts/modules/base/BaseModule.sol
```diff
@@ -25,20 +25,14 @@ abstract contract BaseModule is Module
...
-
-    ControllerCache public controllerCache;
-
     ModuleRegistry public immutable moduleRegistry;
     SecurityStore  public immutable securityStore;
     WhitelistStore public immutable whitelistStore;
     QuotaStore     public immutable quotaStore;
     HashStore      public immutable hashStore;
     address        public immutable walletFactory;
+    PriceOracle    public immutable priceOracle;
```
dong77, line 34: If the priceOracle is immutable, we have to make sure it's fully configurable after deployment, even the EXPIRY_PERIOD should be updatable.

To save more gas, I'd suggest we merge PriceCacheStore and Uniswap2PriceOracle into one single priceOracle contract, and still keep priceOracle cached in each module.

packages/hebao_v1/contracts/modules/transfers/BaseTransferModule.sol
```diff
@@ -78,13 +78,21 @@ abstract contract BaseTransferModule is SecurityModule
...
-    function isTargetWhitelisted(address wallet, address to)
+    function isAddressWhitelisted(address wallet, address to)
+        internal
+        view
+        returns (bool res)
+    {
+        (res,) = whitelistStore.isWhitelisted(wallet, to);
+    }
+
+    function isContractWhitelisted(address wallet, address to)
```
dong77, line 95: Suggest to rename it to `isAddressDappOrWhitelisted`

packages/hebao_v1/contracts/modules/transfers/TransferModule.sol
```diff
@@ -81,14 +81,15 @@ abstract contract TransferModule is BaseTransferModule
         address        token,
         address        to,
         uint           amount,
-        bytes calldata logdata
+        bytes calldata logdata,
+        bool           isWhitelisted
```
dong77, line 85: with this parameter, the caller (the one who steals a wallet) can get around the quota check, so the security will be compromised...

dong77 (changes requested):

packages/hebao_v1/contracts/modules/base/BaseModule.sol
Brechtpd: But can't we just handle that in the `collectTo` contract? This can just be a simple contract that is able to set a dynamic recipient address. It's cheaper to make it configurable there than in the modules because in the modules we have to read it in for each transaction.

packages/hebao_v1/contracts/modules/base/BaseModule.sol
Brechtpd: > If the priceOracle is immutable, we have to make sure it's fully configurable after deployment, even the EXPIRY_PERIOD should be updatable.

Do we care a lot about `EXPIRY_PERIOD`? It doesn't seem that important, anyone is able to update the prices at any time with the oracle, or we can update the prices manually. It doesn't seem to be an important value. To me it's more like when in the worst case maybe a too high a price is set for a long time or something, but even then that's not too bad and we can fix that by overriding the price if we need to.

Brechtpd: > To save more gas, I'd suggest we merge PriceCacheStore and Uniswap2PriceOracle into one single priceOracle contract, and still keep priceOracle cached in each module.

`tokenValue()` doesn't read in the active price oracle so no gas is saved on the frequent path by merging them (only read in when `updateTokenPrice` is called, which won't be that often, I assume at most called once a day for some tokens) , so I don't think we should merge them so we are still able to update the actual price oracle when needed.

packages/hebao_v1/contracts/modules/transfers/TransferModule.sol
Brechtpd: The logic doesn't just trust the value passed into the function. 

- `isWhitelisted == false` -> skip onchain whitelist check and **always** update the quota (unless disabled).
- `isWhitelisted == true` -> do onchain whitelist check and update the quota when not whitelisted.

So I think it's fine if I implemented the logic correctly. `if (!isWhitelisted || !isAddressDappOrWhitelisted(wallet, to)) { ... }`.

Maybe a better name would be `checkWhitelisted`?

packages/hebao_v1/contracts/modules/base/BaseModule.sol
Brechtpd: > To save more gas, I'd suggest we merge PriceCacheStore and Uniswap2PriceOracle into one single priceOracle contract, and still keep priceOracle cached in each module.

Merging `QuotaStore` and `PriceCacheStore` would save a bit of gas however.

packages/hebao_v1/contracts/modules/base/BaseModule.sol
dong77: I think these two solutions will end up using the same gas, the real collectTo address inside the collectTo contract will also have to be read before transferring tokens to it.

packages/hebao_v1/contracts/modules/base/BaseModule.sol
dong77: Fine with me to keep the changes you made to priceOracle (immutable).

packages/hebao_v1/contracts/modules/transfers/TransferModule.sol
dong77: I see the point. Name it as `forceUsingQuota` or something like that is better.

dong77: I like the logic, quite smart :)

packages/hebao_v1/contracts/modules/base/BaseModule.sol
Brechtpd: I don't think the real `collectTo` needs to be read. The `collectTo` would just be a smart contract collecting all funds, so there's no need for the modules to know anything else. This smart contract at a fixed address collects all fees as they come in. Then if we actually want to send those fees to the real destination address it will be possible to specify a destination address (can even be passed in as a function argument).

The immutable `collectTo` address that's used in the modules can just be a simple `Drainable` contract. It's even possible to put the contract behind a proxy to make the contract itself updatable if that would be wanted.

dong77: No need to make it upgradable or use a contract then. Could you please add a simple FeeCollector contract which is drainable by owner?

dong77 (approved):

dong77:
@Brechtpd I'll add the FeeCollector contract.

PR #1988, opened 2020-11-02
[hebao 1.2] allow wallet to sign `perpetual signatures`

dong77:
Allow wallet owner to provide signatures that do not expire within 30 days. This is to accommodate DEX transaction ECDSA approval in the case of wallet changes owner.

### DIFF
--- a/packages/hebao_v1/contracts/iface/Wallet.sol
+++ b/packages/hebao_v1/contracts/iface/Wallet.sol
@@ -18,7 +18,7 @@ interface Wallet
 
     function owner() external view returns (address);
 
-    function previousOwner(uint idx) external view returns (address);
+    function previousOwner(uint idx) external view returns (address, uint);
 
     /// @dev Set a new owner.
     function setOwner(address newOwner) external;
--- a/packages/hebao_v1/contracts/iface/Wallet.sol
+++ b/packages/hebao_v1/contracts/iface/Wallet.sol
@@ -16,9 +16,8 @@ interface Wallet
 {
     function version() external pure returns (string memory);
 
+    // @return The current owner address.
     function owner() external view returns (address);
-
-    function previousOwner(uint idx) external view returns (address, uint);
 
     /// @dev Set a new owner.
     function setOwner(address newOwner) external;
--- a/packages/hebao_v1/contracts/iface/Wallet.sol
+++ b/packages/hebao_v1/contracts/iface/Wallet.sol
@@ -18,6 +18,8 @@ interface Wallet
 
     function owner() external view returns (address);
 
+    function previousOwner(uint idx) external view returns (address);
+
     /// @dev Set a new owner.
     function setOwner(address newOwner) external;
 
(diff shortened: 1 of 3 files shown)


packages/hebao_v1/contracts/iface/Wallet.sol
```diff
@@ -18,6 +18,8 @@ interface Wallet
 
     function owner() external view returns (address);
 
+    function previousOwner(uint idx) external view returns (address, uint);
```
dong77, line 21: I'm not happy with this API, but I don't know how to come up with a more appropriate one.

packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
```diff
@@ -46,13 +52,27 @@ abstract contract ERC1271Module is ERC1271, SecurityModule
...
+            if (securityStore.isLocked(wallet)) {
+                return 0;
+            }
+
+            address _owner = Wallet(wallet).owner();
+            bytes memory _sig = _signature.slice(8, _signature.length - 8);
+            return _signHash.verifySignature(_owner, _sig) ? ERC1271_MAGICVALUE : bytes4(0);
+        } else if (sigType == uint8(SignType.PERPETUAL)) {
+            uint ownerIdx = _signature.toUint32(8);
+            uint ownerTimestamp = _signature.toUint64(40);
```
dong77, line 68: I put ownerTimestamp in the signature so that the client can extract the value and make sure this value is not too old, compared to the current time.

packages/hebao_v1/contracts/modules/core/ERC1271Module.sol
```diff
@@ -46,13 +52,27 @@ abstract contract ERC1271Module is ERC1271, SecurityModule
...
+            address wallet = msg.sender;
+            if (securityStore.isLocked(wallet)) {
+                return 0;
+            }
+
+            address _owner = Wallet(wallet).owner();
+            bytes memory _sig = _signature.slice(8, _signature.length - 8);
+            return _signHash.verifySignature(_owner, _sig) ? ERC1271_MAGICVALUE : bytes4(0);
+        } else if (sigType == uint8(SignType.PERPETUAL)) {
+            uint ownerIdx = _signature.toUint32(8);
```
dong77, line 67: I put ownerIdx here so we don't have to do a binary search in contract for the signing owner address.

Brechtpd:
Doesn't this break locking/recover/inherit? If owners can just sign things that are always valid (regardless of locking/recovery/inherit) it seems many of these features are useless. But if the user can always lock/inherit/recover at any time than it seems these perpetual signatures can never work.

Perhaps a way to solve this is to allow to allow hashes to be approved with majority in `ERC1271Module`, but this would also be an L1 operation so not much difference compared to just calling `approveTransaction` on the exchange I guess.

dong77:
I have to think about it more, if the owner's private key is leaked, then perpetual signatures will be a big problem...

PR #1990, opened 2020-11-02
[protocol 3.6] Allow initializing owner with account update txs

Brechtpd:
https://github.com/Loopring/team/issues/147

Less changes needed than I thought, so seems like a good change regardless if it will be used a lot or not.

### DIFF
--- a/packages/loopring_v3.js/src/request_processors/account_update_processor.ts
+++ b/packages/loopring_v3.js/src/request_processors/account_update_processor.ts
@@ -20,10 +20,15 @@ interface AccountUpdate {
  * Processes account update requests.
  */
 export class AccountUpdateProcessor {
-  public static process(state: ExchangeState, block: BlockContext, txData: Bitstream) {
+  public static process(
+    state: ExchangeState,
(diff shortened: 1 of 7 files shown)


dong77:
Brecht, just want to make sure if the current owner of an account is non-zero, this tx will fail. Also if the new owner is 0x0, it will also fail.

yueawang (approved):

autumn84 (approved):

Brechtpd:
> just want to make sure if the current owner of an account is non-zero, this tx will fail.

Checked in the circuit by `OwnerValidGadget` (already used by deposit).

> Also if the new owner is 0x0, it will also fail.

Not checked in the circuit, but if the new owner is `0x0` it's impossible to authorize the transaction in any way: https://github.com/Loopring/protocols/blob/2e012bd8b26e19cde2e6396158fa85aa4f08ad4e/packages/loopring_v3/contracts/core/impl/libexchange/ExchangeSignatures.sol#L27

dong77 (approved):

PR #2018, opened 2020-11-10
[Hebao1.2] change wallet proxy to OwnedUpgradeabilityProxy

kongliangzhong:


dong77 (approved):

packages/hebao_v1/contracts/modules/core/WalletFactory.sol
```diff
@@ -288,6 +289,7 @@ contract WalletFactory
         )
         private
     {
+        OwnedUpgradeabilityProxy(_wallet.toPayable()).transferProxyOwnership(_owner);
```
Brechtpd, line 292: I don't think we set the wallet owner as the proxy owner. In the previous wallet version that had an upgradeble proxy we set the owner of the proxy to the wallet itself, which I think makes more sense. This way we can allow modules to upgrade the wallet easily going through the modules and don't have to worry about updating the address when the owner changes. And also for security reasons because we can't just allow the wallet owner to update the wallet implementation.

kongliangzhong: Yes, I agree. For security reasons we should not allow anyone to change the implementation.

PR #2025, opened 2020-11-12
add hebao_v2, using hardhat framework

kongliangzhong:


Brechtpd:
Current design is very close to a basic implementation of the Diamond Standard (EIP-2535). Quite a lot of data/discussion available for this design pattern that can be helpful for potential security issues/new ideas:
- https://eips.ethereum.org/EIPS/eip-2535
- https://github.com/mudgen/Diamond
- https://github.com/ethereum/EIPs/issues/2535
- https://blog.trailofbits.com/2020/10/30/good-idea-bad-design-how-the-diamond-standard-falls-short/
- https://dev.to/mudgen/addressing-josselin-feist-s-concern-s-of-eip-2535-diamond-standard-me8

Seems like trailofbits is still more a fan (at least with a complex implementation for the Diamond pattern) of the standard upgradable patterns, like proxies and/or splitting up the contracts into implementation and stores like our current wallet does things. I think the current basic should probably be fine because it's still pretty simple, but it does introduce some potential riskier things which perhaps have less well understood security implications like explained in their audit report linked above. Especially for a wallet important to make a good balance between the different factors.

With `Wallet` now being basically a proxy contract itself I guess there's really no need to but it behind another proxy in any case because it could even be made to work like a normal proxy contract (with some additional overhead).

Another way quite a lot of gas could be saved in some cases is doing things stateless (like was available in protocol 3.6). Instead of doing multiple expensive SLOADs/SSTOREs each time frequently used data like the nonce, last active, quota, ... are used, we just store all that data in a single hash onchain and send the actual data in the calldata. Probably not a good idea though to start doing that now because that'd be a big change how the wallet works and probably makes some things harder to do.

dong77:
> Another way quite a lot of gas could be saved in some cases is doing things stateless (like was available in protocol 3.6). Instead of doing multiple expensive SLOADs/SSTOREs each time frequently used data like the nonce, last active, quota, ... are used, we just store all that data in a single hash onchain and send the actual data in the calldata. Probably not a good idea though to start doing that now because that'd be a big change how the wallet works and probably makes some things harder to do.

Hebao is designed to be a mainnet smart wallet, which means it should secure people's assets without relying on any L2 technology. There will be a market for L2 smart wallet, but it's not our target.

Using a hash to represent a wallet's state is quite an interesting idea. If we put all wallet state DA on chain, when we need to change the some state variable, we need to provide the previous state and the new state of the wallet as calldata, then allow our smart contract to verify the previous hash and save the new hash. I think this can be done. I'll probably create a PR for this idea.

dong77 (dismissed):

PR #2028, opened 2020-11-13
Hebao v2 alternative design

dong77:
@Brechtpd @kongliangzhong 
I wrote these files to demonstrate the (brecht's) idea of stateless wallet. Please take a look.


Brechtpd:
One of the hard things currently is that gas cost for expensive operations will very likely change quite a bit pretty soon. So for a very different design to optimize specifically for gas, it's hard to make a good prediction what is the best approach.

The Berlin HF is currently scheduled for January, so pretty soon. But it's not certain yet what EIPs will end up being included. The current list can be found at https://github.com/ethereum/eth1.0-specs/blob/master/client-integration-testnets/YOLOv3.md. Progress for the different EIPs can be found at https://github.com/ethereum/eth1.0-specs/projects/1#column-10680418.

Most importantly for us is [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929), which changes SLOAD and SSTORE gas costs. Costs will be higher for the initial read/store, but will be much lower for subsequent reads. It didn't seem like this EIP would be included because there was some opposition to it, but now it looks likely it will. 

Another interesting one that is close to being included on the progress tracker, but as far as I know currently not yet planned to be included in Berlin is  [EIP-2046](https://eips.ethereum.org/EIPS/eip-2046). This makes calls to precompiles much cheaper. This is actually important for us because that would mean solidity may again use the identity precompile to do memory->memory copies. These are now expensive, but with the precompile they would almost as cheap as calldata -> memory copies. 

So EIP-2929 important to know what's most optimal: send a lot of data in the calldata (at 16 gas/non-zero byte) or just read the data from storage where needed (800 gas vs 2100 gas).  EIP-2046 important to know what the gas cost may be to copy a lot of wallet data around which is also directly related to the stateless wallet implementation.

So I could say that this or that may be more efficient in your PR, but unfortunately we can't really be sure until what EIPs go into Berlin are finalized to really know what is optimzal. And even then for things like EIP-2046 it also depends on what EVM code the solidity starts generating for memory copies (this used to be the identity precompile before the gas costs of all CALLs were drastically increased and precompile calls were not excluded).

dong77:
These are very valuable inputs. I think we can use the two designs to get a few important features ready, then do some gas comparisons after the fork. 2.0 is not urgent.

kongliangzhong:
Closed after discussed with dong77.

PR #2047, opened 2020-11-22
[protocol 3.6] Circuit fixes

Brechtpd:
Fixed some issues/bugs found while running trusted setup scripts which does some additional checks.

These changes fix the "unconstrained variables" warnings (which isn't always a bug, sometimes just variables that aren't actually used).

Trusted setup scripts still don't work unfortunately because the proving key format was changed after the prover optimizations and my updated code to import the bellman proving key doesn't work yet. Working on that now.

### DIFF
--- a/packages/loopring_v3/circuit/Gadgets/MathGadgets.h
+++ b/packages/loopring_v3/circuit/Gadgets/MathGadgets.h
@@ -1797,6 +1797,7 @@ class SelectGadget : public GadgetT
 
     SelectGadget(
       ProtoboardT &pb,
+      const Constants &_constants,
       const VariableArrayT &selector,
       const std::vector<VariableT> &values,
       const std::string &prefix)
@@ -1806,7 +1807,7 @@ class SelectGadget : public GadgetT
         for (unsigned int i = 0; i < values.size(); i++)
         {
             results.emplace_back(TernaryGadget(
-              pb, selector[i], values[i], (i == 0) ? values[0] : results.back().result(), FMT(prefix, ".results")));
+              pb, selector[i], values[i], (i == 0) ? _constants._0 : results.back().result(), FMT(prefix, ".results")));
         }
     }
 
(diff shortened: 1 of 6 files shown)


kongliangzhong (dismissed):

Brechtpd:
Tested with small small circuits and the trusted setup scripts now work. I will also do some more testing with bigger circuits today/tomorrow.

The problem is related to the circuit itself. The 3.1 circuits still worked fine. I'm still not sure what the issue is exactly, but it seems to be related to some specific constraint constructions which for some reason gives issues when going libsnark -> bellman -> libsnark. The 3.1 had none of these specific cases, and the 3.6 circuits had 2 places that gave issues that were easily worked around.

ethsnarks was updated to convert directly to the latest proving key format, so no additional conversions of the proving key are necessary.

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
```diff
@@ -1806,7 +1807,7 @@ class SelectGadget : public GadgetT
         for (unsigned int i = 0; i < values.size(); i++)
         {
             results.emplace_back(TernaryGadget(
-              pb, selector[i], values[i], (i == 0) ? values[0] : results.back().result(), FMT(prefix, ".results")));
+              pb, selector[i], values[i], (i == 0) ? _constants._0 : results.back().result(), FMT(prefix, ".results")));
```
yueawang, line 1810: Just out of curiosity, does this save constraints?

yueawang (approved):

packages/loopring_v3/circuit/Gadgets/MathGadgets.h
Brechtpd: No, it's a workaround to make our trusted setup work. The specific value is unimportant here, and using `values[0]` makes the trusted setup proving keys not work for reasons that are not completely clear to me. I think there's an issue for some reason when using the same variable multiple times in a linear combination in a constraint, but totally not sure about that.

PR #2055, opened 2020-11-26
[Protocol 3] Allow better processing of 0 deposits

dong77:
There is small issue with the current code, if a user did two 0-deposits, after we process the first one,the second one is gone onchain so the processing of the second one will fail.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/libtransactions/DepositTransaction.sol
+++ b/packages/loopring_v3/contracts/core/impl/libtransactions/DepositTransaction.sol
@@ -43,6 +43,9 @@ library DepositTransaction
     {
         // Read in the deposit
         Deposit memory deposit = readTx(data, offset);
+        if (deposit.amount == 0) {
+            return;
+        }
 
         // Process the deposit
         ExchangeData.Deposit memory pendingDeposit = S.pendingDeposits[deposit.to][deposit.tokenID];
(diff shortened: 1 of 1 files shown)


kongliangzhong:
> There is small issue with the current code, if a user did two 0-deposits, after we process the first one,the second one is gone onchain so the processing of the second one will fail.

If there are two deposits with the same tokenId of a user, Then we should proceed it as a single deposit. So I think this is not a bug.

dong77:
The point is to allow the relayer to process these two 0-deposits seperately.

Brechtpd:
Code looks fine.

But why would the relayer do these multiple 0-deposits? Or even a single 0-deposit? Seems to me these could as well just be ignored by the relayer. I guess it could initialize an account for a user by setting the owner, but in the end that's not very useful without any funds in it because that may as well happen when the user actually transfer funds to his account (non-zero deposit or transfer to a new account). Alternatively AccountUpdate can now also  be used to fully initialize owner + keys with a signature, which will have a lower total gas consumption.

dong77:
> Code looks fine.
> 
> But why would the relayer do these multiple 0-deposits? Or even a single 0-deposit? Seems to me these could as well just be ignored by the relayer. I guess it could initialize an account for a user by setting the owner, but in the end that's not very useful without any funds in it because that may as well happen when the user actually transfer funds to his account (non-zero deposit or transfer to a new account). Alternatively AccountUpdate can now also be used to fully initialize owner + keys with a signature, which will have a lower total gas consumption.

I agree. I'll try not to merge this PR. Let's see if handling 0-deposits are reasonable in certain scenarios.

Brechtpd (approved):

PR #2059, opened 2020-11-27
[protocol 3.6] AMM range check fix

### LINKED ISSUE #2058: A decimal issue on AMM ratio check
Hi, @Brechtpd 

In AMM circuit, there is a ratio check.
```
ratio(
            pb,
            constants,
            numer.result(),
            constants.fixedBase,
            denom.result(),
            NUM_BITS_AMOUNT * 2,
            60,
            NUM_BITS_AMOUNT * 2,
            FMT(prefix, ".ratio")),
ratioRangeCheck(pb, ratio.result(), NUM_BITS_AMOUNT - 14 /*log2(10000)*/, FMT(prefix, ".ratioRangeCheck")),
```

The ratio is the price with fixed_base precision, i.e., ratio = BalanceIn * WeightOut * (10 \*\* 18) / BalanceOut * WeightIn
And the MAX ratio is 2\*\*(96-14) ==4835703278458516698824704.

Normally it's ok, especially when 2 tokens are both in 10 \*\* 18 decimal. But when these 2 tokens have 2 different decimals, say 18 (ETH) vs 6 (USDT). The max ratio limitation is easily broken.
For example:
```
EthAmount = 1 * 1e18
UsdtAmount = 500 * 1e6
ratio =  1e18 * 1e18/(500*1e6)  > 4835703278458516698824704
```

This limitation makes ETH-USDT AMM pool fail validation.

We have some discussion @kongliangzhong @hosschao @lydy @letsgoustc , it looks like a larger ratio can be a workaround for this. To accommodate the biggest decimal gap 18, could we have a max ratio to at least 2 \*\* (96 + 60 - 14) == 2 \*\* 142 ? I think it's OK as an intermedia variable, and the input/output could be still 96bits.
Another hotfix way maybe to decrease the FIXED_BASE to maybe 10**8?? But this workaround maybe not so complete.

Could you please help to check this??
Thanks!

Brechtpd:
#2058 

Increases max range of ratio to 2**(96*2-14).

### DIFF
--- a/packages/loopring_v3/circuit/Gadgets/MatchingGadgets.h
+++ b/packages/loopring_v3/circuit/Gadgets/MatchingGadgets.h
@@ -493,8 +493,8 @@ class OrderMatchingGadget : public GadgetT
 //     return Math.floor((ratio * BASE_BIPS) / invFeeBips);
 // }
 
-// Result is guaranteed to fit inside NUM_BITS_AMOUNT bits.
-// Max ratio between weights guaranteed to be supported is 2**(96-14)/(10**18) = 4835703.278458517
+// Result is guaranteed to fit inside NUM_BITS_AMOUNT*2 bits.
+// Max ratio between weights guaranteed to be supported is 2**(96*2-14)/(10**18)
 // but this depends on the balances inside the pool as well. Normally it should be even much higher.
 class SpotPriceAMMGadget : public GadgetT
 {
@@ -529,7 +529,7 @@ class SpotPriceAMMGadget : public GadgetT
             60,
             NUM_BITS_AMOUNT * 2,
             FMT(prefix, ".ratio")),
-          ratioRangeCheck(pb, ratio.result(), NUM_BITS_AMOUNT - 14 /*log2(10000)*/, FMT(prefix, ".ratioRangeCheck")),
+          ratioRangeCheck(pb, ratio.result(), NUM_BITS_AMOUNT*2 - 14 /*log2(10000)*/, FMT(prefix, ".ratioRangeCheck")),
           invFeeBips(pb, constants._10000, feeBips, FMT(prefix, ".invFeeBips")),
           res(
             pb,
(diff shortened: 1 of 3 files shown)


packages/loopring_v3/circuit/Gadgets/MatchingGadgets.h
```diff
@@ -493,8 +493,8 @@ class OrderMatchingGadget : public GadgetT
 //     return Math.floor((ratio * BASE_BIPS) / invFeeBips);
 // }
 
-// Result is guaranteed to fit inside NUM_BITS_AMOUNT bits.
-// Max ratio between weights guaranteed to be supported is 2**(96-14)/(10**18) = 4835703.278458517
+// Result is guaranteed to fit inside NUM_BITS_AMOUNT*2 bits.
+// Max ratio between weights guaranteed to be supported is 2**(96*2-14)/(10**18)
```
yueawang, line 497: what if get rid of this check?

yueawang (approved):

dong77 (approved):

packages/loopring_v3/circuit/Gadgets/MatchingGadgets.h
Brechtpd: The max number of bits of `ratio` is NUM_BITS_AMOUNT + NUM_BITS_AMOUNT + 60 (log2(baseFixed)) = 252`, which just fits inside a single field element. We still have to do `ratio * BASE_BIPS` which can increase the required number of bits by `14 = log2(10000)`, which could overflow because the results needs `252 + 14 = 266` bits. The highest we can make this is `NUM_BITS_FIELD_CAPACITY - 14`. But NUM_BITS_AMOUNT*2 seems more than large enough, no? If the weights are equal the absolute max ratio is just (NUM_BITS_AMOUNT + 60) bits. And even in this absolute worst case the weights would have to differ by a factor of 2^18.

PR #2074, opened 2020-12-05
[Loopring_v3] demostrate how l2 signature checking can be used in AMM

dong77:
Brecht, I created this draft PR to demostrate an idea I have.

As of now, the wallet has to use L1 approval-hash for joining and existing pools (as the owner address may change so its signature provided to the relayer is not supported). The idea is to add a new circuit that only verifies EDDSA signatures for arbitrary bytes32 hash: if the EdDSA signature is valid, the signed hash will be included in calldata which can be parsed and used by AMM. With such a small circuits, we can allow wallets to authorize AMM join/exists using a L2 request.


Brechtpd:
I think this makes sense. Biggest benefit I assume we want of this is that no extra tx is needed so no extra delays to wait on the authorize Ethereum tx. Gas use wise maybe not that much benefit, but seems like that is much less important. And because we are the only operator we could still monitor lock/owner change events and stop transactions when possible when that happens (similar to other L2 transactions I guess). And doing things on our smart wallet level specifically is still not a general solution for people using other smart wallets, so having some general L2 solution we could always use makes a lot of sense.

Will continue work on the implementation, should be fairly small changes overall.

Brechtpd:
In theory the withdrawal transaction can already be used to sign arbitrary data that can be verified with EdDSA (data can be put in `extraData` which is than hashed to `onchainDataHash` which can be signed with EdDSA), so doing a withdrawal of 0 could also be used for this, but of course a bit stranger to do things like that. So probably a specific circuit still makes sense.

PR #2094, opened 2020-12-14
[Loopring V3] check LP token burn amount nonzero

dong77:


Brechtpd (approved):

Brechtpd:
I think everything is fine with this additional check because for forced exits we do check that `burnAmount > 0` so there is no problem needing to process a forced exit with `burnAmount == 0`.

But I think (like you said) the check isn't really needed right? Just an additional check for not doing useless exits, but the operator wouldn't do those any way.

PR #2124, opened 2020-12-29
[Loopring36] not emit TransactionApproved event in AmmPool 

kongliangzhong:
Currently, we emitted 3 TransactionApproved events for each AMM join/exit request, which are totally useless. Typically there are about 20 AMM join/exit in a block, so will emit about 60 TransactionApproved events by the AMMPool contracts. Each event consumes about 1100 gas, so about 60K gas can be saved for a block.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
+++ b/packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
@@ -586,6 +586,21 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
         }
     }
 
+    function approveTransactionsWithoutEvents(
+        address[] calldata owners,
+        bytes32[] calldata transactionHashes
+        )
+        external
+        override
+        nonReentrant
+    {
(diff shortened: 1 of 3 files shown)


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -586,6 +586,21 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
         }
     }
 
+    function approveTransactionsWithoutEvents(
```
Brechtpd, line 589: Maybe change `approveTransactions` to take an extra boolean argument to log events or not? Or if we don't want to change the existing API use a shared internal function that takes a boolean so the code doesn't need to be duplicated. 

Or just remove the event from `approveTransactions` because it's extremely likely the only variant that will be used any way.

kongliangzhong: Sounds reasonable to me. I prefer just to remove the event from approveTransactions.

Brechtpd (approved):

PR #2136, opened 2021-01-07
[protocol3.6] AMM pool txs optimization

Brechtpd:
Only change for the relayer is that the AMM transactions verified in the pool contracts need to have `approved = true` in `AuxiliaryData`, and otherwise `approved = false`. `bool  approved` is a new field in `AuxiliaryData`.

This seems to save around 25% of the gas cost of a join/exit.

### DIFF
--- a/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
+++ b/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
@@ -98,8 +98,9 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
         if (config.blockCallbacks.length > 0) {
             require(config.receivers.length > 0, "MISSING_RECEIVERS");
             IAgentRegistry agentRegistry = IExchangeV3(target).getAgentRegistry();
+            // Make sure the receiver is authorized to approve transactions
             for (uint i = 0; i < config.receivers.length; i++) {
-                require(agentRegistry.isAgent(address(0), config.receivers[i]), "UNAUTHORIZED_RECEIVER");
+                require(agentRegistry.isUniversalAgent(config.receivers[i]), "UNAUTHORIZED_RECEIVER");
             }
         }
 
@@ -149,7 +150,12 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             require(blockIdx < blocks.length, "INVALID_BLOCKIDX");
             ExchangeData.Block memory _block = blocks[blockIdx];
 
-            _processTxCallbacks(_block, blockCallback.txCallbacks, config.receivers, preApprovedTxs[i]);
+            _processTxCallbacks(
+                _block,
+                blockCallback.txCallbacks,
+                config.receivers,
+                preApprovedTxs[blockIdx]
+            );
         }
 
         // Verify the approved transactions data against the auxiliary data in the block
@@ -158,13 +164,17 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             ExchangeData.Block memory _block = blocks[i];
             ExchangeData.AuxiliaryData[] memory auxiliaryData = _block.auxiliaryData;
             for(uint j = 0; j < _block.auxiliaryData.length; j++) {
+                // Load the data from auxiliaryData, which is still encoded as calldata
                 uint txIdx;
                 bool approved;
                 assembly {
+                    // Offset to auxiliaryData[j]
                     let auxOffset := mload(add(auxiliaryData, add(32, mul(32, j))))
+                    // Load `txIdx` (pos 0) and `approved` (pos 1) in auxiliaryData[j]
                     txIdx := mload(add(add(32, auxiliaryData), auxOffset))
                     approved := mload(add(add(64, auxiliaryData), auxOffset))
                 }
+                // Check that the provided data matches the expected value
                 require(_preApprovedTxs[txIdx] == approved, "PRE_APPROVED_TX_MISMATCH");
             }
         }
@@ -196,6 +206,7 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
                 0,
                 txCallback.numTxs
             );
+
             for (uint j = txIdx; j < txIdx + txCallback.numTxs; j++) {
                 preApprovedTxs[j] = true;
             }
@@ -236,13 +247,10 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             offset += 32;
             uint blockDataOffset = data.toUint(offset);
             offset += 32;
-
             // Skip over proof
             offset += 32 * 8;
-
             // Skip over storeBlockInfoOnchain
             offset += 32;
-
             uint auxiliaryDataOffset = data.toUint(offset);
             offset += 32;
 
@@ -256,6 +264,7 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             assembly {
                 auxiliaryData := add(data, add(32, add(blockOffset, auxiliaryDataOffset)))
             }
+            // Still encoded as calldata!
             _block.auxiliaryData = auxiliaryData;
         }
         return blocks;
--- a/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
+++ b/packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
@@ -97,8 +97,9 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
     {
         if (config.blockCallbacks.length > 0) {
             require(config.receivers.length > 0, "MISSING_RECEIVERS");
-            IAgentRegistry agentRegistry = IExchangeV3(target).getAgentRegistry();
+
             // Make sure the receiver is authorized to approve transactions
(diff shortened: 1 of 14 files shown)


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -168,6 +196,9 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
                 0,
                 txCallback.numTxs
             );
+            for (uint j = txIdx; j < txIdx + txCallback.numTxs; j++) {
+                preApprovedTxs[j] = true;
```
kongliangzhong, line 200: Is it possible that preApprovedTxs[j] == false? If not, then there is unnecessary to build this preApprovedTxs array, just check auxilaryData.approved == true will be enough.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
```diff
@@ -48,6 +48,7 @@ library ExchangeData
     struct AuxiliaryData
     {
         uint  txIndex;
+        bool  approved;
```
dong77, line 51: Maybe we should merge `txIndex` and `preApproved` into one single `unit` field.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -114,10 +116,11 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             "INVALID_DATA"
         );
 
+        // Decode the blocks
```
dong77, line 119: add a comment:

// Decode the blocks so `_beforeBlockSubmission` next line may modify the auxiliary data fields to reflect pre-approvals.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: We need to make sure `approved` is only true for transactions processed and verified in the pool contracts. If the conditional transactions isn't checked there the value has to be false so the transaction is still verified by the exchange contract.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -114,10 +116,11 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+        ExchangeData.Block[] memory blocks = _decodeBlocks(decompressed);
```
dong77, line 120: Brecht, maybe we should check the 'approved' field in the aux data must always be FALSE inside _decodeBlocks. (or we simply assume they are all force without loading them). Then in _beforeBlockSubmission, we allow the callbacks to change some tx's 'approved' field to true.

Inside the Loopring protocol, if approved == true, we skip, otherwise, we through.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -139,14 +149,32 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
             require(blockIdx < blocks.length, "INVALID_BLOCKIDX");
             ExchangeData.Block memory _block = blocks[blockIdx];
 
-            _processTxCallbacks(_block, blockCallback.txCallbacks, config.receivers);
+            _processTxCallbacks(_block, blockCallback.txCallbacks, config.receivers, preApprovedTxs[i]);
+        }
+
+        // Verify the approved transactions data against the auxiliary data in the block
+        for (uint i = 0; i < blocks.length; i++) {
```
dong77, line 156: I think we my previous approach, we can skip this.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -128,6 +131,13 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
         )
         private
     {
+        // Allocate memory to verify transactions that are approved here
+        bool[][] memory preApprovedTxs = new bool[][](blocks.length);
```
dong77, line 135: With my improvement suggestion above, we don' have to use an im-memory preApprovedTxs any more, we can modify the data in `blocks` dirctly.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
Brechtpd: I don't think that will do much. It may save a very small bit of gas in calldata (zeros, which may also be compressed away), but then we have to pay a small gas cost to encode/decode the 2 values. Both are tiny gas costs. So I think gas savings aren't really there, so cleaner I think not to do so.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
kongliangzhong: Okay, that makes sense.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: I think that's basically what is done. The intermediate array `preApprovedTxs` is used to make this easier with values that are new'd are initialized to zero and then values there are set to true. And then all those values are checked against the data passed in without modifying any block data in the smart contracts. We could start modifying the data in the block, but seems less clean to do it like that. Unless this would make the relayer much simpler, not having to set the correct `approved` value, but I would think not really.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: The code gets a quite a bit more complicated if we do that (blocks without callbacks need all approved values false, gaps between callbacks need to be checked that the approved is false, all this needs to be checked so we have to careful with all possible scenarios), the temp arrays makes this easy and simple. 

But certainly possible, would have to measure it to see if it's worth the extra complexity. I still like not modifying the block data on the fly in the smart contracts if not necessary, but if it makes things better/more efficient than sure.

There reason I didn't just pass the data in a new separate array into submitBlocks next to the block data is that makes calling submitBlocks practically impossible to do efficiently. Now we just have to pass in a single bytes array which can be done very efficiently with fastCallAndVerify, even with large data. And so the data needs to be part of the initial bytes array somehow.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -139,14 +149,32 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+            ExchangeData.AuxiliaryData[] memory auxiliaryData = _block.auxiliaryData;
+            for(uint j = 0; j < _block.auxiliaryData.length; j++) {
+                uint txIdx;
+                bool approved;
+                assembly {
+                    let auxOffset := mload(add(auxiliaryData, add(32, mul(32, j))))
+                    txIdx := mload(add(add(32, auxiliaryData), auxOffset))
+                    approved := mload(add(add(64, auxiliaryData), auxOffset))
+                }
+                require(_preApprovedTxs[txIdx] == approved, "PRE_APPROVED_TX_MISMATCH");
```
dong77, line 168: Here the AMM contract verifies that non-AMM related transaction's `approved` field in auxiliary data must be false. My point is that this should really be done inside the core protocol while the data is loaded. If there is no AMM module at all, then the Loopring contract shall make sure all `approved`  fields in the auxiliary data must be false. This logic belongs to the core protocol, not the AMM contracts.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: Oh right, and the temp array also solves the problem of needing to lookup the tx in the aux data. There's no direct mapping between txIdx <-> auxIdx, so without the temp array we'd need some extra lookup data that needs to be sent onchain by the relayer or extra loops to find the necessary location (which could be quite expensive).

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: Not sure I understand this. Do you inlclude `LoopringIOExchangeOwner` as an AMM contract?

It is a critical the data is checked, but we already depend on `LoopringIOExchangeOwner` working correctly, calling the AMM pools as expected, so already critical for security. And this is an additional critical thing.  I think it's fine doing it in `LoopringIOExchangeOwner`, there's no real need having the mechanism for pre approved transactions to be built-in. And other external contracts like the agent contract as equally critical for security. But you're certainly right that the protocol cannot be used any more without an owner contract that checks these values, which is a bit special I guess.

dong77: Brecht, I think I'm fine with your approach. Is it possible to measure the gas reduction before we test this new version in our testing environment?

Brechtpd: Gas consumption in my benchmark with 2 joins went down from 622,771 gas to 498,616 gas.

Looks like the overhead for a join is now around 80,000 gas. So I think a reduction of around 40% on the join/exit part.

dong77 (approved):

PR #2156, opened 2021-01-24
fast withdrawal refactor

kongliangzhong:
Still in draft, I haven't tested it yet, but I had expressed my design, so maybe you can take an early look.

### DIFF
--- a/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent2.sol
+++ b/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent2.sol
@@ -49,7 +49,6 @@ contract FastWithdrawalAgent2 is ReentrancyGuard, IAgent, Claimable
     using ERC20SafeTransfer for address;
     using MathUint          for uint;
 
-    uint public feeBips = 0;
     uint constant public MAX_FEE_BIPS = 200;
     uint constant public FEE_BIPS_BASE = 10000;
 
@@ -59,6 +58,7 @@ contract FastWithdrawalAgent2 is ReentrancyGuard, IAgent, Claimable
         address to,
         address token,
         uint96  amount,
+        uint96  feeBips,
         address provider,
         bool    success
     );
@@ -71,14 +71,6 @@ contract FastWithdrawalAgent2 is ReentrancyGuard, IAgent, Claimable
         address token;
         uint96  amount;
         uint32  storageID;
-    }
-
-    function updateFeeBips(uint newFeeBips)
-        external
-        onlyOwner
-    {
-        require(newFeeBips <= MAX_FEE_BIPS, "EXCEEDE_MAX_VALUE");
-        feeBips = newFeeBips;
     }
 
     // This method needs to be called by any liquidity provider
@@ -105,13 +97,14 @@ contract FastWithdrawalAgent2 is ReentrancyGuard, IAgent, Claimable
             withdrawal.exchange != address(0) &&
             withdrawal.from != address(0) &&
             withdrawal.to != address(0) &&
-            withdrawal.amount != 0,
+            withdrawal.amount != 0 &&
+            withdrawal.feeBips <= MAX_FEE_BIPS,
             "INVALID_WITHDRAWAL"
         );
 
         // The liquidity provider always authorizes the fast withdrawal by being the direct caller
         address payable liquidityProvider = msg.sender;
-        uint transferAmount = withdrawal.amount * (FEE_BIPS_BASE - feeBips) / FEE_BIPS_BASE;
+        uint transferAmount = withdrawal.amount * (FEE_BIPS_BASE - withdrawal.feeBips) / FEE_BIPS_BASE;
 
         bool success;
         // Override the destination address of a withdrawal to the address of the liquidity provider
--- a/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent2.sol
+++ b/packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent2.sol
@@ -1,0 +1,166 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "../../core/iface/IAgentRegistry.sol";
+import "../../core/iface/IExchangeV3.sol";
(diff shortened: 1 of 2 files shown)


packages/loopring_v3/contracts/aux/agents/FastWithdrawalAgent2.sol
```diff
@@ -0,0 +1,166 @@
...
+/// @author Kongliang Zhong - <kongliang@loopring.org>
+/// @author Daniel Wang - <daniel@loopring.org>
+contract FastWithdrawalAgent2 is ReentrancyGuard, IAgent, Claimable
+{
+    using AddressUtil       for address;
+    using AddressUtil       for address payable;
+    using ERC20SafeTransfer for address;
+    using MathUint          for uint;
+
+    uint public feeBips = 0;
```
Brechtpd, line 52: Probably better to just make this a parameter in the `Withdrawal` structs. SLOADs already cost 800 gas, but in Berlin (HF in ~6 weeks) will cost [2100 gas](https://eips.ethereum.org/EIPS/eip-2929), so it'll be cheaper to pass onchain each time. Added benefit that the fee can be changed more dynamically (perhaps based on gas costs).

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -81,21 +114,97 @@ contract FastWithdrawalLiquidityProvider is ReentrancyGuard, OwnerManagable
...
+        uint _balance = 0;
         if (token == address(0)) {
-            to.sendETHAndVerify(amount, gasleft()); // ETH
+            _balance = address(this).balance;
+            address(this).sendETHAndVerify(amount, gasleft()); // ETH
         } else {
-            token.safeTransferAndVerify(to, amount);  // ERC20 token
+            _balance = ERC20(token).balanceOf(address(this));
+            token.safeTransferAndVerify(address(this), amount);  // ERC20 token
         }
```
Brechtpd, line 132: Make function `payable` and just check amount received for ETH, use `safeTransferFromAndVerify` for ERC20.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -81,21 +114,97 @@ contract FastWithdrawalLiquidityProvider is ReentrancyGuard, OwnerManagable
...
+        userTokenBalances[msg.sender][token] = userTokenBalances[msg.sender][token].add(amount);
+
+        uint _sharesToAdd = 0;
+        uint _totalShares = tokenShares[token];
+        if (_totalShares == 0) {
+            _sharesToAdd = 100 * SHARE_BASE;
+        } else {
+            _sharesToAdd = _totalShares.mul(amount) / _balance;
```
Brechtpd, line 140: `_balance` is the current balance in the contract, but there is a very good chance that some funds are currently loaned out for fast withdrawals, so `_balance` is not the actual total balance deposited (+ fees earned). I think that would make this calculation give the depositer more shares than he should have (in the worst case when almost all funds are loaned out this would print almost an infinite number of shares for the depositor).

The fact that the idea is to earn fees directly on this contract makes it very hard to track if funds are currently loaned out or not (without making the fast withdrawals more expensive by storing extra state onchain, which wouldn't make sense I think). The only solution I currently see is that `_sharesToAdd = amount` in all cases, so 1:1 mapping, which at least guarantees that the amount deposited is correctly tracked. But that still doesn't solve how fees can be paid out directly (only the correct share of the fees can be deduced when there are no loans for the token). Perhaps it is good enough in the contract to check for `balance >= totalDeposited` as the check for no open loans, and distribute the difference between depositors, but this would not always be correct of course.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -81,21 +114,97 @@ contract FastWithdrawalLiquidityProvider is ReentrancyGuard, OwnerManagable
...
+        external
+        onlyNotShutdown
+    {
+        WithdrawalRequest memory _request = withdrawalRequests[requestHash];
+        require(_request.timestamp > 0, "WITHDRAWAL_REQUEST_NOT_EXIST");
+        require(block.timestamp > _request.timestamp + MAX_AGE_WITHDRAWAL_REQUEST_UNTIL_SHUTDOWN,
+                "WITHDRAWAL_REQUEST_NOT_TOO_OLD");
+        shutdown = true;
+        emit LiguidityProviderShudown(address(this), block.timestamp);
     }
```
Brechtpd, line 208: Not sure how useful this is because we can always just never return the amount loaned. So no impact on trust with or without this force withdrawal feature. So depositors could simply be asking for a forced withdrawal on a contract with 0 funds.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
```diff
@@ -140,7 +249,7 @@ contract FastWithdrawalLiquidityProvider is ReentrancyGuard, OwnerManagable
             isManager(approval.signer);
     }
 
-    receive() payable external {}
+    // receive() payable external {}
```
Brechtpd, line 252: Still necessary to receive ETH back from the exchange.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
kongliangzhong: good point, I didn't think of it. I think this part of logic should be controlled by the relayer, not the smart contract.

packages/loopring_v3/contracts/aux/fast-withdrawal/FastWithdrawalLiquidityProvider.sol
kongliangzhong: Yes, I agree. I will rethink the whole design.

kongliangzhong:
@Brechtpd, I discussed with Daniel and Hoss, we will do this in a centralized way. This PR is closed.

PR #2162, opened 2021-01-26
[protocol 3.6] Added Chi gas token support

Brechtpd:
Info: 
- https://1inch-exchange.medium.com/1inch-introduces-chi-gastoken-d0bd5bb0f92b
- https://1inch-exchange.medium.com/everything-you-wanted-to-know-about-chi-gastoken-a1ba0ea55bf3

One important note in the above article:
>One other thing. When you mint Chi tokens and indicate a low gas price, your wallet will be blocked. Therefore, we advise that you use a separate wallet specifically for minting Chi tokens.

So I did not implement any minting functionality into the owner contract. Tokens should be minted elsewhere or bought.

Used `chi.free()` to minimize gas costs, but gas tokens need to be owned by `LoopringIOExchangeOwner`. If we want to share gas tokens with e.g. wallet relayer we could also use `freeFrom` and store the gas tokens in some shared contract. The `xUpTo()` variants have a check that the amount of gas tokens burned can never be greater than the amount of gas tokens owned by the burner. Because we pass in a `maxGasTokensToBurn` we can save some gas by letting the relayer check the balance. Of course, all of this depends on the actual relayer implementation and can be modified if needed .

There's a bit of gas that could be saved by removing the token functionality of the ChiToken, basically having a specialized version so we don't have to keep track of balances. But probably not really worth it.

Should be tested more to see if this actually works.


packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -50,9 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
 
-    constructor(address _exchange)
+    // See:
+    // - https://github.com/1inch-exchange/1inchProtocol/blob/a7781cf9aa1cc2aaa5ccab0d54ecbae1327ca08f/contracts/OneSplitAudit.sol#L343
+    // - https://github.com/curvefi/curve-ren-adapter/blob/8c1fbc3fec41ebd79b06984d72ff6ace3198e62d/truffle/contracts/CurveExchangeAdapter.sol#L104
+    modifier discountCHI(uint maxToBurn)
+    {
+        uint gasStart = gasleft();
+        _;
+        uint gasSpent = 21000 + gasStart - gasleft() + 16 * msg.data.length;
```
dong77, line 62: add the following after `_;`

```
if (maxToBurn == 0) return;
```

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -50,9 +52,28 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+        uint fullAmountToBurn = (gasSpent + 14154) / 41947;
+        uint amountToBurn = fullAmountToBurn > maxToBurn ? maxToBurn : fullAmountToBurn;
+        if (amountToBurn > 0) {
+            chi.free(amountToBurn);
```
dong77, line 66: Please use freeFrom

dong77:
> Info:
> 
> * https://1inch-exchange.medium.com/1inch-introduces-chi-gastoken-d0bd5bb0f92b
> * https://1inch-exchange.medium.com/everything-you-wanted-to-know-about-chi-gastoken-a1ba0ea55bf3
> 
> One important note in the above article:
> 
> > One other thing. When you mint Chi tokens and indicate a low gas price, your wallet will be blocked. Therefore, we advise that you use a separate wallet specifically for minting Chi tokens.
> 
> So I did not implement any minting functionality into the owner contract. Tokens should be minted elsewhere or bought.
> 
> Used `chi.free()` to minimize gas costs, but gas tokens need to be owned by `LoopringIOExchangeOwner`. If we want to share gas tokens with e.g. wallet relayer we could also use `freeFrom` and store the gas tokens in some shared contract. The `xUpTo()` variants have a check that the amount of gas tokens burned can never be greater than the amount of gas tokens owned by the burner. Because we pass in a `maxGasTokensToBurn` we can save some gas by letting the relayer check the balance. Of course, all of this depends on the actual relayer implementation and can be modified if needed .
> 
> There's a bit of gas that could be saved by removing the token functionality of the ChiToken, basically having a specialized version so we don't have to keep track of balances. But probably not really worth it.
> 
> Should be tested more to see if this actually works.

Lets use burnFromUpTo for now.

dong77:
Brecht, our wallet has multiple operators, our exchange has one. Maybe we can have a bridge operator that will do all gas-token related work, so all these operators will go through this bridge to call the target contracts (including LoopringIOExchangeOwner and all smart wallet modules). Then we don't have to modify any application specific contracts at all, the downside it's a bit more expensive as we have another contract call, but we can also allow these operators to call the original target contracts directly.

```solidity
contract ChiBridge {
    function callTarget(address target, bytes data, address chiBurnFrom, uint maxChiToBurn) ...
}
```

Brechtpd:
Yes, that should work fine as well. For the wallet that would the best approach I think.

I tried to avoid it here because it does add some overhead which can be easily avoided because we already have this one special contract. And call costs [will increase soon from 700 gas to 2600 gas](https://eips.ethereum.org/EIPS/eip-2929) in the Berlin HF in ~6 weeks. So the overhead for the exchange will probably be something like 5,000 - 7,000 gas. This is also why I initially avoided using the `upTo` variant that verifies not more is spent than the balance. SLOADs will also increase from 800 gas to 2100 gas, so should be avoided even more than currently if possible. So by doing some things slightly different the gas cost can quickly change +-10k gas.

Brechtpd:
[estimateGas](https://web3js.readthedocs.io/en/v1.2.11/web3-eth-contract.html#methods-mymethod-estimategas) cannot be used directly to find the best amount of gas tokens that can be burned because it returns the minimum amount of gas needed by the transaction. Because refunds are subtracted this does not match the actual gas used at the end.

It's pretty simple to modify the function locally in ganache so I modified it to return the gas used by the tx instead to do some testing. However, I'm currently not getting the expected results and burning the tokens is always more expensive because I don't see the amount of gas refunds increase when burning tokens... Don't yet know why that's the case.

kongliangzhong:
I agree, For exchange, we can approve chi token to the LoopringIOExchangeOwner contract, For wallet, we need a bridge contract, upgrade all wallet is too expensive.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -50,9 +53,35 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+    modifier discountCHI(address from, uint maxToBurn)
+    {
+        uint gasStart = gasleft();
+        _;
+        if (maxToBurn == 0) return;
+        uint gasSpent = 21000 + gasStart - gasleft() + 16 * msg.data.length;
+        uint fullAmountToBurn = (gasSpent + 14154) / 41947;
+        uint amountToBurn = fullAmountToBurn > maxToBurn ? maxToBurn : fullAmountToBurn;
+        if (amountToBurn > 0) {
+            chi.freeFromUpTo(from, amountToBurn);
```
dong77, line 68: How about adding this extra logic:

``` if (from == address(0)) {
    chi.freeUpTo(amountToBurn)
} else {
...
}
```

Just in case we later move chi token into this account.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -50,9 +53,35 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+        }
```
dong77, line 69: how about adding one extra logic: if 'from' is 0x0, use freeUpTo. Just in case.

dong77:
Brecht, could you please also create the ChiBridge contract for the wallet operators in the same PR?

Brechtpd:
> Brecht, could you please also create the ChiBridge contract for the wallet operators in the same PR?

Do you think it makes sense just to modify `BatchTransactor`? I think this should work fine because it would be basically the same thing as the `ChiBridge` just with batching built in as well. It's also fully general and can be used for everything, so it could be used to directly store all Chi tokens as well.

dong77:
> > Brecht, could you please also create the ChiBridge contract for the wallet operators in the same PR?
> 
> Do you think it makes sense just to modify `BatchTransactor`? I think this should work fine because it would be basically the same thing as the `ChiBridge` just with batching built in as well. It's also fully general and can be used for everything, so it could be used to directly store all Chi tokens as well.

Let me ask Kongliang and Yanyan.

Brechtpd:
Found the problem with the refunds. The Chi token contract contains hard coded values with the mainnet address, so it doesn't function correctly in any other contract. I changed the code to make it work in any contract, but because the code depends on a very compact contract address (`0x0000000000004946c0e9F43F4Dee607b0eF1fA1c`) with lots of zeros to minimize costs it cannot be modified directly without impacting the gas costs and stay fully functional. I changed it such a way that gas costs remain the same, but it can only be used for testing because anyone is able to destroy the minted contracts (and not just the token contract).

My testing (with modified `estimageGas`) shows that the normal formula actually works very well. I added some options to make the onchain calculation more accurate: 
- `expectedGasRefund`: The amount of gas refunded by the tx itself. Setting this to `0` will work well, but a better approximation will be `numDeposits * 15,000`.
- `calldataCost`: The gas cost for the calldata itself. This is calculated as `16 * msg.data.length` in the contract when set to `0`, but can be calculated perfectly offchain (16 gas for non-zero byte, 4 gas for zero byte).

`uint fullAmountToBurn = (gasSpent - gasRefundOffset) / 41947` will round down to not waste gas tokens. To make it round up `calldataCost` can be increased by [0, 41946] (e.g. when gas prices are very high even a gas token that only provides a smaller refund may still be worth it).

Both  `expectedGasRefund` and  `calldataCost` can just be set to 0 and everything will work fine.


Did not get any answer yet about `BatchTransactor`, but I added gas token support to `BatchTransactor` because even if not wanted directly in that contract it will work similarly in a separate contract.


For getting the best number of gas tokens to burn we need some kind of transaction simulation on mainnet:
- Tenderly supports transaction simulation, but to use it with the API we would need Tenderly Pro which costs $500/month.
- If we run our own node software we could modify `estimateGas` to have an option to return the actual amount of gas used (instead of min needed). Quite easy, but probably not a very good to change node software.
- ...

But this will likely only bring small improvements if any, so not that important I think. If we know the best amount somehow the current code can be used to simply use that amount.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -50,9 +53,48 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+    // - https://github.com/1inch-exchange/1inchProtocol/blob/a7781cf9aa1cc2aaa5ccab0d54ecbae1327ca08f/contracts/OneSplitAudit.sol#L343
+    // - https://github.com/curvefi/curve-ren-adapter/blob/8c1fbc3fec41ebd79b06984d72ff6ace3198e62d/truffle/contracts/CurveExchangeAdapter.sol#L104
+    modifier discountCHI(GasTokenConfig memory config)
+    {
+        uint gasStart = gasleft();
+        _;
+        if (config.maxToBurn == 0) return;
+        uint gasSpent = 21000 + gasStart - gasleft() + 14154;
+        gasSpent += (config.calldataCost == 0) ? 16 * msg.data.length : config.calldataCost;
+        uint gasRefundOffset = (config.expectedGasRefund * 2 > gasSpent) ? gasSpent : config.expectedGasRefund * 2;
```
dong77, line 73: If `config.expectedGasRefund`== 0, then `fullAmountToBurn` will be `gasSpent/41947`. should it be `gasSpent /(2*41947)` instead as we can only refund up to 50% of `gasSpent`

Brechtpd: The 41947 already multiplies the gas refunded by 2 for the 50% limit. Every gas token theoretically refunds 24,000 gas (refund for a self destruct), but there are some additional overheads as well (a CALL most notably) so it's more around 21,000 gas. `41947` is the value used by 1inch and I kept it. CALL cost will increase soon in the HF so the constant will likely need updating then for best estmate.

dong77 (approved):

dong77 (dismissed):

dong77:
Brecht, I don't think we need to push the saving of gas to the extreme, lets start quick-and-dirty.

kongliangzhong (approved):

dong77 (approved):

PR #2165, opened 2021-01-29
Chi gastoken improve

dong77:


packages/hebao_v1/contracts/aux/BatchTransactor.sol
```diff
@@ -5,48 +5,21 @@ pragma experimental ABIEncoderV2;
...
-        bytes[] calldata txs,
-        uint[]  calldata gasLimits,
-        uint    numGasTokensToBurn
+        bytes[]   calldata txs,
+        uint[]    calldata gasLimits,
+        ChiConfig calldata chiConfig
         )
         external
-        discountCHI(numGasTokensToBurn)
+        discountCHI(chiConfig)
```
Brechtpd, line 22: Will need to have some check like `require(isManager(msg.sender), "UNAUTHORIZED");` otherwise anyone can just burn gas tokens burnable by this contract.

`onlyManager` here could make sense here if this contract would used as some privileged contract in the called contract, otherwise `require(isManager(msg.sender), "UNAUTHORIZED");` only when gas tokens are burned will be a bit more efficient.

kongliangzhong (approved):

PR #2204, opened 2021-02-10
[protocol 3.6] Added hard coded vk support to BlockVerifier

Brechtpd:
Created a python script to write the verification keys into solidity code so the keys don't have to be loaded from storage. This should be the only way we use BlockVerifier in the future. If we want to update the verification keys we should deploy a new contract instead of adding/removing keys dynamically. `immutable` can't be used for arrays so created this script instead.

The existing code to add/remove keys dynamically is still fully present and can be used as a backup if needed, but is there mainly to minimize changes and for testing.

Wanted to use the trusted setup keys we currently use, but I can't immediately find them. Please let me know where to find them @kongliangzhong and I'll update the generated contract with them.

Currently saves ~15k gas, but mostly done to prepare for the Berlin HF which would have increased the gas cost by ~45k gas.


packages/loopring_v3/contracts/core/impl/VerificationKeys.sol
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
+
+
+/// @title Hard coded verification keys
+/// @author Brecht Devos - <brecht@loopring.org>
+library VerificationKeys
+{
```
dong77, line 9: Maybe the script should output a timestamp as a version in a comment.

dong77:
Please also check in the original key files as part of this PR.

kongliangzhong (approved):

PR #2212, opened 2021-02-14
[protocol 3.6] AMM join/exit optimizations

Brechtpd:
Reduces the cost of AMM joins/exits by ~40%.

- 1 API change needed, again with the auxiliary data. Instead of passing `AuxiliaryData[] auxiliaryData` in a block, that same data is passed abi encoded in a simple `bytes` array so unnecessary decoding done by solidity can be avoided.

- For conditional transactions that have `approved = true` (like transactions needed for AMM), the `AuxiliaryData.data` can be left empty, which can save quite a bit of gas. This is nothing new, but I don't think this is currently being done by the relayer.


packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -201,72 +201,93 @@ library ExchangeBlocks
                 timestamp: header.timestamp
             });
 
+            ExchangeData.AuxiliaryData[] memory block_auxiliaryData;
```
dong77, line 204: Brecht, it seems we still need a `ExchangeData.AuxiliaryData[]` in memory, aren't we still decoding the aux data for all conditional transactions? What do you mean “unnecessary decoding done by solidity can be avoided.”？

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -201,72 +201,93 @@ library ExchangeBlocks
...
+                uint txDataOffset = 0;
 
                 if (txType == ExchangeData.TransactionType.DEPOSIT) {
                     DepositTransaction.process(
                         S,
                         ctx,
                         txData,
                         txDataOffset,
-                        auxiliaryData.data
+                        auxData
```
dong77, line 258: Brecht, do you mean now we simply pass in a "auxData as bytes", so each conditional transction can decide whether to further decode these data or not?  Previously each tx's aux data will be decoded before being passed into `process`.

What if we don't use one single `bytes` but use something like `bytes[]` for aux data, will it simplify logics between line 218 and 231?

packages/loopring_v3/contracts/aux/access/IBlockReceiver.sol
```diff
@@ -11,10 +11,8 @@ import "../../amm/libamm/AmmData.sol";
 abstract contract IBlockReceiver
 {
     function beforeBlockSubmission(
-        ExchangeData.Block memory _block,
-        bytes              memory data,
-        uint                      txIdx,
-        uint                      numTxs
+        bytes              calldata txsData,
+        bytes              calldata data
```
dong77, line 15: We probably need to rename data to something more explicit and add a comment for this method.

packages/loopring_v3/contracts/amm/libamm/AmmBlockReceiver.sol
```diff
@@ -18,50 +18,43 @@ library AmmBlockReceiver
...
-
-        _processPoolTx(S, ctx, _block, data);
+        _processPoolTx(S, ctx, txsData, data);
 
         // Update state
         S._totalSupply = ctx.totalSupply;
 
         // Make sure we have consumed exactly the expected number of transactions
-        require(numTxs == (ctx.txIdx - txIdx), "INVALID_NUM_TXS");
+        require(txsData.length/ExchangeData.TX_DATA_AVAILABILITY_SIZE == ctx.txIdx, "INVALID_NUM_TXS");
```
dong77, line 41: shall we check `lengh === txIdx * TX_DATA_AVAILABILITY_SIZE`. 7/3==2, but 7!= 3*2

packages/loopring_v3/contracts/aux/transactions/TransactionReader.sol
```diff
@@ -15,119 +15,102 @@ import "../../core/impl/libtransactions/WithdrawTransaction.sol";
...
         pure
-        returns (bytes memory data)
     {
-        data = _block.readTransactionData(txIdx);
-        require(txType == ExchangeData.TransactionType(data.toUint8(0)), "UNEXPTECTED_TX_TYPE");
+        _block.data.readTransactionData(txIdx, _block.blockSize, txData);
     }
 
-    function createMinimalBlock(
+    function extractTransactions(
```
dong77, line 98: is `extractTransactions` the plural form of `readTx`? If we call `extractTransactions` with numTransactions === 1, then in `extractTransactions` we have `txData := add(txData, TX_DATA_AVAILABILITY_SIZE)` but in `readTx` we don't. Is this a bug?

dong77: If `extractTransactions` is the plural form of `readTx`, maybe rename it to `readTxs`?

packages/loopring_v3/contracts/aux/transactions/TransactionReader.sol
```diff
@@ -15,119 +15,102 @@ import "../../core/impl/libtransactions/WithdrawTransaction.sol";
...
 library TransactionReader {
-    using BlockReader       for ExchangeData.Block;
+    using BlockReader       for bytes;
     using TransactionReader for ExchangeData.Block;
     using BytesUtil         for bytes;
 
     function readDeposit(
         ExchangeData.Block memory _block,
-        uint txIdx
+        uint txIdx,
```
dong77, line 24: Seams these readXXX functions are not used anywhere.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -201,6 +206,8 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
     {
         uint cursor = 0;
 
+        // Reuse the data when possible to save on some memory alloc gas
+        bytes memory txsData = new bytes(ExchangeData.TX_DATA_AVAILABILITY_SIZE*6);
```
dong77, line 210: why `*6` instead of `*txCallback.numTxs`

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -119,19 +119,17 @@ contract LoopringAmmPool is
     }
 
     function beforeBlockSubmission(
-        ExchangeData.Block memory   _block,
-        bytes              calldata data,
-        uint                        txIdx,
-        uint                        numTxs
+        bytes              calldata txsData,
+        bytes              calldata data
```
dong77, line 123: data =>callbackData?

packages/loopring_v3/contracts/aux/access/IBlockReceiver.sol
```diff
@@ -11,10 +11,8 @@ import "../../amm/libamm/AmmData.sol";
 abstract contract IBlockReceiver
 {
     function beforeBlockSubmission(
-        ExchangeData.Block memory _block,
-        bytes              memory data,
-        uint                      txIdx,
-        uint                      numTxs
+        bytes              calldata txsData,
```
dong77, line 14: Passing in a bytes instead of  ExchangeData.Block is cheaper? I thought they are all in memory and should be both very cheap.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -111,17 +112,18 @@ library AmmExitProcess
...
+        address                      from,
+        uint32                       burnStorageID,
+        bytes                        memory  signature,
+        TransferTransaction.Transfer memory transfer
         )
         internal
         view
     {
-        TransferTransaction.Transfer memory transfer = _block.readTransfer(ctx.txIdx++);
+        TransferTransaction.readTx(txsData, ctx.txIdx++ * ExchangeData.TX_DATA_AVAILABILITY_SIZE, transfer);
```
dong77, line 126: will use + cheaper than *?

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
```diff
@@ -125,12 +126,13 @@ library ExchangeData
     function MIN_TIME_IN_SHUTDOWN() internal pure returns (uint32) { return 30 days; }
     // The amount of bytes each rollup transaction uses in the block data for data-availability.
     // This is the maximum amount of bytes of all different transaction types.
-    function TX_DATA_AVAILABILITY_SIZE() internal pure returns (uint32) { return 68; }
     function MAX_AGE_DEPOSIT_UNTIL_WITHDRAWABLE_UPPERBOUND() internal pure returns (uint32) { return 15 days; }
     function ACCOUNTID_PROTOCOLFEE() internal pure returns (uint32) { return 0; }
```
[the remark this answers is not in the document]
Brechtpd: No particular reason, I just tried to avoid changing things without a specific reason.

packages/loopring_v3/contracts/core/impl/libtransactions/SignatureVerificationTransaction.sol
```diff
@@ -26,20 +26,24 @@ library SignatureVerificationTransaction
...
+        uint         offset,
+        SignatureVerification memory verification
         )
         internal
         pure
-        returns (SignatureVerification memory verification)
     {
         uint _offset = offset;
+
+        require(data.toUint8Unsafe(_offset) == uint8(ExchangeData.TransactionType.SIGNATURE_VERIFICATION), "INVALID_TX_TYPE");
```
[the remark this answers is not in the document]
Brechtpd: These functions are now used directly in the AMM pool contracts as well, so instead of creating extra functions that just do this additional check I just added these checks here because it is cheap and also be a bit cleaner I think (complete transaction data is now read here instead of jumping 1 byte ahead every time).

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
Brechtpd: There are additional hidden costs for complex data structures that solidity does behind the scees, and I wasn't aware it could be that expensive. The main cost that is avoided by passing the data as `bytes` is that solidity does not need to do all the abi decoding -> memory for the complete array. `abi.decode` is quite expensive especially when working on data in memory because a copy is made of all the data and of course the parsing of the abi encoded data. By just telling solidity to handle it like a simple `bytes` array there is very little decoding to do so it's much cheaper.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
Brechtpd: > Brecht, do you mean now we simply pass in a "auxData as bytes", so each conditional transction can decide whether to further decode these data or not? Previously each tx's aux data will be decoded before being passed into process.

At this place there isn't that much difference any more, the solidity costs are there just by using complex `ABIEncoderV2` data types heavily in function arguments. If `approved == true` we can skip decoding the `AuxiliaryData.data` for each tx completely.

> What if we don't use one single bytes but use something like bytes[] for aux data, will it simplify logics between line 218 and 231?

Not totally sure, but I can't see that being cheaper and the logic would remain pretty much the same because `bytes[]` is very similar to `ExchangeData.AuxiliaryData[]`. In any case, we want to also avoid decoding the array elements in `LoopringIOExchangeOwner` so it would essentially be the same there for both. Abi encoded `bytes[]` just wraps the data for each array element in an extra `bytes` encoding vs abi encoded `ExchangeData.AuxiliaryData[]` wraps the complete data in a single extra bytes encoding.

packages/loopring_v3/contracts/aux/transactions/TransactionReader.sol
Brechtpd: > is `extractTransactions` the plural form of `readTx`? If we call `extractTransactions` with numTransactions === 1, then in `extractTransactions` we have `txData := add(txData, TX_DATA_AVAILABILITY_SIZE)` but in `readTx` we don't. Is this a bug?

It is indeed the plural. `txData` is just a local variable in the function and `TX_DATA_AVAILABILITY_SIZE` is added to jump ahead to the new position for the next loop iteration. So for numTransactions == 1 it's not needed, but it doesn't hurt.

packages/loopring_v3/contracts/aux/transactions/TransactionReader.sol
Brechtpd: Yes, I guess we can remove them, although they may still be useful in other use cases (e.g. receipts).

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: This it just the initial memory allocation for the txsData, and for the AMM pools we need 6 transactions so I just set that as the default. But could be made a bit more general perhaps by checking the first array element there instead (or just a bytes array with length 0).

packages/loopring_v3/contracts/aux/access/IBlockReceiver.sol
Brechtpd: Not using `ABIEncoderV2` data structures is always cheaper no matter what. The simpler the data type the better it seems in all cases. So not using a `Block` datatype here (even though by encoding the auxiliary data as `bytes` makes is much simpler) is still a bit cheaper, and these can be called often (instead of a single call to submitBlocks where it doesn't matter much) so if not reallly needed best to avoid it.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
Brechtpd: In theory ADD costs 3 and MUL costs 5 gas, but in practice the difference is not measurable when used only a few times because of all the other costs that are added on top of it.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -210,12 +221,13 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
             uint16 receiverIdx = txCallback.receiverIdx;
             require(receiverIdx < receivers.length, "INVALID_RECEIVER_INDEX");
 
-            ExchangeData.Block memory minimalBlock = _block.createMinimalBlock(txIdx, txCallback.numTxs);
+            if (txsData.length != ExchangeData.TX_DATA_AVAILABILITY_SIZE*txCallback.numTxs) {
```
dong77, line 224: calculate `ExchangeData.TX_DATA_AVAILABILITY_SIZE*txCallback.numTxs` once?

packages/loopring_v3/contracts/aux/transactions/TransactionReader.sol
```diff
@@ -15,119 +15,102 @@ import "../../core/impl/libtransactions/WithdrawTransaction.sol";
...
+        bytes memory txData
         )
         internal
         pure
-        returns (SignatureVerificationTransaction.SignatureVerification memory)
+        returns (SignatureVerificationTransaction.SignatureVerification memory verification)
     {
-        bytes memory data = _block.readTx(txIdx, ExchangeData.TransactionType.SIGNATURE_VERIFICATION);
-        return SignatureVerificationTransaction.readTx(data, 1);
+         _block.readTx(txIdx, txData);
```
dong77, line 83: remove extra leading ' '

dong77 (approved):

dong77 (approved):

yueawang (approved):

Issue #1196, opened 2020-05-24
[Loopring 4.0] Allow minting tokens on layer2 for spot trading with layer1 collaterals.


dong77:
## The challenge
A challenge we have learned from operating Loopring.io is that market makers (MMs) will have to possess a lot of real assets to start providing service, because all tokens need to be real, and the exchange reserve ratio has to be 100%.

## A potential solution
We want to make sure MMs can utilize their assets more efficiently, by reducing the exchange's reserve ratio but still making sure when users wish to withdraw, they can always get what they deserve.

To make this possible, we will need to change Loopring's spot trading protocol as follows (you'll see why):

On the smart contract level (layer1), we track how many tokens each account owe the exchange, and how much collateral each user has. For simplicity, let us assume all collaterals are Ether, but in theory, the collaterals can be any token with decent layer1 volumes.
Allow permissionless offchain withdrawal. This means offchain withdrawals don't need signatures from the account owner -- the funds will always go to the original owner, not any other recipient address.
Take LRC as an example, if Alice owes 0 LRC in the contract, when she withdraws LRC, the withdrawal works as the current design, her address will receive the LRC. If Alice owes 100 LRC on layer1, and withdraws 40 LRC from the exchange, her layer1 balance will become -60LRC, and she receives 0 LRC in her address. If Alice withdraws 1ETH,  our contract will swap (using Kyber or Uniswap) some ETH to cover up to 60LRC, if there are still Ether left, those Ether will be sent to Alice's address; otherwise, there remains a smaller  negative LRC balance on layer1. Note that if Alice owes any amount in any token, any withdrawals will have to pay back the owing token first. The actual delivery of token/ether can happen only when Alice owes nothing to the exchange.

### Scenario 1
 
1. Alice collateralize 1 ETH (=$200) to borrow-and-deposit 1,000,000 (=$2000USD), Her onchain balance is layer1=[-1,000,000LRC, 1ETH], offchain balance is layer2=[1,000,000 LRC]
1. Alice does not trade and withdrawal all LRC. Now her balances are: layer1=[1ETH], layer2=[]. He can also withdraw 1 Ether from layer1.

### Scenario 2
1. Alice collateralize 1 ETH (=$200) to borrow-and-deposit 1,000,000 (=$2000USD), Her onchain balance is layer1=[-1,000,000LRC, 1ETH], offchain balance is layer2=[1,000,000 LRC]
[...]

PR #2218, opened 2021-02-17
[protocol 3.6] Bridge and Converter

Brechtpd:
https://github.com/Loopring/protocols/wiki/L2-Vaults

Initial idea for using flash loans here: https://github.com/Loopring/protocols/issues/1196#issuecomment-633324180

## Bridge

The Bridge works similarly to an AMM pool in that the contract is the owner of an account on layer 2, and on-chain we verify and approve transactions from and to this account. This is the bridge that is used to communicate from and to Loopring. Most  times the Bridge account and contract contains no funds, it's simply used as an intermediate. To do communication with Loopring the bridge allows doing just two things.

### 1. Batch deposits

There is a `batchDeposit` call that can be done by any external contract that simply deposits a number of tokens to the Bridge account on L2. The functions also takes in a list of transfers that need to be done on layer 2 from the bridge account to all other accounts in this specified list. So this way only one normal deposit for each distinct token is done, and all other batched deposits are done by simple transfers from the bridge account to the actual destination address on L2. The specified transfers are hashed and stored on-chain so we can later process them in a block, an event is logged with the data to make fetching this list of transfers easy. A unique `batchID` is assigned to each batched deposit to make sure the hash is unique and can't fail. The operator doesn't have to do all these batch deposits, if they are too old they can be withdrawn from the bridge contract directly as a safety for both the operator (spam) and users (operator stops working). The bridge also allows a small max fee to be used from the batched deposits to pay for the L2 transfers.

### 2. Bridge calls

The bridge also allows calling external contracts efficiently (with funds) using batching with the help of `Connector`s. The idea of the `Connector` is that an integration with some dApp is wanted, but we need to be able to do these efficiently in a batched way for all kinds of different scenarios like:
- Mass migration
- L1 token trading
- L1 liquidity provision
- L1 vault interaction
- L1 lending
- ...

All these can work differently for batching so we leave that up to the connector to implement, and so the bridge just calls the standard function `processCalls` to let the connector do what's necessary. In some cases these connectors just need to do another L1 call and that's it. In other cases, for example a token swap, another token is bought that needs to be returned to the buyer on L2. This is where the batched deposits come in again to allow the connector to do just that in a standard way.

These connectors also need to handle the funds in some way. In all cases all funds are first moved to the Bridge contract. This was chosen as a way to not impose any risks on all funds in the deposit contract, I think it's too risky to even call trusted functions directly on the deposit contract. But the current method is pretty efficient, in many cases there is only a need to do as many L1 token transfers as there are unique tokens to and from the Bridge contract, even if multiple connectors are called.

All connectors can do everything, but there is the concept of trusted and untrusted connectors for efficiency reasons, which gas a small implementation difference:
- The Bridge calls `processCalls` on trusted connectors using a delegate call. This means no additional token transfers are needed to execute the code in the connector, but the connector has access to all funds currently being processed in the bridge.
- The Bridge calls `processCalls` on non-trusted connectors using a real call after transferring the necessary funds to the connector contract. This makes sure that only funds of users that want to interact with that connector are at risk in that connector.

What is a trusted and untrusted connector can be set by the Bridge owner. Implementation wise this is only a very small difference at the Bridge side (the connector doesn't need to know), but for gas efficiency it can be worth it to make the connector trusted after we make sure it's safe. On the other hand, untrusted connectors are also great because users can call any connector without us having to worry about the connector misbehaving and putting other people's funds at risk.

No matter if the connector is trusted or not, connectors are given a max gas limit and are always allowed to fail in the  `processCalls` call. If the call fails the Bridge automatically approves the transfers back from the Bridge to the user's account.

There is also the notion of a connector _group_. This is simply a way that the same connector can be used with all kinds of inputs. For example, for a uniswap connector a group would be the two tokens that need to be swapped, and this data is encoded in `groupData` and grouped together so all these swaps can be handled together.

`userData` on the other hand is for user specific data a connector could us. Think something like max slippage for a swap, or the destination address for a mass migration to a different rollup.

Calls submitted to the connector will be sorted on group and tokens so they can be easily processed.

Similarly to the AMM pools, the user can sign a `BridgeCall` transaction which is verified on-chain using the `SignatureVerification` tx so the L2 keys can be used. So every bridge call requires 2 L2 transactions: an L2 transfer and the signature verification tx. 

#### Fees

When people sign a transaction they don't know how big a batch it will be part of. The operator itself also doesn't always know how much a bridge call will end up costing.

To solve this problem users sign a transaction and a pay fee for an abosolute amount of gas that the operator needs to provide to the batched bridge calls. The total amount of gas the operator needs to provide to the connector is at least the sum of all these gas amounts.

The other problem is that, when just using this mechanism, the operator could submit a single bridge call that is very likely to have low gas amount just by itself, which will fail because the whole operation will take much more gas than what that single bridge call provides. And so there is an additional `IBridgeConnector.getMinGasLimit` which also forces the operator to at least provide the minimum amount of gas required for all the calls by the connector itself.

## Converters

`Converter`s are the implementation of the scheme described [here](https://github.com/Loopring/protocols/issues/2221#issuecomment-783597146). They are very efficient at doing many conversions from tokenA -> tokenB, but each converter needs its own token and contract deployed.

`BaseConverter` handles all the general code that can be reused between all converters (failure logic for the conversion, but also the main vault stuff like the token code, burning, minting, depositing, withdrawing,...). Specific implementations just need to implement a simple `convert` function to do the specific implementation for the conversion. There is no problem with rounding errors for the vault token transfers if the user doesn't pay a fee on this token (which I think wouldn't make a lot of sense) because the vault token is transferred in and out using a float representation.

Everything Converters can do, the Bridge can do as well, but a bit less efficient in some cases (although the Bridge could also be more efficient in others). On the other, the Bridge can do everything converters can do and more (like mass migrations between L2s).

### Converter

The converter can do anything that can be represented by a token <-> token trade, under parameters shared by everyone in the batch. Each conversion from tokenA -> tokenB needs a separate contract.
- Swap tokens on L1 without minimum price or using a minimum price encoded in the converter contract
- Add liquidity to an L1 pool, even with multiple tokens, in return for a token representing the user's share. Yearn, AAVE, ETH staking, ....
- ....

### Bridge

The bridge can do anything the Converter can do and more. For example, the bridge can do anything that can be represented by a token <-> token trade, but now with parameters that can be defined by each user separately. The bridge also supports operations that only require tokens to go out (like mass migration). Only a single `BridgeConnector` contract needs to be deployed to enable a certain kind of L1 operation for all tokens. For example, just a single Uniswap connector is able to convert between any tokens.

Even L1 operations that are not batchable could have some small benefits by using the bridge because the token withdrawals/deposits from/to the exchange can be shared by all bridge operations.

- Swap/Add liquidity on L1 with per user settings (minimum price, minimum pool share,...)
- Calling any L1 function that requires knowing the user's Ethereum address (mass migration,...)
- ....

### Flash minting

- Allow arbitrage on L2 without needing any funds
- Allows batching AMM joins/exits
- ...


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -382,6 +382,44 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
         return state.pendingDeposits[owner][tokenID].amount;
     }
 
+    function flashDeposit(
+        address to,
```
dong77, line 386: this `to` is not used...

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -46,26 +47,53 @@ library ExchangeDeposits
...
-        uint96 amountDeposited = S.depositContract.deposit{value: msg.value}(
-            from,
-            tokenAddress,
-            amount,
-            extraData
-        );
+        uint96 amountDeposited = amount;
+        if (flash) {
+            require(msg.value == 0, "INVALID_FLASH_DEPOSIT");
+            S.amountFlashDeposited[tokenAddress] = S.amountFlashDeposited[tokenAddress].add(amount);
```
dong77, line 53: maybe we should credit these deposits to an operator account *explicitly* instead of allowing the relayer to credit them to any L2 account.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -135,7 +155,31 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
...
+        for (uint i = 0; i < flashConfig.vaults.length; i++) {
+            require(flashConfig.vaults[i].to != target, "EXCHANGE_CANNOT_BE_VAULT");
+            (bool success, ) = flashConfig.vaults[i].to.call(flashConfig.vaults[i].data);
+            require(success, "VAULT_CALL_FAILED");
+        }
+
+        // Make sure flash deposits were repaid
+        for (uint i = 0; i < flashConfig.deposits.length; i++) {
+            require(
+                IExchangeV3(target).getAmountFlashDeposited(flashConfig.deposits[i].token) == 0,
```
dong77, line 179: What if we change 0 to nonzero value? 
```
IExchangeV3(target).getAmountFlashDeposited(flashConfig.deposits[i].token) <= credit[token],
```

This allow us the transaction to success and we just need to manage these credits ...

Brechtpd: I don't think it's a good idea to leave non-existing tokens in the exchange, I think that's much too risky for users because there is no guarantee it would be paid back and the exchange can be left in an insolvent state. 

Is this to solve the cases where we get less tokens than expected? In general we should do the L1 swap at a price that we are sure will get us a bit more tokens than needed to repay the the flash deposit. Doing that will also build up a bit of surplus we can use in the cases where we need a bit more tokens than we expected. I think this is similar to your `credit` system without passing any of the risk to the users. At the start we could also deposit a small amount of tokens ourselves to build up this reserve.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
Brechtpd: I think it is, and it should be. We deposit to some account on L2 so we need an address that will receive the flash deposit and so we process a normal deposit request to this `to` address.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
Brechtpd: Is there a benefit in doing that? It doesn't seem to matter to me who which account receives the tokens, and the implementation would remain the same as far as I can see.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -52,6 +52,25 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
         address[]       receivers;
     }
 
+    struct FlashDepositInfo
```
dong77, line 55: Lets call it `FlashDeposit`

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -52,6 +52,25 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
...
+    {
+        address to;
+        address token;
+        uint96  amount;
+    }
+
+    struct FlashVaultInfo
```
dong77, line 62: Lets call it `FlashVault`?

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -101,7 +120,8 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
         bool                     isDataCompressed,
         bytes           calldata data,
         CallbackConfig  calldata config,
-        ChiConfig       calldata chiConfig
+        ChiConfig       calldata chiConfig,
```
dong77, line 123: ChiConfig is removed

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -135,7 +155,31 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
         // Process the callback logic.
         _beforeBlockSubmission(blocks, config);
 
+        // Do flash deposits
+        for (uint i = 0; i < flashConfig.deposits.length; i++) {
+            IExchangeV3(target).flashDeposit(
```
dong77, line 160: Should we change `flashDeposit` to a plural form to reduce inter-contract calls?

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -135,7 +155,31 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ChiDiscount, ERC
...
+                flashConfig.deposits[i].amount
+            );
+        }
+
         target.fastCallAndVerify(gasleft(), 0, decompressed);
+
+        // Do vault logic
+        for (uint i = 0; i < flashConfig.vaults.length; i++) {
+            require(flashConfig.vaults[i].to != target, "EXCHANGE_CANNOT_BE_VAULT");
+            (bool success, ) = flashConfig.vaults[i].to.call(flashConfig.vaults[i].data);
```
dong77, line 172: Do we need to consider how to do ERC20 authorization for `valuts[i].to` inside this contract?

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: Yep, working on some changes now.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
Brechtpd: Not exactly sure what you mean, but the vault itself can do what it needs to do depending on the token and what is required for the specific case. We don't have to do anything for that here.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -46,26 +47,53 @@ library ExchangeDeposits
...
-        uint96 amountDeposited = S.depositContract.deposit{value: msg.value}(
-            from,
-            tokenAddress,
-            amount,
-            extraData
-        );
+        uint96 amountDeposited = amount;
+        if (flash) {
+            require(msg.value == 0, "INVALID_FLASH_DEPOSIT");
+            S.amountFlashMinted[tokenAddress] = S.amountFlashMinted[tokenAddress].add(amount);
```
dong77, line 53: can tokenAddress == 0x0?

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -46,26 +47,53 @@ library ExchangeDeposits
...
-            to,
             tokenAddress,
-            tokenID,
-            amountDeposited
+            amount,
+            extraData
         );
+
+        // Paid back
+        S.amountFlashMinted[tokenAddress] = S.amountFlashMinted[tokenAddress].sub(amountDeposited);
```
dong77, line 97: should we allow over-payment?

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
```diff
@@ -343,6 +343,53 @@ abstract contract IExchangeV3 is Claimable
...
+        )
+        external
+        virtual;
+
+    /// @dev Repays funds minted using `flashMint`.
+    /// @param from The address that deposits the funds to the exchange
+    /// @param tokenAddress The address of the token, use `0x0` for Ether.
+    /// @param amount The amount of tokens to deposit
+    /// @param extraData Optional extra data used by the deposit contract
+    function repayFlashMint(
```
dong77, line 364: flashRepay?

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
```diff
@@ -343,6 +343,53 @@ abstract contract IExchangeV3 is Claimable
...
+        uint96  amount,
+        bytes   calldata extraData
+        )
+        external
+        virtual
+        payable;
+
+    /// @dev Verifies all minted tokens were paid back.
+    /// @param flashMints The list of flash mints that were done.
+    function verifyFlashMintsPaidBack(
```
dong77, line 376: verifyFlashClearance?

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -383,6 +383,63 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
...
+
+    function verifyFlashMintsPaidBack(
+        ExchangeData.FlashMint[] calldata flashMints
+        )
+        external
+        override
+        view
+    {
+        for (uint i = 0; i < flashMints.length; i++) {
+            require(state.amountFlashMinted[flashMints[i].token] == 0, "FLASH_MINT_NOT_REPAID");
```
dong77, line 439: two lines

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -236,9 +236,9 @@ library ExchangeBlocks
 
                 minTxIndex = txIndex + 1;
 
-                if (approved) {
+                /*if (approved) {
                     continue;
-                }
+                }*/
```
dong77, line 241: remove

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -249,6 +249,10 @@ library ExchangeBlocks
                 );
                 uint txDataOffset = 0;
 
+                if (approved && txType != ExchangeData.TransactionType.WITHDRAWAL) {
```
dong77, line 252: `!=` change to `==`?

packages/loopring_v3/contracts/aux/bridge/IBridge.sol
```diff
@@ -0,0 +1,17 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./BridgeData.sol";
+
+
+interface IBridge
+{
+     function batchDeposit(
+        address                   from,
+        BridgeTransfer[] calldata deposits
```
dong77, line 13: deposits => transfers?

packages/loopring_v3/contracts/aux/bridge/IBridgeConnector.sol
```diff
@@ -0,0 +1,19 @@
...
+
+
+interface IBridgeConnector
+{
+    function processCalls(ConnectorCalls calldata connectorCalls)
+    external
+    payable;
+
+    function getMinGasLimit(ConnectorCalls calldata connectorCalls)
+    external
```
dong77, line 16: format

dong77 (commented):
Only partially done. The rest will take a few more days:(

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+            domainSeparator: DOMAIN_SEPARATOR,
+            accountID: accountID,
+            txIdx: 0
+        });
+
+        _processTransfers(ctx, operations.transferOperations, txsData);
+        _processCalls(ctx, operations, txsData);
+
+        // Make sure we have consumed exactly the expected number of transactions
+        require(txsData.length == ctx.txIdx * ExchangeData.TX_DATA_AVAILABILITY_SIZE, "INVALID_NUM_TXS");
```
dong77, line 174: It seems `txIdx` is a counter, not an index.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+        }
+
+        // Do the call
+        bool success;
+        try Bridge(this)._executeConnectorCall(connectorCalls) {
+            success = true;
+            emit BridgeCallSuccess(keccak256(abi.encode(connectorCalls)));
+        } catch Error(string memory reason) {
+            success = false;
+            emit BridgeCallFailed(keccak256(abi.encode(connectorCalls)), reason);
```
dong77, line 435: make a variable for `keccak256(abi.encode(connectorCalls))`?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+                assembly { revert(add(returnData, 32), mload(returnData)) }
+            }
+            // TODO: maybe return transfers here to batch all of them together in a single `deposit` call across all trusted connectors.
+        } else {
+            // Transfer funds to the external contract with a real call, so the connector doesn't have to be trusted at all.
+
+            // Transfer funds to the connector contract
+            uint ethValue = 0;
+            for (uint i = 0; i < connectorCalls.tokens.length; i++) {
+                if (connectorCalls.tokens[i].amount > 0) {
```
dong77, line 484: `address token  = connectorCalls.tokens[i].token` for better readability

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
Brechtpd: Yes, for ETH.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
Brechtpd: The amount flash minted and repaid is completely deterministic so I don't see a reason why they would ever be different. Unless as some kind of fee, but because we handle the handle the on-chain logic for flash mints that doesn't seem to make a lot of sense, we would just ask for a fee on L2.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
Brechtpd: If it's a withdrawal the token transfer still needs to be done regardless if it's approved or not. If a withdrawal is approved like this we just don't check the normal approval methods.

packages/loopring_v3/contracts/aux/bridge/IBridge.sol
Brechtpd: Depends how you look at it. They are done with transfers internally, but they are all separate deposit operations from the caller's point of view.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+    }
+
+    function batchDeposit(
+        address                 from,
+        BridgeTransfer[] memory deposits
+        )
+        external
+        payable
+    {
+        require(from == msg.sender, "UNAUTHORIZED");
```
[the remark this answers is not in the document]
Brechtpd: Removed it for now. In theory it could also come from somewhere else if the bridge would implement some other authorization method, so more of thing to make the API more general, but if ever needed we can just add another function.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+        // Needs to be possible to do all transfers in a single block
+        require(deposits.length <= MAX_NUM_TRANSACTIONS_IN_BLOCK, "MAX_DEPOSITS_EXCEEDED");
```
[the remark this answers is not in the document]
Brechtpd: Currently the deposit itself does not need to be done through the bridge because it's just a normal deposit request.

There are points that could be made to let the deposit itself also to be done through the bridge, but that complicates things somewhat as well. The current method just puts the gas costs for all the expensive things to the caller of `batchDeposit`, the operator just needs to do the cheap L2 transfers.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+    // Instead of querying the exchange each time, the tokenID
+    // is automatically cached inside this contract to save gas.
+    function _getTokenID(address tokenAddress)
+        internal
+        returns (uint16 cachedTokenID)
+    {
+        if (tokenAddress == address(0)) {
+            cachedTokenID = 0;
+        } else {
+            cachedTokenID = cachedTokenIDs[tokenAddress];
```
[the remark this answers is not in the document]
Brechtpd: Maybe, maybe not. Now it definitely does (needs to go through multiple contracts to get the data). With [eip-2929](https://eips.ethereum.org/EIPS/eip-2929) maybe not.

eip-2929 makes it a bit hard to optimize solidity now. It will go live in 4 weeks so we need to optimize for it, but tools (like truffle) don't support the Berlin HF yet, so not easy to test and measure yet.

eip-2929 makes things more logical where reading/calling things the first time is 3x more expensive, but than all other reads/calls are cheap (compared to the behaviour now where each read/call costs the same no matter how many times it was accessed before). So manually batching some stuff may not be necessary any more. But because there are many hidden costs with solidity only measuring the gas costs on the EVM is to be trusted in practice.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+        TransferTransaction.Transfer transfer;
+        BridgeCall                   call;
+    }
+
+    bytes32 constant public BRIDGE_CALL_TYPEHASH = keccak256(
+        "BridgeCall(address from,address to,uint16 tokenID,uint96 amount,uint16 feeTokenID,uint96 maxFee,uint32 storageID,uint32 minGas,address connector,bytes groupData,bytes userData,uint256 validUntil)"
+    );
+
+    uint               public constant  MAX_NUM_TRANSACTIONS_IN_BLOCK = 386;
+    uint               public constant  MAX_AGE_PENDING_TRANSFERS     = 7 days;
```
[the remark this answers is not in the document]
Brechtpd: The deposit happens immediately in `batchDeposit`, but all L2 transfers to do for the batch don't need to be handled by the operator. So within `MAX_AGE_PENDING_TRANSFERS` seconds the operator can do those transfers, any older and people can withdraw directly from the bridge contract the amount that should have been distributed on L2.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+                transfers[idx].amount,
+                transfers[idx].owner
+            );
+        }
+    }
+
+    // Can be used to withdraw funds that were already deposited to the bridge,
+    // but need to be returned to be able to withdraw from old pending transfers.
+    function forceWithdraw(address[] calldata tokens)
+        external
```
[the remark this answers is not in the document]
Brechtpd: It's to allow withdrawing directly from the Bridge for old batched deposit transfers. Because the deposit already happened users can just call `forceWithdraw` to withdraw funds back to the bridge so the withdrawal can be done on L1.

The operator can still schedule the actual forced withdrawal when it wants, so can be made public without any problem.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,605 @@
...
+
+        // Make sure we have consumed exactly the expected number of transactions
+        require(txsData.length == ctx.txIdx * ExchangeData.TX_DATA_AVAILABILITY_SIZE, "INVALID_NUM_TXS");
+    }
+
+    // Allows withdrawing from pending transfers that are at least MAX_AGE_PENDING_TRANSFERS old.
+    function withdrawFromPendingTransfers(
+        uint                              batchID,
+        InternalBridgeTransfer[] calldata transfers,
+        uint[]                   calldata indices
```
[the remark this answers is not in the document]
Brechtpd: If there's a batch of 386 deposits with lots of different tokens, doing all withdrawals in the batch on L1 will be super expensive and could cost even more than the Ethereum block gas limit.

So if somebody was part of the batched deposit and now wants to withdraw his funds he can just choose to do his own one without having to do all other ones.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: It's used as an index to read all the transactions, but indeed here it is also used as a counter.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Removed it for now and replaced it with just the connector address because `abi.encode(connectorCalls)` is pretty expensive. I think that should be fine, but will have to see what is cheap and useful.

packages/loopring_v3/contracts/lib/LPERC20.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    function permit(
+        address owner,
+        address spender,
+        uint256 value,
+        uint256 deadline,
+        bytes   calldata signature
+        )
+        external
```
dong77, line 119: return a bool?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+        require(txsData.length == ctx.txIdx * ExchangeData.TX_DATA_AVAILABILITY_SIZE, "INVALID_NUM_TXS");
+    }
+
+    // Allows withdrawing from pending transfers that are at least MAX_AGE_PENDING_TRANSFERS old.
+    function withdrawFromPendingBatchDeposit(
+        uint                              batchID,
+        InternalBridgeTransfer[] calldata transfers,
+        uint[]                   calldata indices
+        )
+        external
```
dong77, line 181: for all 'external' functions, maybe we want to add 'nonReentrant’ guard

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
dong77: I see the point. But how about changing `withdrawn[hash][idx]` to `withdrawn[batchID][idx]`? It will be much easier to query.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+                require(tokens[i].amount == msg.value, "INVALID_ETH_DEPOSIT");
+            }
+            _deposit(msg.sender, tokens[i].token, uint96(tokens[i].amount));
+        }
+
+        // Store the transfers so they can be processed later
+        _storeTransfers(transfers);
+    }
+
+    function beforeBlockSubmission(
```
dong77, line 153: this is the bridged call to L2 dapp, is it? I almost confused it with `batchDeposit`.

If `batchDeposit` and `beforeBlockSubmission` are two independed features, maybe we should extract `batchDeposit` to a dedicated file...

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+    {
+        BridgeOperations memory operations = abi.decode(callbackData, (BridgeOperations));
+
+        Context memory ctx = Context({
+            domainSeparator: DOMAIN_SEPARATOR,
+            accountID: accountID,
+            txIdx: 0
+        });
+
+        _processTransfers(ctx, operations.transferBatches, txsData);
```
dong77, line 168: _processTransfers => processBatchDeposit
_processCalls => processConnectorCalls

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+
+    // --- Internal functions ---
+
+    function _processTransfers(
+        Context         memory ctx,
+        TransferBatch[] memory operations,
+        bytes           memory txsData
+        )
+        internal
+    {
```
dong77, line 242: Do we allow the corresponding deposit to be processed in a different block? If only the deposit is processed, but its L2 transfers are not, will users withdraw their tokens with `withdrawFromPendingBatchDeposit`? (in such case, the token is already transfered to DepositContract I guess).


If processing of deposit and transfers must be *atomic*, maybe we should also change `_hashTransfers` to hash the deposit data.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+            // Max rounding error for a float24 is 2/100000
+            // But relayer may use float rounding multiple times
+            // so the range is expanded to [100000 - 8, 100000 + 0],
+            // always rounding down.
+            uint ratio = (uint(amount) * 100000) / uint(targetAmount);
+            return (100000 - 8) <= ratio && ratio <= (100000 + 0);
+        }
+    }
+
+    function _hashTransfers(
```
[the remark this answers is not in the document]
Brechtpd: > this function can take a `TransferOperation` as the only parameter

I optimized this (code is still a mess so didn't push this yet) so only a list of amounts are part of `TransferBatch`, all the other data can also be found directly in the transfer and so the `InternalBridgeTransfer` is recreated using that data as well for hashing.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+            // Verify transfers
+            TransferTransaction.Transfer memory transfer;
+            for (uint i = 0; i < transfers.length; i++) {
+                TransferTransaction.readTx(txsData, ctx.txIdx++ * ExchangeData.TX_DATA_AVAILABILITY_SIZE, transfer);
+
+                uint16 tokenID = transfers[i].tokenID;
+                require(
+                    transfer.fromAccountID == ctx.accountID &&
+                    transfer.to == transfers[i].owner &&
+                    transfer.from == address(this) &&
```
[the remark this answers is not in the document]
Brechtpd: Yes, could be removed, checking the accountID is enough. The address will always be part of the transfer DA though because it's a conditional transfer.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -91,9 +97,11 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
     }
 
     function submitBlocksWithCallbacks(
-        bool                     isDataCompressed,
-        bytes           calldata data,
-        CallbackConfig  calldata config
+        bool                              isDataCompressed,
+        bytes                    calldata data,
+        CallbackConfig           calldata config,
```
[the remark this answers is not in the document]
Brechtpd: Agreed on all names, accept `PostBlocksCallback` => `BridgeCallbacks`. Those callbacks could be any function and are currently used for repaying flash mints for the `Converter' design, not the 'Bridge' (the Bridge doesn't need flash mints). But it could be used to call any function for any reason.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -187,6 +209,23 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
 
+    function _afterBlockSubmission(
+        ExchangeData.Block[] memory   /*blocks*/,
+        PostBlocksCallback[] calldata postBlocksCallbacks
+        )
+        private
+    {
+        for (uint i = 0; i < postBlocksCallbacks.length; i++) {
+            // Disallow calls to the exchange and pre block callback contracts
+            require(postBlocksCallbacks[i].to != target, "EXCHANGE_CANNOT_BE_POST_CALLBACK_TARGET");
```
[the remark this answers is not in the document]
Brechtpd: Can't hurt to add, but in theory this can be allowed.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -187,6 +209,23 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+            require(postBlocksCallbacks[i].data.toBytes4(0) != IBlockReceiver.beforeBlockSubmission.selector, "INVALID_POST_CALLBACK_FUNCTION");
```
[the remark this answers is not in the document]
Brechtpd: Did it like this because we know exactly which function cannot be called, and it's much cheaper than having to check if an address is an agent or not.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -123,13 +131,27 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
         // Decode the blocks
         ExchangeData.Block[] memory blocks = _decodeBlocks(decompressed);
 
-        // Process the callback logic.
-        _beforeBlockSubmission(blocks, config);
```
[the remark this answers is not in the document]
Brechtpd: It may be useful, but I don't know if we should add support for things that we currently don't plan to use (and if we use it maybe it needs additional changes again).

But a boolean before/after may indeed be useful with perhaps very little work, so perhaps still worth it.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeDeposits.sol
```diff
@@ -34,7 +34,8 @@ library ExchangeDeposits
         address to,
         address tokenAddress,
         uint96  amount,                 // can be zero
-        bytes   memory extraData
+        bytes   memory extraData,
+        bool    flash
```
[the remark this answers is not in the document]
Brechtpd: If we do that we'll have to be very careful not to allow normal deposits to have `from = 0x0` as input (of course normally the case). An explicit parameter always set by the protocol seems to be more explicit and safe I think than using `from = 0x0` as a magic value.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeBlocks.sol
```diff
@@ -249,6 +245,12 @@ library ExchangeBlocks
                 );
                 uint txDataOffset = 0;
 
+                if (approved &&
+                    txType != ExchangeData.TransactionType.WITHDRAWAL &&
+                    txType != ExchangeData.TransactionType.DEPOSIT) {
+                    continue;
```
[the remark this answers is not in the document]
Brechtpd: The `approved` system was only used for AMM pool contracts (contracts we fully control), which don't do any deposits/withdrawals so the behavior didn't matter. Now we also use it for the Bridge, and the Bridge needs to approve withdrawals (but only from its own address), so now the behavior does matter for withdrawals a bit (but things like `withdrawalRecipient` still don't really matter because the Bridge is a contract so no practical way we'd use that without EdDSA signatures).

There are 2 options:
1. We skip all checks for deposits/withdrawals in the exchange (previous behaviour). This is possible, because e.g. the bridge can just transfer funds to the deposit contract and then just do the deposit transaction to reflect that. The same with withdrawals, the bridge could in theory have direct access to funds in the deposit contract and the withdrawal transaction again is just done to reflect that change.
2. We still do all normal checks for deposits/withdrawals (current behaviour).

I have opted for 2) because with 1) the security of the complete rollup depends on the security of the bridge (even though it could be more efficient to do 1).

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -383,6 +383,66 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
...
+        view
+        returns (uint96)
+    {
+        return state.amountFlashMinted[tokenAddress];
+    }
+
+    function verifyFlashMintsPaidBack(
+        ExchangeData.FlashMint[] calldata flashMints
+        )
+        external
```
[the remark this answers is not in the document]
Brechtpd: Yes, but maybe making it memory directly could be a bit cheaper. Will have to measure.

packages/loopring_v3/contracts/lib/LPERC20.sol
Brechtpd: Although there is [no finalized standard](https://eips.ethereum.org/EIPS/eip-2612), none of the existing implementations return a value.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+        depositContract = _exchange.getDepositContract();
+        exchangeOwner = _exchange.owner();
+
+        DOMAIN_SEPARATOR = EIP712.hash(EIP712.Domain("Bridge", "1.0", address(this)));
+    }
+
+    function batchDeposit(
+        BridgeTransfer[] memory deposits
+        )
+        external
```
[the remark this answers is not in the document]
Brechtpd: Should be okay to save some gas not having it, the only external function that is called is a hard coded `exchange.deposit`. Even with reentrance I don't think there can be a problem.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
Brechtpd: Do you think there is a problem if the deposit L2 tx is processed in a different block than all other transfers? The current implementation is just limited to do all L2 transfers, which is exactly `deposits.length` transfers, in the same block because the callback only works on a single block. But even that could be solved (but I don't think that will a problem in practice so not worth doing for now).

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+        accountID = _accountID;
+
+        depositContract = _exchange.getDepositContract();
+        exchangeOwner = _exchange.owner();
+
+        DOMAIN_SEPARATOR = EIP712.hash(EIP712.Domain("Bridge", "1.0", address(this)));
+    }
+
+    function batchDeposit(
+        BridgeTransfer[] memory deposits
```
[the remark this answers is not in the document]
Brechtpd: We can implement multiple permutations of `batchDeposit`, depending on the expected input. I assume it won't always be easy to order the deposits per token,  but the current implementation should work fairly well for all inputs, ordered or not.

However, one think to keep in mind is that this function will very likely almost never be called directly in a tx, but called indirectly through some other contract (the bridge itself of course, from another rollup/state channel, etc...). And so the gas that needs to be paid to call this function is not dependent on the calldata cost for the inputs (the well known 4 gas for zero bytes, 16 gas for non-zero bytes), these do not apply for internal contract calls.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: > this is the bridged call to L2 dapp, is it? I almost confused it with batchDeposit.

To L1 dapps yes.

> If batchDeposit and beforeBlockSubmission are two independed features, maybe we should extract batchDeposit to a dedicated file

`batchDeposit` and `BridgeCall`s are quite related. `batchDeposit` is for L1 -> L2, `BridgeCall`s is for L2 -> L1, but often that also requires and L1 -> L2 operation, which is done using  `batchDeposit`. Both are indeed done in the Bridge contract, making it a two-way bridge, which I think makes sense. I think splitting them up could also work, but probably a bit less efficient.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Currently the deposit can be processed in a different block, completely separate to the L2 transfers.

In all cases funds are directly deposited to the deposit contract in `batchDeposit` so they have to be removed from there if the L2 transfers are not processed.
Two scenario's:
- Deposit itself not processed: funds can be withdrawn to the bridge by anyone with `Exchange.withdrawFromDepositRequest`.
- Deposit processed: funds can be withdrawn to the bridge by anyone with `Bridge.forceWithdraw`.

Once the funds are in the Bridge `withdrawFromPendingBatchDeposit` can be used to get the tokens back. So it's definitely not something we actually want normal users to have to go through (similar to how we always process all Exchange deposits), and so we should do all transfers in normal situations. But the biggest deposit cost is paid by the caller of `batchDeposit`, and we only have to do the cheaper L2 transfers, costing ~4,000 gas each. To help pay for that we can use up to `MAX_FEE_BIPS` of the deposited amount. But if people spam `batchDeposit` with very low value deposits we can choose not to process them.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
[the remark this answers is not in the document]
dong77: Great, please add it.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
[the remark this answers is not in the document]
dong77: I wonder if someone can use an global agent as `to` to attach us or a user.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
dong77: But exchange.deposits calls a token's `transfer` method which may call back into our contract.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: I am ok with not spling the code into multiple ones.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: The above comment of yours answers some of my questions.
I wonder: if we process deposit and its L2 transfers in a single block **atomicly**, may we be able to completely remove `Bridge.forceWithdraw` to simplify the code? Or maybe from the DEX's perspecive, there is no way to guarantee the deposit is processed with the transfers all together as the deposit is a normal deposit and the DEX/operator can process it without even considering the rules in a Bridge. If the answer is NO, then we have to keep the current design and allow deposit/transfers to be processed in different blocks.

packages/loopring_v3/contracts/aux/bridge/IBridgeConnector.sol
```diff
@@ -0,0 +1,19 @@
...
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "./BridgeData.sol";
+
+
+interface IBridgeConnector
+{
+    function processCalls(ConnectorCalls calldata connectorCalls)
```
dong77, line 11: I suggest we do not pass-in `ConnectorCalls`, but `ConnectorGroup[]`. Many fields of `ConnectorCalls` should be transparent to `IBridgeConnector`

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+                tokens[i].amount = tokens[i].amount.sub(connectorCall.tokens[i].amount);
+            }
+
+            // Call the connector
+            _connectorCall(connectorCall);
+
+            // Verify the transactions
+            for (uint g = 0; g < connectorCall.groups.length; g++) {
+                ConnectorGroup memory group = connectorCall.groups[g];
+                for (uint i = 0; i < group.calls.length; i++) {
```
dong77, line 308: I'm having a hard time understanding the notion of groups.  If we use adding liquidity of two tokens into Uniswap as an example, in such a group, there are two transfers and one Join. So maybe the group should have **multiple transfers** with **one call**. But the current implementation seems to assume each call has one matched transfer.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
Brechtpd: Yep, first bring back funds to the Bridge contract (if nobody else has done so already, any funds already in the Bridge contract can be used) and then withdraw from there.

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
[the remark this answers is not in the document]
Brechtpd: If a global agent allows doing stuff for users it will have it's own authorization method. If the authorization method is simply `msg.sender = exchange.owner` there could be a problem (like for the ITransactionReceivers is the case), but otherwise I don't see why the owner would have any special powers in other contracts (except the exchange itself of course).

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
Brechtpd: That could indeed be true, but I guess only really a problem if we make token registrations public.

But even if a token does that, what can happen? `exchange.deposit` is `nonReentrant`, so doing additional deposits is not possible. But even if that would not be `nonReentrant` only additional deposits can happen which I think can never be a problem that can be exploited.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: You can do it atomicly, but that brings other problems:
- `batchDeposit` only does a transfer to the bridge. And for processing the deposits we also have to pay again to transfer funds to the deposit contract. So less efficient this way and more costs we have to pay (and recover fees for).
- Instead of moving tokens to the Deposit contract we could just keep the funds in the Bridge contract, and do "fake" deposit/withdrawals which just increase/decrease balances. Essentially this makes the Bridge contract a 2nd deposit contract where exchange funds could be stored. But that brings extra risk and I think overall more complexity.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+        BridgeOperations memory operations = abi.decode(callbackData, (BridgeOperations));
+
+        Context memory ctx = Context({
+            domainSeparator: DOMAIN_SEPARATOR,
+            accountID: accountID,
+            txIdx: 0
+        });
+
+        _processTransfers(ctx, operations.transferBatches, txsData);
+        _processCalls(ctx, operations, txsData);
```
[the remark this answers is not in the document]
Brechtpd: Will depend a lot on the optimizations. Having 1 inputs leaves more stack slots for local variables which could help to reduce gas costs.

Still working on the optimizations, but it's almost inevitable that the code will look substantially uglier and harder to understand, but I think that's unfortunately very much worth it now.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,618 @@
...
+        SignatureVerificationTransaction.SignatureVerification memory verification;
+        HashData memory hashData;
+        for (uint c = 0; c < operations.connectorCalls.length; c++) {
+
+            ConnectorCalls memory connectorCall = operations.connectorCalls[c];
+
+            // Verify token data
+            require(connectorCall.tokens.length == tokens.length, "INVALID_TOKEN_DATA");
+            for (uint i = 0; i < tokens.length; i++) {
+                require(tokens[i].token == connectorCall.tokens[i].token, "INVALID_CONNECTOR_TOKEN_DATA");
```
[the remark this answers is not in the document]
Brechtpd: Yeah, maybe doing a single array will be better. Will have to experiment a bit and measure what ends up being cheaper.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: A group is per specific operation. So for a tokenA -> tokenB swap, a group would simply be a list of tokenA transfers with. 
For batch migration a group could also just be a list of all kinds of tokens.

For adding liquidity for tokenA/tokenB the group would just contain 2 bridge calls from the same owner doing the 2 transfers.  But you're right, maybe it does indeed make sense to more tightly encode these together for cases like these with multiple token transfers (similar to how we do things for AMM joints/exits). It could potentially make cases with multiple transfers a little bit cheaper, but I think only worth if it doesn't make the more normal cases more expensive.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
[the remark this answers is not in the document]
dong77: As long as `exchange.deposit` is nonReentrant, it should be fine.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: I don't like the *second deposit contract* idea, so lets keep the design as-is.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: Ok, lets talk about this later after your optimizations.

packages/loopring_v3/contracts/aux/bridge/IBridgeConnector.sol
dong77: I suggest we move the definitation of `ConnectorCalls` or whatever parameters taken by `processCalls` here into this contract to make it eaiser for people to understand the parameter struct without visiting other files. If `IBridgeConnector` is easy to understand, it's going to be less error-prone for implementation.

`ConnectorCalls` is currently hard to understand, and people will have to go to the Bridge.sol file to understand it's fields.

dong77:
The bridge enabled two functionalities: 1) batch deposits which involves one deposit and multiple L2 distribution, and 2) batch withdrawal that invloves mutliple L2 collections and one withdrawal. I wonder if we can better abstract these two functionalities into their easy-to-understand names, something like `RollupTransaction` (Connector call) and `RolldownTransaction` (batch deposit)...

dong77:
Maybe we should bump version to 3.7

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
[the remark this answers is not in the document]
Brechtpd: The normal callbacks can now be run before or after the blocks submission using a flag.

I kept the `TransactionReceiver`s (AMM, Bridge) to after the block submissions only to keep things simple (a bit more difficult to do then for the normal callbacks).

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Kept the bridge call with 1 transfer/bridge call for now to keep things a bit simpler but also the gas cost increased quite a bit the way I implemented it with more arrays that needed to passed on-chain (which mostly contained a single element, but still has some noticeable overhead of ~20% gas cost increase because we're only talking about ~13,000 gas per bridge call, so even ~2,000 gas extra is quite a lot). Most common scenario's will very likely do just a single transfer so I think it should be fine, and multiple bridge calls to do multiple transfers can still easily be submitted together.

Brechtpd:
> Maybe we should bump version to 3.7

Not that many changes to the core protocol except the flash minting (pretty minor chance). But some version bump seems necessary, 3.7 seems logical.

PR #2240, opened 2021-03-01
[hebao] Smart wallet redesign

Brechtpd:
Big simplification of the current design. Compiles for L1 and L2 optimism.

Does not implement any L1 specific optimizations like some stateless wallet ideas because that makes the L2 solutions more expensive (biggest cost for L2s is transaction data). So if we want to optimize for both we will want to do some small platform specific changes, which should be pretty easy because all the actual logic can be reused. But in most cases that should only really impact deployment costs which can setup all wallet settings. In normal use there shouldn't be that much of difference because not that much wallet state is read/modified. So nothing fancy there yet.

Meta-transactions will also be much less useful not so long from now because of the account abstraction stuff that's coming (and will be native for L2s) so also tried to keep that as simple as possible.

Also a much easier "who can call what". Only the wallet factory has special permissions on the smart wallet for the `initialize` function, which is only necessary to support the blank contracts logic and this can only be done once, so afterwards no external addresses have any special permissions.

Current proxy supports full upgradeability so will be a bit more expensive to deploy than a fixed proxy, but we could support both. But I like always being able to support upgrading the implementation.

Code not tested and also not yet optimized.


dong77:
Great, Brecht, I may only have time to look into this next week. In the meanwhile, you can also dig into BAMM if you have time.

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -0,0 +1,156 @@
...
+    function executeMetaTx(
+        Wallet      storage wallet,
+        bytes32             DOMAIN_SEPARATOR,
+        PriceOracle         priceOracle,
+        MetaTx      memory  metaTx
+        )
+        public
+        returns (bool success)
+    {
+        require(!wallet.locked, "WALLET_LOCKED");
```
dong77, line 85: If we require the wallet must be unlocked, then we cannot use meta-tx to unlock a wallet...
```

File: packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol [hunk 2]
```diff
@@ -0,0 +1,156 @@
...
+            wallet.nonce[msg.sender] = metaTx.nonce;
+        }
+
+        validateMetaTx(
+            wallet,
+            DOMAIN_SEPARATOR,
+            metaTx
+        );
+
+        (success, ) = metaTx.to.call{gas : metaTx.gasLimit}(metaTx.data);
```
dong77, line 103: what if we are calling `executeMetaTx` inside another meta-tx. Should we disallow it?

dong77: Maybe we should check "msg.sender != address(this)" inside executeMetaTx.

packages/hebao_v3/contracts/base/SmartWallet.sol
```diff
@@ -0,0 +1,546 @@
...
+    {
+        require(!isImplementationContract, "DISALLOWED_ON_IMPLEMENTATION_CONTRACT");
+        _;
+    }
+
+    modifier onlyFromWalletOrOwnerWhenUnlocked()
+    {
+        // If the wallet's signature verfication passes, the wallet must be unlocked.
+        require(
+            msg.sender == address(this) ||
```
dong77, line 70: Please see my comments in `executeMetaTx` inside "MetatxLib". 
It seems even if `msg.sender == address(this)`,we should also check the wallet is not locked.

dong77 (changes requested):

packages/hebao_v3/contracts/base/libwallet/ERC20Lib.sol
```diff
@@ -0,0 +1,315 @@
...
+        returns (bytes memory returnData)
+    {
+        if (priceOracle != PriceOracle(0)) {
+            // Disallow general calls to token contracts (for tokens that have price data
+            // so the quota is actually used).
+            require(priceOracle.tokenValue(to, 1e18) == 0, "CALL_DISALLOWED");
+        }
+
+        bool success;
+        (success, returnData) = to.call{value: value}(txData);
```
dong77, line 310: maybe we should check `to != address(this)`?

packages/hebao_v3/contracts/base/libwallet/SignedRequest.sol
```diff
@@ -0,0 +1,52 @@
...
+
+/// @title SignedRequest
+/// @dev Utility library for better handling of signed wallet requests.
+///      This library must be deployed and linked to other modules.
+///
+/// @author Daniel Wang - <daniel@loopring.org>
+library SignedRequest {
+    using SignatureUtil for bytes32;
+
+    function verifyRequest(
```
dong77, line 20: verifyApproval

dong77 (changes requested):

dong77:
@Brechtpd please also remove **hebao_2** and rename **hebao_v3** to **hebao_v2**

packages/hebao_v3/contracts/base/SmartWallet.sol
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    modifier onlyFromWalletFactory()
+    {
+        require(msg.sender == walletFactory, "NOT_FROM_WALLET_FACTORY");
+        _;
+    }
+
+    constructor(
+        PriceOracle _priceOracle,
```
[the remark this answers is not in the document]
Brechtpd: I don't think that's possible because the price oracle is only used when a transaction is done by the wallet owner, so we can't trust the wallet owner with providing any price related data.

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -0,0 +1,156 @@
...
+
+        uint gasUsed = gasLeft - gasleft() + metaTx.gasOverhead;
```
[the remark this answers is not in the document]
Brechtpd: The idea would be to put that cost in `gasOverhead`:
- More accurate calculation (cost differs between zero and non-zero bytes)
- It's not really seen as good practice because gas costs keep on changing even on L1, let alone for L2.
- Cost for this will likely differ quite a bit depending on the chain. For an optimistic rollup this will be the large majority of the cost so an accurate calculation is critical, the actual execution cost can be pretty much ignored (currently I even think the fee for the tx always is 100% the gas limit set in the tx). On the other hand, for a zk rollup it may be free.

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
Brechtpd: Should be fine in theory, but yeah in practice I don't see any benefit so we may as well disallow it just to be safe.

packages/hebao_v3/contracts/base/SmartWallet.sol
Brechtpd: I think both approaches for checking for locked wallets is quite similar, but a small benefit of checking if the wallet is locked or not on the meta tx level is that using wallet funds to pay for transactions is disallowed. I think this is a benefit for wallet users compared to just checking it on the internal call level where the cal may fail quickly but transactions costs can still be reimbursed.

packages/hebao_v3/contracts/base/libwallet/ERC20Lib.sol
Brechtpd: Also a case where it should be fine in theory, but can't see any benefit in practice so will disallow.

packages/hebao_v3/contracts/base/libwallet/WalletData.sol
```diff
@@ -0,0 +1,69 @@
...
+
+// Optimized to fit into 32 bytes (1 slot)
+struct Guardian
+{
+    address addr;
+    uint8   status;
+    uint64  timestamp; // validSince if status = ADD; validUntil if adding = REMOVE;
+}
+
+struct Wallet
```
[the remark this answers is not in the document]
Brechtpd: Will change if you really want to, but because it's used everywhere I kind of like it being short. It also follows the style of other structs like `Guardian`, `Quota`, `Approval`,... which also don't end with `Data` either so I thought why do it for `Wallet`?

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
Brechtpd: Oh I guess it could be useful if the unlock is indeed done on the wallet itself, which also means the wallet owner can be trusted to sign the meta-tx itself (otherwise there would be no unlock). I think we can also allow that specific use-case by moving the `require(!wallet.locked, "WALLET_LOCKED");` check after the meta-tx call, perhaps at some extra risk to the relayer.

packages/hebao_v3/contracts/base/SmartWallet.sol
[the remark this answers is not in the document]
dong77: Good point.

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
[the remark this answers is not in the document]
dong77: I see. Since gasOverhead is provided and verified by us, it should be fine.

packages/hebao_v3/contracts/base/libwallet/WalletData.sol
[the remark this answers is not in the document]
dong77: I'm OK with the existing naming.

packages/hebao_v3/contracts/base/libwallet/MetaTxLib.sol
dong77: Like this idea.

dong77:
@Brecht, I don't have more feedback for now. Kongliang is reviewing the PR. Do you want to make the PR out of draft or you are still working on some improvement?

Brechtpd:
Not working on any improvements any more, but I'm sure testing will find some more areas for improvement.

dong77:
> Not working on any improvements any more, but I'm sure testing will find some more areas for improvement.

In that case, I'll approve this PR and @kongliangzhong is also looking at it. Tests and fixes can be in other PRs.

dong77 (approved):

packages/hebao_v2/contracts/base/WalletFactory.sol
```diff
@@ -0,0 +1,233 @@
...
+    event BlankConsumed (address blank);
+
+    bytes32             public immutable DOMAIN_SEPERATOR;
+    address             public immutable walletImplementation;
+
+    BaseENSManager      public immutable ensManager;
+    ENSResolver         public immutable ensResolver;
+    ENSReverseRegistrar public immutable ensReverseRegistrar;
+
+    mapping(address => bool) blanks;
```
dong77, line 34: make this public as well?

packages/hebao_v2/contracts/thirdparty/proxy/WalletProxy.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Taken from: https://github.com/gnosis/safe-contracts/blob/development/contracts/proxies/GnosisSafeProxy.sol
+pragma solidity ^0.7.0;
+
+/// @title IProxy - Helper interface to access masterCopy of the Proxy on-chain
+/// @author Richard Meissner - <richard@gnosis.io>
+interface IProxy {
```
dong77, line 7: This interface is not used anywhere...

packages/hebao_v2/contracts/thirdparty/proxy/WalletProxy.sol
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/// @title WalletProxy - Generic proxy contract allows to execute all transactions applying the code of a master contract.
+/// @author Stefan George - <stefan@gnosis.io>
+/// @author Richard Meissner - <richard@gnosis.io>
+contract WalletProxy {
+
+    // masterCopy always needs to be first declared variable, to ensure that it is at the same location in the contracts to which calls are delegated.
+    // To reduce deployment costs this variable is internal and needs to be retrieved via `getStorageAt`
+    address internal masterCopy;
```
dong77, line 18: Maybe we should make this masterCopy public so people can view and verify.

dong77: @Brechtpd It doesn't seem this "masterCopy" is used anywhere in the code in a meaningful way. Could you please explain the idea, especailly compare to 'walletImplementation`?

packages/hebao_v2/contracts/base/libwallet/Utils.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
```
dong77, line 3: Shall we upgrade to use solidity ^0.8.2?

packages/hebao_v2/contracts/base/libwallet/WhitelistLib.sol
```diff
@@ -0,0 +1,126 @@
...
+            approval,
+            abi.encode(
+                ADD_TO_WHITELIST_TYPEHASH,
+                approval.wallet,
+                approval.validUntil,
+                addr
+            )
+        );
+
+        wallet._addToWhitelist(
```
dong77, line 63: one line

packages/hebao_v2/contracts/price/KyberNetworkPriceOracle.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.7.0;
```
dong77, line 3: Lets remove Kyber.

packages/hebao_v2/contracts/thirdparty/proxy/WalletProxy.sol
dong77: Oh, I get it. This is behind the WalletProxy contract.

packages/hebao_v2/contracts/thirdparty/proxy/WalletProxy.sol
Brechtpd: It's a trick for the proxy. `address internal masterCopy;` is made internal, because making it public would auto generate getter methods that would increase the proxy contract size, so deployment costs increase (I tested this out, is makes a significant difference, I think it saves 2 storage slots). But `masterCopy` is still public because the function is implemented in the `fallback` methdod:
```
// 0xa619486e == keccak("masterCopy()"). The value is right padded to 32-bytes with 0s
if eq(calldataload(0), 0xa619486e00000000000000000000000000000000000000000000000000000000) {
   mstore(0, _masterCopy)
   return(0, 0x20)
}
```

So the proxy contract doesn't implement the interface from solidity's point of view, but it does actually implement it in a non-standard way so the interface can be used to interact with the proxy contract.

packages/hebao_v2/contracts/thirdparty/proxy/WalletProxy.sol
Brechtpd: Yes, it's to make the upgrade of the wallet controlled by the wallet itself in a clear and efficient way (unlike `OwnedUpgradabilityProxy` which needs additional data for `owner` for example, which is always the wallet itself in this case).

PR #2259, opened 2021-03-05
[Protocol 3] only allow owner to call `initialize` in DefaultDepositContract

dong77:
This PR is created based on the feedback from turbo3696@gmail.com:

The contract  DefaultDepositContract [https://github.com/Loopring/protocols/blob/Loopring_3.6.0/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol] can lose ownership due to improper access control.

Vulnerability:
The function named  initialize [https://github.com/Loopring/protocols/blob/c918b164d30f7b9a3d948225f09b635257b06844/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol#L51] can be used by attacker to claim ownership and set arbitrary exchange address.
Scenario:

1.Attacker sees the contract being deployed.
2.Calls initialize function immediately and claims the ownership.
    This function can also be front-run.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
+++ b/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
@@ -52,6 +52,7 @@ contract DefaultDepositContract is IDepositContract, Claimable
         address _exchange
         )
         external
+        onlyOwner
     {
         require(
             exchange == address(0) && _exchange != address(0),


packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
```diff
@@ -52,6 +52,7 @@ contract DefaultDepositContract is IDepositContract, Claimable
         address _exchange
         )
         external
+        onlyOwner
```
Brechtpd, line 55: The contract is used behind a proxy, so when calling `initialize` there is no owner yet.

Brechtpd:
I don't really agree with the issue because normally you just create a proxy for an implantation and directly call the initialize function in the same tx. But even if that's not the case, the expected `initialize` call done by the proxy contract deployer will fail, which is an obvious sign the contract cannot be used.

PR #2275, opened 2021-03-15
[hebao 2.0] remove ens and chiToken

kongliangzhong:


packages/hebao_v2/contracts/base/WalletFactory.sol
```diff
@@ -176,36 +123,14 @@ contract WalletFactory
             config.feeRecipient,
             config.feeToken,
             config.feeAmount,
-            keccak256(bytes(config.ensLabel)),
-            config.ensRegisterReverse,
             salt,
-            blankAddress
+            address(0)
```
Brechtpd, line 127: Can be removed if blanks are removed as well?

Brechtpd:
If blanks are removed then all `walletFactory` references in `SmartWallet` can be removed, there is no need any more for a `onlyFromWalletFactory` on `initialize` because deployment and initialization are done immediately after each other. So `address     public immutable walletFactory;` and it's uses can be removed to make the wallet completely independent of the wallet factory.

Brechtpd (approved):

PR #2336, opened 2021-04-14
[hebaov2]add loopring label in all event emitted by wallet

kongliangzhong:


Brechtpd:
Could you provide a bit more detail what this allows us to do?

dong77:
I think I know a bit more about the reason behind this PR. Before it, we will have to listen to all events or all wallet addresses for events; with this one, we can filter based on a field with this constant value, then most events are our wallet events. It’s to make event subscription easier.

dong77:
@kongliangzhong The Java library shall allow us to subscribe to multiple events. If that's the case, then we don't need this change at all.

kongliangzhong:
After some discussion, we decide to use topic0 of each event as the filter. So this PR is closed.

Issue #2221, opened 2021-02-21
[Protocol 3] Discussion on https://github.com/Loopring/protocols/wiki/L2-Vaults


dong77:
Brecht, lets have some discussion here regarding your L2-Vaults.

Regarding idea number 2, do you assume the join/exit prices are **controlled** by us during **one Loopring block time**. If so, the deficit/surplus idea make sense to me. In my option, I think we have two approaches to handle defict/surplus. The first approach is batching it on L1 with flash loans; the second one is automatically cancel all pending joins/exits requests on L2 (after 1 or 2 Loopring blocktime) to zero out the defict/furplus. This is helpful when we believe batched L1 join/exit with flash loan will fail.

By "controlled", I mean the relayer will use the realtime L1 price as an oracle, then adjust join and exit prices accordingly. This will have to be done case-by-case. Some pool's token/vtoken price depends on the join/exit amount, our relayer will have a mirror implementation to simulate the price calculation, the more precise the better.

Regarding flash loan, this is my understanding:
- if we need 100 vETH on L2, we can take a flash loan of 100ETH on L1, join the pool to get 100vETH, deposit the vETH before submitting blocks; then we have a withdrawal of 100 ETH directly to the flash loan to pay back the loan.

- if we need 100 ETH on L2, we take a flash loan of 100ETH on L2, deposit to L2; then we withdraw 100vETH and exist on L1 and pay the flash loan back after block submission.

I believe a loan of Token is much easier of a loan of vToken.

Btw, I don't really get the term "Flash deposits". My understading is still a layer-1 flash loan used by deposits, but "Flash deposits are basically flash loans on L2" confuses me. Could you please clarify.

dong77:
Now I have read your PR and I know what you mean by "flash deposits" - mint some token out of the thin air then deposit to L2, then on L1 pay it back after submit blocks.

Brechtpd:
> Regarding idea number 2, do you assume the join/exit prices are controlled by us during one Loopring block time. If so, the deficit/surplus idea make sense to me.

For 2) we check the actual on-chain price each time somebody does join/exit to the L2 vault. If we can directly match the deposit/withdrawal with a deposit/withdrawal in the queue we just do it immediately at the latest on-chain price. If that's not possible and we support flash deposits, we can then indeed batch settle all the remaining ones at some fixed price.

[...]

PR #2344, opened 2021-04-16
[protocol 3.7] Loopring Ethport (Bridge and Converter) and AMM Optimizations (#2218)

dong77:
https://github.com/Loopring/protocols/wiki/L2-Vaults

Initial idea for using flash loans here: https://github.com/Loopring/protocols/issues/1196#issuecomment-633324180

## Bridge

The Bridge works similarly to an AMM pool in that the contract is the owner of an account on layer 2, and on-chain we verify and approve transactions from and to this account. This is the bridge that is used to communicate from and to Loopring. Most  times the Bridge account and contract contains no funds, it's simply used as an intermediate. To do communication with Loopring the bridge allows doing just two things.

### 1. Batch deposits

There is a `batchDeposit` call that can be done by any external contract that simply deposits a number of tokens to the Bridge account on L2. The functions also takes in a list of transfers that need to be done on layer 2 from the bridge account to all other accounts in this specified list. So this way only one normal deposit for each distinct token is done, and all other batched deposits are done by simple transfers from the bridge account to the actual destination address on L2. The specified transfers are hashed and stored on-chain so we can later process them in a block, an event is logged with the data to make fetching this list of transfers easy. A unique `batchID` is assigned to each batched deposit to make sure the hash is unique and can't fail. The operator doesn't have to do all these batch deposits, if they are too old they can be withdrawn from the bridge contract directly as a safety for both the operator (spam) and users (operator stops working). The bridge also allows a small max fee to be used from the batched deposits to pay for the L2 transfers.

### 2. Bridge calls

The bridge also allows calling external contracts efficiently (with funds) using batching with the help of `Connector`s. The idea of the `Connector` is that an integration with some dApp is wanted, but we need to be able to do these efficiently in a batched way for all kinds of different scenarios like:
- Mass migration
- L1 token trading
- L1 liquidity provision
- L1 vault interaction
- L1 lending
- ...

All these can work differently for batching so we leave that up to the connector to implement, and so the bridge just calls the standard function `processCalls` to let the connector do what's necessary. In some cases these connectors just need to do another L1 call and that's it. In other cases, for example a token swap, another token is bought that needs to be returned to the buyer on L2. This is where the batched deposits come in again to allow the connector to do just that in a standard way.

These connectors also need to handle the funds in some way. In all cases all funds are first moved to the Bridge contract. This was chosen as a way to not impose any risks on all funds in the deposit contract, I think it's too risky to even call trusted functions directly on the deposit contract. But the current method is pretty efficient, in many cases there is only a need to do as many L1 token transfers as there are unique tokens to and from the Bridge contract, even if multiple connectors are called.

All connectors can do everything, but there is the concept of trusted and untrusted connectors for efficiency reasons, which gas a small implementation difference:
- The Bridge calls `processCalls` on trusted connectors using a delegate call. This means no additional token transfers are needed to execute the code in the connector, but the connector has access to all funds currently being processed in the bridge.
- The Bridge calls `processCalls` on non-trusted connectors using a real call after transferring the necessary funds to the connector contract. This makes sure that only funds of users that want to interact with that connector are at risk in that connector.

What is a trusted and untrusted connector can be set by the Bridge owner. Implementation wise this is only a very small difference at the Bridge side (the connector doesn't need to know), but for gas efficiency it can be worth it to make the connector trusted after we make sure it's safe. On the other hand, untrusted connectors are also great because users can call any connector without us having to worry about the connector misbehaving and putting other people's funds at risk.

No matter if the connector is trusted or not, connectors are given a max gas limit and are always allowed to fail in the  `processCalls` call. If the call fails the Bridge automatically approves the transfers back from the Bridge to the user's account.

There is also the notion of a connector _group_. This is simply a way that the same connector can be used with all kinds of inputs. For example, for a uniswap connector a group would be the two tokens that need to be swapped, and this data is encoded in `groupData` and grouped together so all these swaps can be handled together.

`userData` on the other hand is for user specific data a connector could us. Think something like max slippage for a swap, or the destination address for a mass migration to a different rollup.

Calls submitted to the connector will be sorted on group and tokens so they can be easily processed.

Similarly to the AMM pools, the user can sign a `BridgeCall` transaction which is verified on-chain using the `SignatureVerification` tx so the L2 keys can be used. So every bridge call requires 2 L2 transactions: an L2 transfer and the signature verification tx. 

#### Fees

When people sign a transaction they don't know how big a batch it will be part of. The operator itself also doesn't always know how much a bridge call will end up costing.

To solve this problem users sign a transaction and a pay fee for an abosolute amount of gas that the operator needs to provide to the batched bridge calls. The total amount of gas the operator needs to provide to the connector is at least the sum of all these gas amounts.

The other problem is that, when just using this mechanism, the operator could submit a single bridge call that is very likely to have low gas amount just by itself, which will fail because the whole operation will take much more gas than what that single bridge call provides. And so there is an additional `IBridgeConnector.getMinGasLimit` which also forces the operator to at least provide the minimum amount of gas required for all the calls by the connector itself.

## Converters

`Converter`s are the implementation of the scheme described [here](https://github.com/Loopring/protocols/issues/2221#issuecomment-783597146). They are very efficient at doing many conversions from tokenA -> tokenB, but each converter needs its own token and contract deployed.

`BaseConverter` handles all the general code that can be reused between all converters (failure logic for the conversion, but also the main vault stuff like the token code, burning, minting, depositing, withdrawing,...). Specific implementations just need to implement a simple `convert` function to do the specific implementation for the conversion. There is no problem with rounding errors for the vault token transfers if the user doesn't pay a fee on this token (which I think wouldn't make a lot of sense) because the vault token is transferred in and out using a float representation.

Everything Converters can do, the Bridge can do as well, but a bit less efficient in some cases (although the Bridge could also be more efficient in others). On the other, the Bridge can do everything converters can do and more (like mass migrations between L2s).

### Converter

The converter can do anything that can be represented by a token <-> token trade, under parameters shared by everyone in the batch. Each conversion from tokenA -> tokenB needs a separate contract.
- Swap tokens on L1 without minimum price or using a minimum price encoded in the converter contract
- Add liquidity to an L1 pool, even with multiple tokens, in return for a token representing the user's share. Yearn, AAVE, ETH staking, ....
- ....

### Bridge

The bridge can do anything the Converter can do and more. For example, the bridge can do anything that can be represented by a token <-> token trade, but now with parameters that can be defined by each user separately. The bridge also supports operations that only require tokens to go out (like mass migration). Only a single `BridgeConnector` contract needs to be deployed to enable a certain kind of L1 operation for all tokens. For example, just a single Uniswap connector is able to convert between any tokens.

Even L1 operations that are not batchable could have some small benefits by using the bridge because the token withdrawals/deposits from/to the exchange can be shared by all bridge operations.

- Swap/Add liquidity on L1 with per-user settings (minimum price, minimum pool share,...)
- Calling any L1 function that requires knowing the user's Ethereum address (mass migration,...)
- ....

### Flash minting

- Allow arbitrage on L2 without needing any funds
- Allows batching AMM joins/exits
- ...


packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
```diff
@@ -343,6 +343,53 @@ abstract contract IExchangeV3 is Claimable
...
+        )
+        external
+        virtual;
+
+    /// @dev Repays funds minted using `flashMint`.
+    /// @param from The address that deposits the funds to the exchange
+    /// @param tokenAddress The address of the token, use `0x0` for Ether.
+    /// @param amount The amount of tokens to deposit
+    /// @param extraData Optional extra data used by the deposit contract
+    function repayFlashMint(
```
dong77, line 364: shall we also allow users to "withdrawlToPayFlashMint", this may enable another set of use cases.

packages/loopring_v3/contracts/converters/BaseConverter.sol
```diff
@@ -0,0 +1,202 @@
...
+            balance = address(this).balance;
+        } else {
+            balance = ERC20(token).balanceOf(address(this));
+        }
+
+        // Share to withdraw
+        uint amount = balance.mul(poolAmount) / totalSupply;
+
+        // Burn pool tokens
+        _burn(msg.sender, poolAmount);
```
dong77, line 120: This pool token can only be owned by the DepositContract, so only the DepositContract can call this function?  I'm having a hard time understand who are supposed to be the msg.sender.

packages/loopring_v3/contracts/converters/BaseConverter.sol
```diff
@@ -0,0 +1,202 @@
...
+        } catch {
+            failed = true;
+            emit ConversionFailed("unknown");
+        }
+
+        // Mint pool tokens representing each user's share in the pool, with 1:1 ratio
+        _mint(address(this), amountIn);
+
+        // Repay the flash mint used to give user's their share on L2
+        _repay(address(this), amountIn);
```
dong77, line 95: The new-minted pool tokens will be eventually transferred to the exchange's deposit contract, correct?

packages/loopring_v3/contracts/converters/BaseConverter.sol
```diff
@@ -0,0 +1,202 @@
...
+
+        // Share to withdraw
+        uint amount = balance.mul(poolAmount) / totalSupply;
+
+        // Burn pool tokens
+        _burn(msg.sender, poolAmount);
+
+        // Use to repay flash mint directly if requested
+        if (repayAmount > 0) {
+            _repay(token, repayAmount);
```
dong77, line 124: I see the idea is: for the second phase, we still use flash deposit, then use this `withdraw` method to pay either Token or yToken to trade with users' vToken. 

But since we already have the real token (Token or yToken) inside the Converter contract, an alternative is that the Converter does a real deposit of either Token or yToken before the next block is submitted using the same Ethereum tx. Therefore, the second phase doesn't need flash mint/deposit. 

I think that will be more efficient.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    function _processPoolTx(
+        AmmData.State   storage  S,
+        AmmData.Context memory   ctx,
+        bytes           calldata callbackData
+        )
+        private
+    {
+        // Manually decode the encoded PoolTx in `callbackData`
```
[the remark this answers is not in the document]
Brechtpd: Yes, no other reason. This saves on an abi decode that copies the data from calldata to memory. And that would also make the 2nd abi decode that is still used for decoding the PoolJoin/PoolExit more expensive because it would then need to be done from memory.

packages/loopring_v3/contracts/test/SinglePhaseConverter.sol
```diff
@@ -0,0 +1,54 @@
...
+        );
+    }
+
+    function canDrain(address drainer, address /* token */)
+        public
+        override
+        view
+        returns (bool)
+    {
+        return drainer == owner;
```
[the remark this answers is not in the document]
Brechtpd: The single phase converter does not use pool tokens and the contract normally does not contain any funds. In a single phase converter all actual trades already happened on L2, the converter contract just helps do the conversion and repay the flash loan. So when funds are stored here they are fully owned by whoever uses this contract, the funds are not owned by users, they already got what they wanted on L2.

packages/loopring_v3/contracts/test/SinglePhaseConverter.sol
```diff
@@ -0,0 +1,54 @@
...
+pragma solidity ^0.7.0;
+
+import "../core/iface/IExchangeV3.sol";
+import "../lib/Claimable.sol";
+import "../lib/Drainable.sol";
+import "../lib/ERC20.sol";
+
+
+/// @author Brecht Devos - <brecht@loopring.org>
+contract SinglePhaseConverter is Claimable, Drainable
```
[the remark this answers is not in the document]
Brechtpd: Single phase converter does not need any vault logic, just a simple conversion + repay. See my other comment above and/or my slides for scaling Ethereum.

packages/loopring_v3/contracts/core/iface/IExchangeV3.sol
Brechtpd: I don't think it enables any new use case, but it could be a bit more efficient in case the flash mints are used for L2 only operations (otherwise some token conversion needs to happen on L1 and a different token is withdrawn than is used for repayment). If we would offer users access to flash minting I would assume all of them would simply pass through an account owned by us because that way we can batch them together, and then we can also just use the existing methods to repay it.

I would say we can add it we ever would do that if we want to save gas in that specific case (presumably, don't know if it will save much gas), but not add features on critical paths until we plan on using them.

packages/loopring_v3/contracts/converters/BaseConverter.sol
Brechtpd: I should have explained this somewhere a bit better, the test in `testConverter.ts` makes it more clear what happens but I'll describe it here.

Phase 1:
- First flash mint the necessary vault tokens
- Do the phase 1 trades to the vault token
- Withdraw tokens collected to the convert address (!). This is because the L1 swap will be done in the converter, so it's cheaper to just transfer the tokens there directly. There's also no risk someone else being able to use those funds because the withdrawal transfer + converter deposit happens in the same transaction.
- Than we temporarily mint the vault tokens to the convert address because that's also the address that will repay the flash loan
- Use those minted tokens to repay the flash loan

Phase 2:
- First flash mint the necessary tokens people want to actually buy
- Do the phase 2 trades to the token the user wanted to buy
- Withdraw tokens collected to the operator contract address (!). The operator contract will do a callback to will batch withdraw from the vault by burning all vault tokens together.
- The tokens unlocked by burning the vault tokens are now used to repay the flash loan

packages/loopring_v3/contracts/converters/BaseConverter.sol
Brechtpd: Not sure if I understand the question correctly, but yes that's what the repay function will do. They are already on L2, the repay function makes sure the tokens actually exist on L1.

packages/loopring_v3/contracts/converters/BaseConverter.sol
Brechtpd: > I see the idea is: for the second phase, we still use flash deposit, then use this withdraw method to pay either Token or yToken to trade with users' vToken.

Yep.

> But since we already have the real token (Token or yToken) inside the Converter contract, an alternative is that the Converter does a real deposit of either Token or yToken before the next block is submitted using the same Ethereum tx. Therefore, the second phase doesn't need flash mint/deposit.

The tokens are indeed already n the converter contract, but we cannot access them! The tokens in the vault are fully owned by the owners of the vault tokens, and we gave those to the users on L2 in phase 1 in return for the tokens they want to sell. And it's important we give the vault tokens to the users, otherwise they would stop being the owner of their funds in the converter and someone else/we could just take them. Only in return for the vault tokens are the tokens in the converter unlocked, so we first need to to collect them on L2.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
```diff
@@ -80,6 +80,28 @@ library AmmExitRequest
...
-        );
+        );*/
+        bytes32 typeHash = POOLEXIT_TYPEHASH;
+        address owner = exit.owner;
+        uint burnAmount = exit.burnAmount;
+        uint burnStorageID = exit.burnStorageID;
+        uint96[] memory exitMinAmounts = exit.exitMinAmounts;
+        uint fee = exit.fee;
+        uint validUntil = exit.validUntil;
+        assembly {
```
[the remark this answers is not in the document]
Brechtpd: I just read the solidity docs, print out the byte encoding of things and then try everything I can think of that could reduce the amount of gas used and then measure if I was right or wrong. No big brain stuff, just trying things out and see what sticks.

packages/loopring_v3/contracts/converters/BaseConverter.sol
```diff
@@ -0,0 +1,202 @@
...
+        )
+    {
+        exchange = _exchange;
+        depositContract = _exchange.getDepositContract();
+    }
+
+    function initialize(
+        string memory _name,
+        string memory _symbol,
+        uint8         _decimals,
```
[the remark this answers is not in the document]
Brechtpd: The reason I made it configurable is that the vault token is always minted 1:1 with the `tokenIn` token, so it would make sense if they have the same amount of decimals. But this of course has no impact on anything directly, except perhaps how we decide how the token is shown in the UI. If that always just looks at the contract it makes sense to set the "correct" value, otherwise a fixed value can indeed be used without any effect.

packages/loopring_v3/contracts/amm/libamm/AmmExitRequest.sol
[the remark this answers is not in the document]
dong77: Maybe you can public an article "How I pushed Loopring performance to the extreme", something that people can also learn.

packages/loopring_v3/contracts/converters/BaseConverter.sol
dong77: I think I also figured it out last night. I'll explain the whole process to the relayer team for them to implement. I will forward questions that I cannot answer :)

packages/loopring_v3/contracts/converters/BaseConverter.sol
dong77: I see the point.

dong77 (commented):
Brecht, when you have new changes, would you create a PR on top of this branch?

Brechtpd:
> Brecht, when you have new changes, would you create a PR on top of this branch?

#2347 is on top of this branch.

packages/loopring_v3/contracts/test/TestMigrationBridgeConnector.sol
```diff
@@ -0,0 +1,123 @@
...
+        IBridge     _bridge
+        )
+    {
+        exchange = _exchange;
+        depositContract = _exchange.getDepositContract();
+
+        bridge = _bridge;
+    }
+
+    function processCalls(ConnectorGroup[] memory groups)
```
dong77, line 51: A lot of code in this method are the same as in TestSwappperBridgeConnector's `processCalls`. Shall we extract the same logic into Bridge.sol?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+        Context          memory   ctx,
+        ConnectorCalls   calldata connectorCalls,
+        uint                      n,
+        ConnectorCalls[] calldata allConnectorCalls
+        )
+        internal
+        returns (BridgeTransfer[] memory transfers)
+    {
+        require(connectorCalls.connector != address(this), "INVALID_CONNECTOR");
+        require(trustedConnectors[connectorCalls.connector], "ONLY_TRUSTED_CONNECTORS_SUPPORTED");
```
dong77, line 654: So untrusted connectors are not yet supported? I remember I saw code that support untrusted ones, did you remove it?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+            packedData & 0xffffffffffffffffffffffffffffffffffffffffffffffffff ==
+            (uint(ExchangeData.TransactionType.SIGNATURE_VERIFICATION) << 192) | ((uint(owner) & 0x00ffffffffffffffffffffffffffffffffffffffff) << 32) | _accountID &&
+            data == uint(txHash) >> 3,
+            "INVALID_OFFCHAIN_L2_APPROVAL"
+        );
+
+        ctx.txsDataPtr += ExchangeData.TX_DATA_AVAILABILITY_SIZE;
+    }
+
+    function encode(BridgeOperations calldata operations)
```
dong77, line 898: do we need this?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+        exchange = _exchange;
+        accountID = _accountID;
+
+        depositContract = _exchange.getDepositContract();
+        exchangeOwner = _exchange.owner();
+
+        DOMAIN_SEPARATOR = EIP712.hash(EIP712.Domain("Bridge", "1.0", address(this)));
+    }
+
+    function batchDeposit(
```
dong77, line 141: shall we add `onlyFromExchangeOwner` just as we did for BaseConverter?

dong77, line 141: shall we add `onlyFromExchangeOwner` just as we did for BaseConverter?

packages/loopring_v3/contracts/test/TestMigrationBridgeConnector.sol
Brechtpd: Both are just test implementations and actual implementations may need to do very different things depending on the dapp that needs to be interacted with. So I don't think we can put more into Bridge without losing flexibility in the connectors to do what is needed. If some functionality can be reused an extra external library seems better.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Yes, I removed it for now because trusted connectors allow for more optimizations:
- Tokens don't have to be transferred to the connector to keep things secure
- All returned bridge transfers from trusted connectors can be grouped together in a single batch deposit. For untrusted connectors this is not possible because they can return garbage data that could make all connectors or even the complete Ethereum transaction fail.

I'm assuming we will (definitely in the short and medium term) will only use trusted connector even just for UI reasons because we want to know what actually happens. And of course also because it's cheaper. So didn't think the extra complexity was worth it for now.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: No, will remove once everything is final! I currently use this function to help encode the data that needs to be sent on-chain encodede in `bytes`. It's easier to do it like this if the function signature changes a lot in the tests, otherwise I had to update the encoding each time manually.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: No, we want this to be callable by everyone! This is the function other rollups/CEXs can use to mass migrate users to Loopring. The relayer doesn't need to know who called it because it's the same format as the batch deposits done by the connectors.

Most of the gas cost is paid by the caller of this function because this already does the actual deposit to the exchange. The only thing left to do after this function is the distribution on L2 which costs ~3,000 gas. To help pay for this we can take `MAX_FEE_BIPS` of the amount deposited. But if we get spammed or this is not enough to cover the costs the relayer can also just ignore the batch deposit request.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+            deposits[c] = _connectorCall(
+                ctx,
+                connectorCall,
+                c,
+                connectorCalls
+            );
+        }
+
+        // Verify withdrawals
+        _processWithdrawals(ctx, totalAmounts);
```
dong77, line 478: @Brechtpd shall we process withdrawals before calling connectors?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+        )
+        internal
+        returns (BridgeTransfer[] memory transfers)
+    {
+        require(connectorCalls.connector != address(this), "INVALID_CONNECTOR");
+        require(trustedConnectors[connectorCalls.connector], "ONLY_TRUSTED_CONNECTORS_SUPPORTED");
+
+        // Check if the minimum amount of gas required is achieved
+        bytes memory txData = _getConnectorCallData(ctx, IBridgeConnector.getMinGasLimit.selector, allConnectorCalls, n);
+        (bool success, bytes memory returnData) = connectorCalls.connector.fastCall(GAS_LIMIT_CHECK_GAS_LIMIT, 0, txData);
```
dong77, line 658: What is this `fastCall` for? The `fastDelegatecall` is the real interaction right?

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -0,0 +1,902 @@
...
+        if (success) {
+            require(connectorCalls.gasLimit >= abi.decode(returnData, (uint)), "GAS_LIMIT_TOO_LOW");
+        } else {
+            // If the call failed for some reason just continue.
+        }
+
+        // Execute the logic using a delegate so no extra transfers are needed
+        txData = _getConnectorCallData(ctx,IBridgeConnector.processCalls.selector, allConnectorCalls, n);
+        (success, returnData) = connectorCalls.connector.fastDelegatecall(connectorCalls.gasLimit, txData);
```
dong77, line 667: shall we have something like this:
```if (trusted) connector.fastDelegateCall() else connector.fastCall()```

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: > Brecht, do we really want this Bridge to provide the batchDeposit method? It seems the internal _batchDeposit is really needed to do the L1->L2 token distribution, but allowing other people to call the external batchDeposit seems not an expected use case.

I would say it's certainly very much an expected use case, I would say it's even one of the 2 main features of the Bridge ( 1) general L1 -> L2 and 2) general L2 -> L1, and that's basically all the bridge does). The fact that the L1 -> L2 token distribution is slightly differen because an internal batch deposit function is only for optimization for trusted connectors, otherwise connectors would also just use the public batchDeposit function (untrusted connectors have to do batch deposits like that).

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Putting this here for other people looking at this PR: The bridge transactions are processed *after* the block has already been processed, which means all withdrawals on L1 already happened. So the funds are already in the Bridge contract before any of these transactions are even verified. So the order for L1 doesn't matter.
For the L2 transactions we do have to first process all transfers to the Bridge before we can withdraw from the Bridge, otherwise the Bridge accounts wouldn't contain any funds so the withdrawal would be invalid on L2.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Non-trusted connectors are a bit more different then that because for a normal call to work we would first have to transfer the necessary funds the connector needs to that connector contract. And also some other differences like not being able to trust any data returned from the contract so non-trusted connectors have to do a batchDeposit inside the connector call if needed.

Trusted connectors are easier/cheaper because we can just a delegate call and all code is executed as if it was from the Bridge contract. And it's the Bridge contract that contains all user funds necessary for the connector calls.

Brechtpd (approved):

Brechtpd:
All tests pass.

PR #2345, opened 2021-04-16
[Protocol 3.7] a few improvements and quite some renaming for clarity

dong77:
@Brechtpd Here are some of my suggestions for your PR. I still need to review Bridge/Connectors. If I made further changes, I'll create another PR. This PR is done and ready to be reviewed.

I didn't run tests, so please help me verify all tests still pass.


packages/loopring_v3/contracts/core/iface/ExchangeData.sol
```diff
@@ -158,7 +158,7 @@ library ExchangeData
         uint[24]                 balanceMerkleProof;
     }
 
-    struct FlashMint
+    struct LoanDeposit
```
Brechtpd, line 161: Any specific reason to call this a LoanDeposit instead of a FlashMint? With FlashMint/FlashLoan being the standard name for this. And if you would explain what a LoanDeposit is (which probably nobody would call it when talking about it) you would simply say it's an L2 flash mint/loan. So seems easier and less confusing to just call it that directly without creating new terms?

I also like FlashMint because it contains the 2 most important characteristics:
- Flash: only usable within the same transaction
- Mint: the tokens aren't loaned from somewhere, they are really just created (which is why I chose FlashMint and not FlashLoan).

LoanDeposit does not make that clear, and the fact that internally a deposit is used is just an implementation detail.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -43,6 +43,9 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
     using ExchangeWithdrawals   for ExchangeData.State;
 
     ExchangeData.State private state;
+    address public  __DEPRECATED__loopringAddr; // kept for proxy compatibility
+    uint8   private __DEPRECATED__ammFeeBips;   // kept for proxy compatibility
+    bool    public  __DEPRECATED__allowOnchainTransferFrom; // kept for proxy compatibility
```
Brechtpd, line 49: I added these values to the struct, and that way they should still map to the correct storage slot in the old contract:
https://docs.soliditylang.org/en/v0.8.3/internals/layout_in_storage.html

> The elements of structs and arrays are stored after each other, just as if they were given as individual values.

I did some external testing with this behavior and I think I did it right.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
dong77: I thought you will ask this, sorry that I didn't explain it before.

Based on my understanding, the minted pool token will live more than one Ethreum transaction across multiple L2 blocks. It will really exist on L1 and L2, especially for the two-phase converters. Therefore, flash is misleading (at least it did mislead me). 

My understanding the following may be called a flash mint:

1. mint a token on L1
1. deposit it to L2
1. use it somehow on L2 but get it back to the same L2 account in the same Ethereum transaction
1. withdraw it to L1
1. burn it all on L1

All these actions need to be in one Ethreum block, so the new token can be known as "flash minted". Whereas our current design is more like:
**[loan ->deposit -> mint -> pay-loan]**

`amountFlashMinted` is past tense, but the tokens are yet to be minted. And I really think `amountFlashMinted` represents a loan. 

The repayFlashMint/repayDepositLoan also doesn't require the token need to be a newly minted token, it can be LRC or Ether, or any existing token. So *Mint* is misleading and *Flash* seems not accurate.

Let me know what you think.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
dong77: That's true if we never add variables to this contract. But if we do in the future, the new variable will conflict with these three deprecated variables we delete.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
Brechtpd: Oh okay I see what you're trying to say. But let's make sure we keep the two kinds of minting separate because like you said, they don't have anything to do with each other because flash mints work for all tokens.

a) The flash mint on L2 (exchange feature, nothing to do with converters)
b) The pool token minting in converters (external vault feature, nothing to do with the exchange)

> the minted pool token will live more than one Ethereum transaction across multiple L2 blocks. It will really exist on L1 and L2, especially for the two-phase converters. Therefore, flash is misleading (at least it did mislead me).

- The minted tokens for a) (the exchange level) only live for a single Ethereum tx. The flash minting feature only works on this level.
- The minted tokens for b) (external vault level) can exist for as long as needed. These don't directly have anything to do with the flash minting going on on the exchange.

Both should not be mixed together, they are completely separate mechanisms.

>My understanding the following may be called a flash mint:
>    1. mint a token on L1
>    2. deposit it to L2
>    3. use it somehow on L2 but get it back to the same L2 account in the same Ethereum transaction
>    4. withdraw it to L1
>    5. burn it all on L1

The only difference for our implementation is 5), because instead of burning the minted tokens, we actually want to keep them around and so we transfer them in to make them "real". So in that respect we have created some kind of hybrid between a flash mint and a flash loan. The tokens are actually minted like in a flash mint (they may not exist yet), but it's repaid like a flash loan.

> All these actions need to be in one Ethereum block, so the new token can be known as "flash minted". 

This is true in our implementation as well as the feature only impacts a).

> amountFlashMinted is past tense, but the tokens are yet to be minted.

That's not true. `amountFlashMinted` is only from the exchange's point of view, which is a) where the flash minting is happening. The minting has happened on the exchange and `amountFlashMinted` needs to be paid back in the same Ethereum tx. The exchange does not care where the tokens come from (minted in some converter or not).

Now the confusing thing is that for the converter there is also another kind of minting happening which is b). So indeed for b) the minting still needs to happen, but that has nothing to do with a). But it is a bit strange to understand because of how we use the converter: in a) we mint tokens where the total supply on the converter's side is 0!

> The repayFlashMint/repayDepositLoan also doesn't require the token need to be a newly minted token, it can be LRC or Ether, or any existing token. So Mint is misleading and Flash seems not accurate.

But they are minted because they don't need to exist yet (all deposits are actually mints from L2s point of view but that's beside the point). The converter is actually a good example why it's a mint and not a loan because the total supply in a converter is still 0 when we mint the tokens on the exchange! The tokens cannot be loaned from anywhere because they don't exist yet. We say some amount of tokens will exist so we have them immediately and will then prove it later by transferring them in.

So I would say:
- Flash is still very much what you would expect because it solely describes the mechanism on the exchange level 
- Mint makes it clear that they are minted initially, though you correctly point out that the repayment mechanism works like a flash loan.

packages/loopring_v3/contracts/aux/bridge/IBridge.sol
```diff
@@ -42,8 +42,8 @@ interface IBridge
     ///      The sender will send the funds to Loopring exchange, so just like with normal
     ///      deposits the sender first has to approve token transfers on the deposit contract.
     ///
-    /// @param deposits The deposits
-    function batchDeposit(BridgeTransfer[] calldata deposits)
+    /// @param transfers The L2 transfers from Bridge to owners
+    function batchDeposit(BridgeTransfer[] calldata transfers)
```
Brechtpd, line 46: Isn't it a bit confusing to call these `transfers`, especially on the interface level with a function external people may want to use? People calling this function want to batch deposits, so the input is a bunch of `deposits`, that makes sense to me. The contract will handle them as efficiently as possible, but how this is done they don't have to worry to much about if they don't care.

But I can imagine having an input called `transfers` is confusing. And the new `param` description doesn't really help because the caller needs to specify a bunch of transfers from the bridge to the owners for some reason? Even when reading the complete description people wouldn't necessarily understand how this works because we would need to give more info in the description for this to make sense. But I don't see a reason to do so because people just want to do bunch of deposits together cheaper, no need to explain how it works exactly (which could always change).

That's why I still like to call them deposits, even internally sometimes. They are deposits, but the optimized implementation is to use L2 transfers but that's still just an implementation detail.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -120,41 +120,42 @@ contract Bridge is IBridge, ReentrancyGuard, Claimable
 
     modifier onlyFromExchangeOwner()
     {
-        require(msg.sender == exchangeOwner, "UNAUTHORIZED");
+        require(msg.sender == exchange.owner(), "UNAUTHORIZED");
```
Brechtpd, line 123: The reason to cache this is that this is/was x10 more expensive than a locally cached value.
However, now that we're after the Berlin HF and the gas prices have changed for calls/storage reads this may not be the case anymore. Because the gas cost new depends on other operations as well I don't know what the cheapest method is. I don't know yet if truffle supports Berlin gas costs, but if so I will measure both versions to see. I'll do the same with the token caching mechanism, that one may also be impacted in the same way.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
Brechtpd: The ones in the struct now map to the old variables outside of the struct. If variables are part of a struct or not is irrelevant for the storage slot the are assigned.

This

```
Struct Data
{
    uint dataA;
    uint dataB;
}

contract Ex
{
    Data   data;
    uint   dataC;
    uint   dataD;
    uint   dataE;
}
```

has the same storage layout as this:
```
Struct Data
{
    uint dataA;
    uint dataB;
    uint dataC;
    uint dataD;
}

contract Ex
{
    Data   data;
    uint   dataE;
}
```

So adding additional storage values after the struct creates new variables taking up new slots. The separate values in the old contract will map to the values in the struct in the new implemenation.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
dong77: I see, all you said above make sense, but it does take some explanation which is not ideal for Bridge developers (the two different 'mint' is confusing). If you don't like the term "loanDeposit", how about we do not use 'mint', or 'loan' for a). We can change the names as follows:

- flashMint => flashDeposit
- repayFlashMint => repayFlashDeposit
- getDepositLoanAmount => getFlashDepositAmount
- amountFlashMinted => flashDepositAmounts

The names are cooler than loan***.

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
dong77: Thank you for this knowledge which is good to know. I'll revert the change.

packages/loopring_v3/contracts/aux/bridge/IBridge.sol
dong77: I see your point, but the type is called BridgeTransfer but the variable name is called 'deposits', this is not consistent and is also confusing. So we have two options:
```
function batchDeposit(BridgeDeposit[] calldata deposits)
```

```
function batchDeposit(BridgeTransfer[] calldata transfers)
```

Which one is your pick？

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: If you find out the cost is higher, I will revert the change to have a local owner (whose value will be set to the exchange owner).

dong77: Brecht, how about we simply use  Claimable and allow the owenr to access onReceiveTransactions and trustConnector?

dong77: Same thing with Converter, how about we also simply check the owner(), not the exchange's owner. I believe converters and bridge needs to be added as Agent, if we do that, we already authorized them.

packages/loopring_v3/contracts/core/iface/ExchangeData.sol
Brechtpd: I'm fine with flashDeposit! (this was actually my original name, so we're back to the start haha :).

> but it does take some explanation which is not ideal for Bridge developers (the two different 'mint' is confusing). 

Normally developers building on the Bridge/BaseConverter will not know about any of the minting going in in the exchange or even in the converter, that's all abstracted away.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: `onReceiveTransactions` needs to be called by the exchange owner contract otherwise it cannot be called together with a block. `trustConnector` could indeed be called by anyone.

For the Converter we use the exchange owner as the intermediate holder of the vault tokens, so when the callback is called from the exchange owner contract it is simply `msg.sender` like usual so no special logic is needed. This could work differently, but this still seems to be an easy and efficient way to do things.

> I believe converters and bridge needs to be added as Agent, if we do that, we already authorized them.

Only the Bridge is an Agent (because L2 transactions are approved in it). Converts are not agents because all user transactions are all just simple L2 trades signed by them on L2, so nothing special about them on L1.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -243,27 +244,24 @@ contract Bridge is IBridge, ReentrancyGuard, Claimable
...
     }
 
-    function setConnectorTrusted(
+    function trustConnector(
         address connector,
         bool    trusted
         )
         external
-        onlyOwner
+        onlyFromExchangeOwner
```
Brechtpd, line 252: This won't work because the exchange owner is a `SelectorBasedAccessManager` which can only target a single contract which needs to be the exchange. So to be able to be the owner of all these converters we would have to implement extra functionality to be able to do so, which I don't think makes sense. Only our multisig should be able to call this function.

packages/loopring_v3/contracts/aux/bridge/IBridge.sol
Brechtpd: `function batchDeposit(BridgeDeposit[] calldata deposits)`!

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: So it seems support for Berlin is sill missing in the tools we use... Still an open issue on the github of truffle/ganache-cli and also remix doesn't expose it yet. But support should be coming pretty soon.

My approximation for the cost is 2,500 gas when not caching it, because we are on the cheap path with all contracts/storage already needing to be accessed for the block submission anyway. The cached gas cost would be 2,100 gas, so a bit cheaper. The other option is use an immutable value for it, but that makes deploying the contract harder and the implementation would need to be updated for each exchange owner change, so doesn't seem like a good idea. Because of the very small difference it does seem like caching the value isn't worth it any more so keeping your change.

For the token caching, it may be cheaper not to used the cached versions for the deposits, but still use the for the connector calls. Easy to change so would leave it with caching for now, will update if necessary after I measure it.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: Thank you for pointing this out. I'll revert the change.

dong77:
Brecht, I'm done with all the refractory and renaming. Now I'm going to review the logic.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -853,7 +575,7 @@ contract Bridge is IBridge, ReentrancyGuard, Claimable
     function verifySignatureL2(
         Context memory ctx,
         address        owner,
-        uint           _accountID,
+        uint           accountID,
```
Brechtpd, line 578: `accountID` now shadows the contract variable `BatchDepositor.accountID`... Compiler doesn't give a warning for this, but still probably best to avoid this I feel.

PR #2353, opened 2021-04-19
Split Bridge.sol into two files, one for BatchDeposits, one for connector calls.

dong77:


packages/loopring_v3/contracts/aux/bridge/IBridge.sol
```diff
@@ -5,34 +5,16 @@ pragma experimental ABIEncoderV2;
...
-
 struct BridgeDeposit
 {
     address owner;
     address token;
     uint96  amount;
 }
 
-/// @title  IBridge interface
+/// @title  IBatchDeposit interface
```
dong77, line 15: If this way of splitting code is reasonable, then BridgeDeposit should be named to something else, as it is a structure used by IBatchDeposit, not a Bridge.

Similarly, BridgeCall should probably be renamed to ConnectorCall -- we are calling connectors, not some contract calls the Bridge itself.

Brechtpd: I think this split makes sense, seems to work nicely.

Perhaps we should do `IBatchDeposit` -> `IBatchDepositor`. We can also move the `Deposit` struct inside this interface like this:

```
interface IBatchDepositor
{
    struct Deposit
    {
        address owner;
        address token;
        uint96  amount;
    }
}
```

Then we can just use IBatchDepositor.Deposit without worying about naming conflicts against other things called `Deposit`. Or we can just call the struct `BatchDeposit` not in the interface, I think that can also work.

`ConnectorCall` does indeed make more sense. Another option would be to call it a `BridgeTransaction`.

Brechtpd (approved):

PR #2357, opened 2021-04-20
Minor improvement8

dong77:


packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -227,11 +239,11 @@ contract Bridge is IBridge, BatchDepositor, Claimable
             transfersData := sub(transfersData, add(32, mul(34, mload(amounts))))
         }
         // Check if these transfers can be processed
-        bytes32 hash = _hashTransfers(transfersData);
-        require(!_arePendingDepositsTooOld(batch.batchID, hash), "BATCH_DEPOSITS_TOO_OLD");
+        bytes32 hash = _hashTransferBatch(batch.batchID, transfersData);
```
dong77, line 242: @Brechtpd , is this method to process the inbound transfers (users -> bridge)? If so, when is the corresponding hash be saved?

packages/loopring_v3/contracts/aux/bridge/BatchDepositor.sol
```diff
@@ -55,7 +55,7 @@ abstract contract BatchDepositor is IBatchDepositor, ReentrancyGuard
     uint32             public immutable accountID;
     IDepositContract   public immutable depositContract;
 
-    mapping (uint => mapping (bytes32 => uint)) public pendingDeposits;
+    mapping (bytes32 => uint)                   public pendingTransferBatches;
```
dong77, line 58: I thought this pendingTransferBatches only represents unhandled outbound L2 transfers (bridge -> users), but now I'm not sure...

Brechtpd: That's correct. These are pending deposits, so `batchDeposit` already moved the funds to the Bridge account on L2, and these transfers are for the distribution of these deposits on L2.

The source of these deposits could be 
- An external call to `batchDeposit`
- The deposits needed as returned by connectors
- If the connector call failed these will also be used to return funds to the users on L2

In all cases they are pending deposits needing to be processed on L2, so `pendingDeposits` seems like a good name to me, `pendingTransferBatches` more implementation specific so a bit less clear I think what it does using just the name.

packages/loopring_v3/contracts/aux/bridge/BatchDepositor.sol
```diff
@@ -287,21 +288,21 @@ abstract contract BatchDepositor is IBatchDepositor, ReentrancyGuard
...
     }
 
-    function _hashTransfers(bytes memory transfersData)
+    function _hashTransferBatch(uint batchID, bytes memory transfersData)
         internal
         pure
         returns (bytes32)
     {
-        return keccak256(transfersData);
+        return keccak256(abi.encodePacked(batchID, transfersData));
```
Brechtpd, line 296: I did the same in the beginning, but I changed it to keep the ID separate to save gas and keep things simple. `abi.encodePacked` isn't a cheap operation in most cases, especially when using it on data already in memory. So I would prefer it's not done this way because it increases gas costs.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: Any time a batch deposit is done, either from externally or internally by connectors, in `_batchDeposit`.

packages/loopring_v3/contracts/aux/bridge/BatchDepositor.sol
dong77: Thank you, will revert this change then.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -243,6 +255,9 @@ contract Bridge is IBridge, BatchDepositor, Claimable
         // Total amounts transferred to the bridge
         uint[] memory totalAmounts = new uint[](ctx.tokens.length);
 
+        // Verify withdrawals
+        _processWithdrawals(ctx, totalAmounts);
```
Brechtpd, line 259: So yeah these are done at the end because that's how things work on L2:
- First all transfers to the bridge
- Then we are able to withdraw funds from the bridge account

So `_processWithdrawals` after all transfers makes the most sense. Otherwise we wouldn't be able to run over all L2 transactions in order here and we'd have to jump around in `txsData`.

The current implementation also counts on the current order because `totalAmounts` is calculated using the transfers, so it's only correctly to pass this value into `_processWithdrawals` after all `_processConnectorCall` calls.

PR #2358, opened 2021-04-20
Hebaov2 metatx fix (replay bug)

kongliangzhong:
MetaTx can be replayed if  metaTx.nonce = 0 and didn't call a *WA method.

### DIFF
--- a/packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
+++ b/packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
@@ -26,7 +26,7 @@ library MetaTxLib
     using ERC20Lib      for Wallet;
 
     bytes32 public constant META_TX_TYPEHASH = keccak256(
-        "MetaTx(address relayer,address to,uint256 timestamp,address gasToken,uint256 gasPrice,uint256 gasLimit,uint256 gasOverhead,bytes data)"
+        "MetaTx(address relayer,address to,uint256 validUntil,address gasToken,uint256 gasPrice,uint256 gasLimit,uint256 gasOverhead,bytes data)"
     );
 
     event MetaTxExecuted(
@@ -39,7 +39,7 @@ library MetaTxLib
     struct MetaTx
     {
         address to;
-        uint    timestamp;
+        uint    validUntil;
         address gasToken;
         uint    gasPrice;
         uint    gasLimit;
@@ -62,7 +62,7 @@ library MetaTxLib
             META_TX_TYPEHASH,
             msg.sender,
             metaTx.to,
-            metaTx.timestamp,
+            metaTx.validUntil,
             metaTx.gasToken,
             metaTx.gasPrice,
             metaTx.gasLimit,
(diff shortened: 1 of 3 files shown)


Brechtpd:
It was feature that made is possible to save gas, similar to the v1 version of the contracts. But simplified without extra checks because I'm a believer that we should trust the data singed by a user in a transaction. If we cannot trust the data signed by a user, then what can we count on as a source of truth? Similar to how a user can sign a transfer to a "wrong" address, or pays an extremely high fee for the meta tx. I feel like there's no way to guard against "invalid" transactions, the only thing we can do whatever the user asked.

But I do think for L2 this cost saving feature isn't important, so if we don't care too much about L1 removing it makes sense.

However, only using a hash for replay protection does remove the ability for users to create multiple meta transactions that need to be executed in the correct expected order. If the user wants to do operations that need be done sequentially than that can now only be done using the batch feature in a single meta tx, all meta txs created by users can be submitted on-chain out of order. Maybe that's fine, meta transactions on L2 will normally not be used a lot in the future because users will be able to pay for transactions in any token they want, they will directly interact with the smart contract using normal L2 transactions (which will likely still be nonce based, but with account abstraction a lot should be possible).

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -26,20 +26,20 @@ library MetaTxLib
...
+        bytes32 metaTxHash,
         bool    success,
         uint    gasUsed
     );
 
     struct MetaTx
     {
         address to;
-        uint    nonce;
+        uint    timestamp;
```
Brechtpd, line 42: Maybe make this `validUntil`, and check the timestamp against this so meta transactions cannot be used forever.

kongliangzhong:
> It was feature that made is possible to save gas, similar to the v1 version of the contracts. But simplified without extra checks because I'm a believer that we should trust the data singed by a user in a transaction. If we cannot trust the data signed by a user, then what can we count on as a source of truth? Similar to how a user can sign a transfer to a "wrong" address, or pays an extremely high fee for the meta tx. I feel like there's no way to guard against "invalid" transactions, the only thing we can do whatever the user asked.
> 
> But I do think for L2 this cost saving feature isn't important, so if we don't care too much about L1 removing it makes sense.
> 
> However, only using a hash for replay protection does remove the ability for users to create multiple meta transactions that need to be executed in the correct expected order. If the user wants to do operations that need be done sequentially than that can now only be done using the batch feature in a single meta tx, all meta txs created by users can be submitted on-chain out of order. Maybe that's fine, meta transactions on L2 will normally not be used a lot in the future because users will be able to pay for transactions in any token they want, they will directly interact with the smart contract using normal L2 transactions (which will likely still be nonce based, but with account abstraction a lot should be possible).

Yep, I got your points.  There were no such requirements as `sequentially txs`, If there are, relayer should send the first tx, wait until it scceeded, and then send the next one.
I think the user may not pay attention to 'nonce' in metaTx, if the relayer gives a wrong nonce value (0), then this metatx can be replayed many times, which may cause the user to lose money. But he will pay attention to the amount and gas value.

In hebaov1 there is a field txAwareHash, if nonce == 0, then txAwareHash must be set. and if this field is set, it will be checked later in the approval validation function. This approach may save some gas but it adds some extra complexity to our contract code. Since gas is not our main concern, we should make it as simple as we can.

Brechtpd (approved):

PR #2359, opened 2021-04-20
Minor improvement9

dong77:


packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -160,8 +168,22 @@ contract Bridge is IBridge, BatchDepositor, Claimable
...
+        // to L1 to interact with connectors,
+        uint[]                      memory totalAmounts;
+        IBatchDepositor.Deposit[][] memory depositsList;
+        (totalAmounts, depositsList) = _processConnectorCalls(ctx, calls);
+
+        // Verify withdrawals
+        _verifyWithdrawals(ctx, totalAmounts);
+
+        // Do all resulting transfers back from the bridge to the users
+        _batchDeposit(address(this), depositsList);
```
Brechtpd, line 186: Any reason not to do this in a separate `_processConnectorCalls` call like before? `_processDepositBatches` and `_processConnectorCalls` or 2 completely separate operations and these don't have anything to do with each other so it makes sense to me to give each its own function and inputs.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
```diff
@@ -440,17 +465,18 @@ contract Bridge is IBridge, BatchDepositor, Claimable
...
+                        token:  connectorTx.token,
+                        amount: connectorTx.amount
                     });
                 }
             }
             assert(txIdx == totalNumCalls);
-            emit ConnectorCalled(call.connector, false, returnData);
         }
+
+        emit ConnectorTransacted(call.connector, check, success, returnData);
```
Brechtpd, line 479: Do we care if the check failed? The connector call will be done as if it was successful, so it doesn't impact anything. It can't hurt to know, but I don't really know what we'd do with it. If it's false the connector is written badly which could be good to know, but seems like the connector needs to be tested better then. :)

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: I think `_processDepositBatches` should be moved to `BatchDepositor`.

dong77: > _processDepositBatches

If we do this, then _readTransfers, and Context structure will also be moved to the depositor.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: It is weird to me if we don't even care about the failure while still calling the `getMinGasLimit` method. If we are certain the call will not fail, then we should not call `getMinGasLimit`, otherwise, we'd better know it failed then later figure out why.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: > Any reason not to do this in a separate `_processConnectorCalls` call like before? `_processDepositBatches` and `_processConnectorCalls` or 2 completely separate operations and these don't have anything to do with each other so it makes sense to me to give each its own function and inputs.

I don't feel strongly about my change, but I think it doesn't hurt to split the main logic inside the _processTransactions function to lay out the major steps. If there a performance impact, I will revert it back (are there?)

packages/loopring_v3/contracts/aux/bridge/IBatchDepositor.sol
```diff
@@ -14,6 +14,7 @@ interface IBatchDepositor
         address owner;
         address token;
         uint96  amount;
+        uint16  tokenID;
```
dong77, line 17: I added this optional field to get rid of `BatchDeposit` assuming this new field won't increase the data layout. Let me know if this works.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: > f we do this, then _readTransfers, and Context structure will also be moved to the depositor. But the Context is more Bridge specific, so I think we should keep it inside the Bridge code.

Oh, I see. It's a bit of a strange now that batch deposits request are made in contract A and they are processed in contract B, so the split doesn't really work 100% like it's currently implemented but that's fine.

> I don't feel strongly about my change, but I think it doesn't hurt to split the main logic inside the _processTransactions function to lay out the major steps. If there a performance impact, or if you prefer, I will revert it back (are there?)

I think I just see the split a bit differently then you. For me all batch deposit stuff is fully separate from all the bridge transactions stuff, so each is handled in it's own function, `_processTransactions` just makes sure the data they need is given correctly. It seems you see it more like one thing where batched deposits are an optional part of the complete flow. I don't really see it that way because the batched deposits that are processed in `_processTransactions` are unrelated to the bridge transactions. They could be processed in completely separate contracts (which is why the split makes sense), the only reason not to do so is performance.

It's fine for me like this, I don't think it has a performance impact but I'll have to measure it to be sure.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
Brechtpd: The way I see it's like people doing a metatx (which these bridge transactions basically are, just with some batching on top of them). Do relayers care that something users try to do will fail? Not really. But the relayer does need to follow the rules the user expects. And one of these rules is these gas limits, which the relayer needs to folllow. 

So yes, we will make sure the call will be successful, but of course users can't just count on us doing so without just trusting us. But if it would not be successful (e.g. gas limit is exceeded to try to calculate the approximation) than that's fine too. If we didn't create the Connector it's not our problem to fix (like if a user calls some dapp using a metatx that fails, we don't have to care why it failed).

packages/loopring_v3/contracts/aux/bridge/IBatchDepositor.sol
Brechtpd: I wouldn't do this because it can be very confusing, especially because it's part of the external interface. We already had to be very careful that the data passed in matches the correct values, because it's not because some `token` and `tokenID` is passed in from somewhere that `token == getTokenAddress(tokenID)`. So now in the code we are in the situation where:
- `token` is the only value we can use, we have to add comments when this is the case and follow them strictly
- `tokenID` is the only value we can use, we have to add comments when this is the case and follow them strictly
- we cannot depend on `token == getTokenAddress(tokenID)` because data is often just passed in from somewhere else, so they won't match

It seems this is quite error prone even for us in the Bridge. An extra struct only we normally use internally seems to be the lesser of two evils. It will also slightly increase gas costs.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: Brecht, feel free to create a PR for improving the split. For me, the split makes it easier to understand the code, the original Bridge.sol is too big. The BatchDepositor is not a complete feature by itself, that's why it is now an abstract contract.

packages/loopring_v3/contracts/aux/bridge/Bridge.sol
dong77: We'll talk about it on zoom. Basically we want to answer ourselves this question: why we call `getMinGasLimit`? Potential answers are:

1. We call it and if it fails, then we promise our users not to proceed with the `processProcessorTransactions`.
2. We call it, but even if it fails, we proceed (users don't care), but we'll learn something from the failure and optimize future gas calculations offchain.

I would say we chose 2.

packages/loopring_v3/contracts/aux/bridge/IBatchDepositor.sol
dong77: reverted

PR #2361, opened 2021-04-21
[hebaov2] metatx nonce check reinforcement

kongliangzhong:
@Brechtpd, After some discussion with Daniel, we believe that gas saving is also important even for L2. So I revert the previous changes and add some extra checks to the metaTx.nonce. please take another look.

### DIFF
--- a/packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
+++ b/packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
@@ -96,6 +96,8 @@ library MetaTxLib
         require(isNonceValid(wallet, msg.sender, metaTx.nonce, metaTx.data.toBytes4(0)), "INVALID_NONCE");
         if (metaTx.nonce != 0) {
             wallet.nonce[msg.sender] = metaTx.nonce;
+        } else {
+            require(metaTx.requiresSuccess, "META_TX_WITHOUT_NONCE_REQUIRES_SUCCESS");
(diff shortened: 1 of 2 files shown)


packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -86,12 +86,18 @@ library MetaTxLib
...
-        require(metaTx.validUntil >= block.timestamp, "METATX_EXPIRED");
 
         uint gasLeft = gasleft();
         require(gasLeft >= (metaTx.gasLimit.mul(64) / 63), "OPERATOR_INSUFFICIENT_GAS");
 
+        // Update the nonce before the call to protect against reentrancy
+        require(isNonceValid(wallet, msg.sender, metaTx.nonce, metaTx.data.toBytes4(0)), "INVALID_NONCE");
+        if (metaTx.nonce != 0) {
+            wallet.nonce[msg.sender] = metaTx.nonce;
+        }
```
Brechtpd, line 99: If we try to protect users from themselves we also need to add this here when `metaTx.nonce == 0`:
```
} else {
    require(metaTx.requiresSuccess, "META_TX_WITHOUT_NONCE_REQUIRES_SUCCESS");
}
```
If `metaTx.requiresSuccess == false` a `nonce == 0` metatx can be reused as many times as the relayer wants, while still getting paid the fee.

PR #2400, opened 2021-05-08
[hebaov2] Add resetGuardians method in SmartWallet

kongliangzhong:


packages/hebao_v2/contracts/base/libwallet/GuardianLib.sol
```diff
@@ -108,6 +111,48 @@ library GuardianLib
...
+    {
+        wallet.verifyApproval(
+            domainSeperator,
+            SigRequirement.MAJORITY_OWNER_REQUIRED,
+            approval,
+            abi.encode(
+                RESET_GUARDIANS_TYPEHASH,
+                approval.wallet,
+                approval.validUntil,
+                newGuardians
```
Brechtpd, line 146: Needs to be `keccak256(abi.encodePacked(newGuardians))`?

kongliangzhong: Sorry for the delay, fixed.

dong77:
Kongliang, please add some tests, especailly, test resetting guardians when there are pending guardians or when one guardian is already an existing active guardian.

Brechtpd (approved):

kongliangzhong:
> Kongliang, please add some tests, especailly, test resetting guardians when there are pending guardians or when one guardian is already an existing active guardian.

OK, will do this in another PR.

PR #2432, opened 2021-05-22
[Protocol 3.8] AMM v2

Brechtpd:
https://github.com/Loopring/team/issues/163

Most important practical changes:
- AMM weights are now used as the virtual balances. The virtual balances define the AMM curve. The circuits have been modified to remove the weight functionality. The only other change to the circuits is that AMM always needs to be enabled when the virtual balances of tokenS/tokenB are non-zero. The trade circuit now also updates the virtual balances by sbutracting/adding fillS/fillB just like the balances.
- AMM pools now always needs ending AMM updates for joins/exits which update the virtual balances as needed.
- It's possible to withdraw/deposit funds stored in the pool to L1 without changing the curve.

For the AMM pools the following was added:
- Some AMM pool logic is moved to an `IAmmController` which can implement custom virtual balance logic and can shut down the pool if necessary (needed for hybrid AMM pools which doesn't provide pool liquidity tokens itself).
- New AMM pool operation that allows setting the virtual balances, as defined by the `IAssetManager`.
- New AMM pool operations deposit/withdraw which allows the exchange owner to move funds between L1<->L2. Funds are stored in the pool contract on L1 and can be withdrawn to the `IAssetManager` which can use them productively.

So I think all existing pools and all new amplified pools can use the same `IAssetManager` and `IAmmController` contracts, so only a single implementation contract needs to be deployed. After updating the pool contracts with 3.8 the pool virtual balances need to be set to the correct values using the set virtual balances operation of the pool contract.

These pool settings are not stored in the AMM pool as storage but as immutable variables baked into the contract because the settings will be shared between many pools:
- For all non-hybrid pools there will likely be only a single `IAssetManager` which can handle all funds of all pools together (e.g. put all ETH of all pools into AAVE). I think it doesn't make sense to do this for each pool separately, funds of each pool are moved to this `IAssetManager` contract so L1 investing costs are shared.
- All pools of a certain kind can reuse the same `IAmmController`, making it easy and cheap to do updates to all pools depending on the contract if necessary. Because the logic implemented in the controller contract is rarely used (not for trades/joins/exits) the small extra overhead of doing the logic in an external contract is insignificant. Doing the logic in an external contract also makes it easier for using new data in storage without having to worry about different storage layouts in the pool contract implementations.


dong77:
> The trade circuit now also updates the virtual balances by sbutracting/adding fillS/fillB just like the balances.
After a swap, will virtual balances change for the pool?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -54,6 +71,17 @@ contract LoopringAmmPool is
         _;
     }
 
+    constructor(
+        IAmmController _controller,
```
kongliangzhong, line 75: add set methods for constructor parameters?

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
```diff
@@ -83,6 +103,9 @@ library AmmData
 
         Token[]  tokens;
         uint96[] tokenBalancesL2;
+        uint96[] vTokenBalancesL2;
```
dong77, line 106: tokenL2Balances
tokenL2VirtualBalances or tokenL2VBalances

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -188,8 +192,12 @@ library AmmJoinProcess
         // Calculate the amounts to deposit
         uint ratio = uint(AmmData.POOL_TOKEN_BASE).mul(mintAmount) / ctx.totalSupply;
 
+        uint newTotalSupply = ctx.totalSupply.add(mintAmount);
         for (uint i = 0; i < ctx.tokens.length; i++) {
             amounts[i] = (ratio.mul(ctx.tokenBalancesL2[i]) / AmmData.POOL_TOKEN_BASE).toUint96();
+
+            // Update virtual balances
+            ctx.vTokenBalancesL2[i] = (uint(ctx.vTokenBalancesL2[i]).mul(newTotalSupply) / ctx.totalSupply).toUint96();
```
dong77, line 200: The virtual balances always scale according to the new LP token balance. I believe with this approach, joins and exists will change the 'current price' in the pool. Correct? 

If so, is it possible that someone joins the pool to change the price in favor for his next trade?

dong77 (commented):
Brecht, could you share your simulation scripts so the team can run the program for better understanding of how the trades/joins/exists work?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
Brechtpd: These addresses will set immutable variables in the contract, so they will be hardcoded in the implementation contract itself, not in storage. All pools (especially now) will use the same `controller` and `assetManager`, so I don't really see any reason to make these parameters configurable or updateable. By hard-coding these addresses in the implementation contract they are automatically used in all pools. By making the `controller` and `assetManager` contract updateable as well we can update the logic in these contracts by updating a single contract (instead of having to update every pool contract independently).

packages/loopring_v3.js/src/request_processors/spot_trade_processor.ts
```diff
@@ -116,6 +125,15 @@ export class SpotTradeProcessor {
         .balance.iadd(s.fillBB)
         .isub(s.feeB);
 
+      // virtual balances
+      if (
+        accountB.getBalance(tokenA).weightAMM.gt(new BN(0)) &&
```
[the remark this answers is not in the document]
Brechtpd: I didn't update all references of "weight" to "virtual balance" because I think that would bring too many unnecessary changes to both the protocol and the relayer (would change the block format passed into the prover etc...), and so seemed more error prone than just to keep the name for all existing code. "weight" also not really that bad of a name even in the new system because it's still the weight of the token in the pool, just used a bit differently.

packages/loopring_v3/contracts/amm/IAmmController.sol
```diff
@@ -0,0 +1,31 @@
...
+        view
+        returns (uint96[] memory);
+
+    /// @dev Called by the pool contract when a SET_VIRTUAL_BALANCES operation is done
+    ///      on the pool.
+    /// @param tokenBalancesL2 The balances in the pool
+    /// @param vTokenBalancesL2 The current virtual balances in the pool
+    /// @return The new virtual balances in the pool
+    function getVirtualBalances(
+        uint96[] memory tokenBalancesL2,
```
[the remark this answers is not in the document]
Brechtpd: No, the `tokenBalancesL2` are just the real balances in the pool, and these values are not related to the values of the virtual balances in any way from the protocols view.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+
+/// @author Brecht Devos - <brecht@loopring.org>
+contract AmplifiedAmmController is IAmmController, Claimable
+{
+    using MathUint          for uint;
+    using MathUint96        for uint96;
+    using SafeCast          for uint;
+
+    uint public constant AMPLIFICATION_FACTOR_BASE = (10 ** 18);
```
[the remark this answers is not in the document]
Brechtpd: No, that's not necessary. It could be any value as this doesn't impact the protocol in any way. The protocol only sees the virtual balances (which are also base/decimals/... agnostic). This value is only important in the `AmplifiedAmmController` itself for how the AF values are stored and used.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    function getInitialVirtualBalances(
+        uint96[] memory joinAmounts
+        )
+        external
+        view
+        override
+        returns (uint96[] memory)
+    {
+        uint amplificationFactor = getAmplificationFactor(msg.sender);
```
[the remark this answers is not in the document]
Brechtpd: I only really implemented to special case when `AF == 0` ->  `AF =1` is so that all existing pools automatically keep the existing behavior without having to explicitly set the value to `0` in this contract. But setting the virtual balance of a token to `0` will disable the AMM in the circuit for the token (similar to previously how setting the weight to `0` also disabled AMM functionality for the pool with v1). So virtual balances being 0 is allowed, the pool just cannot be used when that's the case.

I don't really see how an AMM with virtual balance 0 or even very close to zero would be useful (unless for making it inactive)? A pool with the virtual balance set to `0` (or very close to `0`) would have infinite price.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,80 @@
...
+        uint96[] memory tokenBalancesL2,
+        uint96[] memory /*vTokenBalancesL2*/
+        )
+        external
+        view
+        override
+        returns (uint96[] memory)
+    {
+        // Only allow updating the virtual balances if the AF = 1
+        require(getAmplificationFactor(msg.sender) == AMPLIFICATION_FACTOR_BASE, "INVALID_OPERATION");
```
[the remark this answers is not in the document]
Brechtpd: Do you think there's a good reason to have a separate function? Would be 2 function calls instead of just one now so a bit more expensive, and making it fail in the function seems to have the same effect. Because the relayer needs to update all values as expected on-chain I also think it doesn't really help with making the transaction not fail because if not updated correctly the block will fail as well.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +101,110 @@ library AmmTransactionReceiver
...
+            ctx.approveAmmUpdates(true);
             S.processExit(
                 ctx,
                 abi.decode(data, (AmmData.PoolExit)),
                 signature
             );
+            ctx.approveAmmUpdates(false);
+        } else if (txType == AmmData.PoolTxType.SET_VIRTUAL_BALANCES) {
+            ctx.approveAmmUpdates(true);
+            ctx.vTokenBalancesL2 = ctx.settings.controller.getVirtualBalances(ctx.tokenBalancesL2, ctx.vTokenBalancesL2);
```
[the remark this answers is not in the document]
Brechtpd: The pool contracts allow the virtual balances to be updated with arbitrary values by the controller, so it just depends on what the controller does.

`AmplifiedAmmController.sol` currently only supports normal and amplified pools. For pools where we want to update the virtual balances differently either another controller needs to be used for the pool or `AmplifiedAmmController.sol` needs to be upgraded to allow additional functionality. 

The reason I did not allow arbitrary virtual balances updates in the current controller is because there is not a single good way to do that as it will need to depend on external data/mechanism to do so. Just allowing the virtual balances to be updated to whatever values would of course not be very secure for LPs.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
Brechtpd: No, joins/exits will keep the existing price. Like you said, the virtual balances are only scaled by joins/exits so the AMM price  will remain the same (because the AMM price if defined by the division of the virtual balances, which remain the same when  both virtual balances were multiplied by the same factor).

Brechtpd:
> Brecht, could you share your simulation scripts so the team can run the program for better understanding of how the trades/joins/exists work?

All code I have used is available here: https://github.com/Loopring/team/issues/163

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
kongliangzhong: OK, got it.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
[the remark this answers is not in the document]
dong77: When I ask this question, I was thinking there will be cases that changing virtual balances by the AMM owner is allowed (is it?), such as when one token's actual balance is 0. If that's the case, then having a method `allowSettingVirtualBalances` will make sense.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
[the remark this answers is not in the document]
dong77: Virtual balance being close to 0s doesn't make sense to me. I even think the virtual balances shall always be >= the actual balances, but I'm not sure if this is explicitly guaranteed/checked anywhere.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: You are right. I misunderstand it.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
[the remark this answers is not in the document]
dong77: >This value is only important in the AmplifiedAmmController itself for how the AF values are stored and used.
So if we choose a default value, `AmplifiedAmmController` wont' get AF == 1, right?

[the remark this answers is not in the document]
dong77: I see: the amplification factors are stored in the mapping inside `AmplifiedAmmController`.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+        // Just set the virtual balances to the actual balances
+        uint96[] memory vTokenBalancesL2 = new uint96[](tokenBalancesL2.length);
+        for (uint i = 0; i < tokenBalancesL2.length; i++) {
+            vTokenBalancesL2[i] = tokenBalancesL2[i];
```
dong77, line 55: The current implement seems to **reset** the virtual balances to the same value as the actual balances. Why not allow the owner to set them to arbitrary values (while keep the price)

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,80 @@
...
+import "./IAmmController.sol";
+import "../amm/LoopringAmmPool.sol";
+import "../lib/Claimable.sol";
+import "../lib/MathUint.sol";
+import "../lib/MathUint96.sol";
+import "../thirdparty/SafeCast.sol";
+
+
+/// @author Brecht Devos - <brecht@loopring.org>
+contract AmplifiedAmmController is IAmmController, Claimable
```
dong77, line 14: Once a pool has any LP token minted, changing its getAmplificationFactor value won't have any impact, except that SET_VIRTUAL_BALANCE can be transacted. I think I'm having a hard time figuring out why.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +101,110 @@ library AmmTransactionReceiver
...
             );
+            ctx.approveAmmUpdates(false);
         } else if (txType == AmmData.PoolTxType.EXIT) {
+            ctx.approveAmmUpdates(true);
             S.processExit(
                 ctx,
                 abi.decode(data, (AmmData.PoolExit)),
                 signature
             );
+            ctx.approveAmmUpdates(false);
```
dong77, line 118: Why wrapping this tx but not deposit/withdrawal?

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +101,110 @@ library AmmTransactionReceiver
...
+            ctx.approveAmmUpdates(true);
+            ctx.vTokenBalancesL2 = ctx.settings.controller.getVirtualBalances(ctx.tokenBalancesL2, ctx.vTokenBalancesL2);
+            ctx.approveAmmUpdates(false);
+        } else if (txType == AmmData.PoolTxType.DEPOSIT) {
+            AmmData.PoolDeposit memory poolDeposit = abi.decode(data, (AmmData.PoolDeposit));
+            require(poolDeposit.amounts.length == ctx.tokens.length, "INVALID_DEPOSIT_AMOUNTS");
+            for (uint i = 0; i < ctx.tokens.length; i++) {
+                deposit(S, ctx.tokens[i].addr, poolDeposit.amounts[i]);
+            }
+        } else if (txType == AmmData.PoolTxType.WITHDRAW) {
```
dong77, line 129: Need to learn more about deposit/withdrawal during our conversation.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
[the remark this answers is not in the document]
Brechtpd: It's not checked anywhere because it's valid, but I think virtual balances >= actual balances will indeed always be true in practice because otherwise some funds in the pool would be impossible to use.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: If AF == 1 then the virtual balances are the actual balances, so this can be done without changing the behavior of the pool and this is what will need to be done to update our current pools to the new virtual balance system, which is the main reason this function is implemented like it is. If AF != 1 then the pool was already created in the new system and no updates are necessary to the virtual balances after the pool has been created.

Allowing the virtual balances to be updated using different pools could be valid, but it won't be a normal amplified pool so I did not make it possible. Allowing the curve to change while funds are already in the pool will impact things like IL even when price remains the same so it's more of an actively managed pool where people would need to trust whatever controls what happens with the pool.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: > If AF == 1 then the virtual balances are the actual balances, so this can be done without changing the behavior of the pool and this is what will need to be done to update our current pools to the new virtual balance system, which is the main reason this function is implemented like it is. If AF != 1 then the pool was already created in the new system and no updates are necessary to the virtual balances after the pool has been created.

The current implementation does not allow changing the AF while the pool is being used for the safety of LPs.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
Brechtpd: We don't need balances/virtual balance L2 data on L1 to be able to correctly process deposits/withdrawals. Virtual balances also don't need to get updated just by moving funds between L1 and L2.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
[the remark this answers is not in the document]
Brechtpd: The amplification factor allows us to settle orders using reduced slippage, but we don't have _have_ to make use of it. You can see it as an upper bound for the risk of LPs. If we want to settle orders with higher slippage we can start doing so without any actual changes to the pool parameters.

> I think the biggest concern I have is a pool becomes illiquid (the actual balance of one token is 0). I wish we have a way to reuse the pool by forcing all LPs to exit then reinitializing it using a fair market price.

But wouldn't that just be a new pool? And LPs of the previous pool may still want to remain in the old pool for when price moves within range again (and if they would have added their liquidity to the new pool they will have suffered some kind of extra IL).

dong77:
@Brechtpd I like to summarize what we talked about last night, especially our objectives:

- The flexibility to use SET_VIRTUAL_BALANCE on L2 to change the virtual balances after pools are initialized, so we can re-use the same pool contract and the LP token. We can use a time lock to prevent ourselves from being evil.
- The relayer can decide to use the virtual balances as-is, or use smaller values (while keeping the ratio/price) that are >= actual balances. With this approach, any pool with a >1 initial AF will fall back to the full spectrum of the curve. In otherwords, the virtual balances are the maximum virtual balances. we can use any values between the actual balances and the max virtual balances.
- We need to handle the following scenario: Alice (as the only user) joins a pool, 50% of pool's assets are withdrawn to L1 and transferred out to the asset manager. Now Alice wants to exit the pool. **My understanding is that Alice's exit will fail if the amount is > 50%, is that right?.** (otherwise, if Alice can exit with all her LP, then she transferred his risk to other people on L2).

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +101,110 @@ library AmmTransactionReceiver
...
                 ctx,
                 abi.decode(data, (AmmData.PoolExit)),
                 signature
             );
+            ctx.approveAmmUpdates(false);
+        } else if (txType == AmmData.PoolTxType.SET_VIRTUAL_BALANCES) {
+            ctx.approveAmmUpdates(true);
+            ctx.vTokenBalancesL2 = ctx.settings.controller.getVirtualBalances(ctx.tokenBalancesL2, ctx.vTokenBalancesL2);
+            ctx.approveAmmUpdates(false);
+        } else if (txType == AmmData.PoolTxType.DEPOSIT) {
```
dong77, line 123: Brecht, deposit from Pool to L2 is done using the 'depositToL2' method, which is a L1 tx. So why do we still need this 'AmmData.PoolTxType.DEPOSIT' type?

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +101,110 @@ library AmmTransactionReceiver
...
+            AmmData.PoolDeposit memory poolDeposit = abi.decode(data, (AmmData.PoolDeposit));
+            require(poolDeposit.amounts.length == ctx.tokens.length, "INVALID_DEPOSIT_AMOUNTS");
+            for (uint i = 0; i < ctx.tokens.length; i++) {
+                deposit(S, ctx.tokens[i].addr, poolDeposit.amounts[i]);
+            }
+        } else if (txType == AmmData.PoolTxType.WITHDRAW) {
+            AmmData.PoolWithdrawal memory poolWithdrawal = abi.decode(data, (AmmData.PoolWithdrawal));
```
dong77, line 130: Where is such a AmmData.PoolTxType.WITHDRAW authorized? For Joins/exits, we do L2 signature validation, for this  WITHDRAW, do we also validate a L2 signature? It seems we don't have a L2 EDDSA key for the pool contract.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -162,6 +164,8 @@ library AmmJoinProcess
         amounts = new uint96[](ctx.tokens.length);
 
         if (ctx.totalSupply == 0) {
+            // Set virtual balances
+            ctx.vTokenBalancesL2 = ctx.settings.controller.getInitialVirtualBalances(join.joinAmounts);
```
dong77, line 168: On line 177, `ctx.tokenBalancesL2[i]` may not represent the real actual balances in this pool - some token may be withdrawn to L1 for DeFi. 

The real total actual balances should be **ctx.tokenBalancesL2[i] + tokens in Defi**

dong77 (dismissed):

packages/loopring_v3/contracts/amm/libamm/AmmData.sol
Brechtpd: Because balances also include L1 balances now, what do you think of just `balances` and `vBalances`?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -140,10 +183,29 @@ contract LoopringAmmPool is
...
+        onlyFromExchangeOwner
+    {
+        for(uint i = 0; i < state.tokens.length; i++) {
+            require(amounts[i] != 0, "INVALID_DEPOSIT_AMOUNT");
+            state.deposit(state.tokens[i].addr, amounts[i]);
+        }
+    }
+
+    function transferOut(
+        address to,
```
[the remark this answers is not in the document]
Brechtpd: Would be fine with me, though I guess it could be possible there are different "vault" contracts the asset manager manages depending on token or something

>  I don't think the manager should be an EOA address.

True, but just checking that the asset manager is a contract doesn't really have any security benefits I think because contracts are not necessarily more secure than a EOA.

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -140,10 +183,29 @@ contract LoopringAmmPool is
         state.withdrawWhenOffline();
     }
 
-    function updateExchangeOwnerAndFeeBips()
+    function depositAssetsToL2(
```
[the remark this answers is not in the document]
Brechtpd: Removed this function because this doesn't work in all cases (see explanation my PR).

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -140,10 +183,29 @@ contract LoopringAmmPool is
...
+        address to,
+        address token,
+        uint    amount
+        )
         external
         nonReentrant
+        onlyFromAssetManager
     {
-        state.updateExchangeOwnerAndFeeBips();
+        token.transferOut(to, amount);
```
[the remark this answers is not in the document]
Brechtpd: Hmmm although obviously useless not sure is failing the transaction is better than just not doing anything.

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -140,10 +183,29 @@ contract LoopringAmmPool is
         state.withdrawWhenOffline();
     }
 
-    function updateExchangeOwnerAndFeeBips()
+    function depositAssetsToL2(
+        uint96[] memory amounts
+        )
+        external
+        nonReentrant
+        onlyWhenOnline
```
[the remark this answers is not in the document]
Brechtpd: When the pool is offline all funds need to be returned to L1 so the LPs can withdraw on L1 (same behavior as before). So when the pool is offline we cannot allow the funds to be deposited back because than another force withdrawal would be needed to get the funds back to L1 (but the operator can then deposit to L2 again so could be a never ending loop if the operator wants to be difficult).

packages/loopring_v3/contracts/amm/libamm/AmmUpdateProcess.sol
```diff
@@ -54,13 +55,19 @@ library AmmUpdateProcess
...
                 "INVALID_AMM_UPDATE_TX_DATA"
             );
 
-            ctx.tokenBalancesL2[i] = /*tokenWeight*/uint96(packedDataB & 0xffffffffffffffffffffffff);
+            if (start) {
+                ctx.tokenBalancesL2[i] = /*balance*/uint96(packedDataB & 0xffffffffffffffffffffffff);
+                ctx.vTokenBalancesL2[i] = /*vbalance*/uint96((packedDataB >> 128) & 0xffffffffffffffffffffffff);
+            } else {
+                require(
+                    ctx.vTokenBalancesL2[i] == /*vbalance*/uint96((packedDataB >> 128) & 0xffffffffffffffffffffffff),
```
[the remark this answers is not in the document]
Brechtpd: We update these on L1 in the exact same way the balances are updated on L2 using transfers (so on L1 we use the amount transferred as provided in the DA), so they have to be the same, otherwise there is a bug somewhere.

This is unlike the virtual balances, which can be updated to any value on L1.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
Brechtpd: The only authorization needed is passing the checks in this contract, no L1 or L2 signatures are used. Only the operator can deposit/withdraw to/from the pool usable by the asset manager so no other authorization is required. Remember that transactions that are processed by the pool contract are automatically approved on L1 (and these are all conditional transactions which don't need any L2 signatures).

Brechtpd:
> We need to handle the following scenario: Alice (as the only user) joins a pool, 50% of pool's assets are withdrawn to L1 and transferred out to the asset manager. Now Alice wants to exit the pool. My understanding is that Alice's exit will fail if the amount is > 50%, is that right?. (otherwise, if Alice can exit with all her LP, then she transferred his risk to other people on L2).

That's correct. If the operator doesn't transfer the necessary funds to L2 to do the exit, Alice will have to withdraw her LP tokens and force the exit on L1 which forced the operator to make the necessary funds available on L2 at the risk of shutting down the pool if he still refuses. IIf he still refuses the pool shuts down and all funds can be withdrawn to the pool on L1 without the help of the operator.

Brechtpd:
#2445

dong77:
Brecht, I think we are over-designing the L1 liquidity feature and making things unnecessarily complicated with the current approach.

Assuming there is an *AssetManager* contract on L1 for managing some users' assets. The AssetManager can be designed to allow Loopring Exchange owner to make *capital calls* to withdraw some assets from L1 defi and automatically deposit these assets to L2 to join a predetermined Amm pool (*Capital returns* can also be done by the exchange owner on L2 to return funds to AssetManager and later trigger an L1 DeFi deposit).  The assets in the same AssetManager can also be called for multiple pools. I think what we need to add on top of Loopring 3.6 is simply **a way to deposit-then-join a pool from L1** with on-chain approvals.  In other words, from an AMM pool's perspective, the AssetManager address is a regular address, there is no need to handle its deposit/withdrawals/joins/exits differently.

Brechtpd:
Not sure I understand correctly. Do you mean that funds in the pool aren't used in both the AMM and L1 defi at the same time? To still achieve balancer v2 style capital efficiency we would have to manage the capital a bit like uniswap v3 "ticks" where for example only the funds of the current tick are on L2, and once the price has moved funds are removed from the current and the funds necessary for the next tick are moved to L2, while updating the curve amplified just for that tick. That's an interesting idea and I think that could work, not sure if that makes things easier or not. :) Or do we not care about capital efficiency like this?

Otherwise I think the idea is pretty close to the caspian design:
- People put money in a contract on L1 saying how they want it to be used in an AMM.
- All the bookkeeping is done in this contract. Funds are stored in defi, but also partially in the AMM pools on L2.
- Joins/exits on the pool level are not necessary because they will be handled on L1 and just setting the virtual balances is enough to update the curve on L2.
- Current method: Funds are moved between L1 and L2 without any issue because joins/exits from L2 are not allowed in these kinds of pools (this is where the `joinsDisabled` parameter comes in to support exactly this).
- Potentially new method: Change curve and joins/exits to achieve the same?

Actively managing the "ticks" instead of depositing/withdrawing could be pretty interesting. I guess wouldn't work very well for funds in our existing pools so people will probably have to exit and rejoin.

> Brecht, I think we are over-designing the L1 liquidity feature and making things unnecessarily complicated with the current approach.

I don't think the current method is that complicated because it really is just withdrawing/depositing from the pool, so we have to be careful with L1/L2 differences. Depending on how we want the AssetManager to work it would shift more complexity there which I think will be comparable (or with the tick system perhaps even a bit more complex).

> I think what we need to add on top of Loopring 3.6 is simply a way to deposit-then-join a pool from L1 for contract addresses with on-chain approvals. In other words, from an AMM pool's perspective, the AssetManager address is a regular address, there is no need to handle its deposit/withdrawals/joins/exits differently.

Joins/exits from contracts are possible. If we want people to be able to add funds from L2 we'd have to do something similar to AMM joins/exits, although depending how this is implemented we may or may not have the same difficulties tracking everything correctly on L1 (as long as we don't give users shares there shouldn't be any problem).

dong77 (dismissed):

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -94,19 +105,59 @@ library AmmTransactionReceiver
...
+             S.processWithdrawal(
+                ctx,
+                abi.decode(data, (AmmData.PoolWithdrawal))
+            );
         } else {
             revert("INVALID_POOL_TX_TYPE");
         }
     }
+
+    function processSetVirtualBalances(
```
dong77, line 145: Can we move this method to its own library file (AmmVirtualBalanceProcess.sol)

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,140 @@
...
+        returns (bool)
+    {
+        address pool = msg.sender;
+
+        // Check if a curve change was explicitly authorized
+        if (consumeCurveChangeAuthorized(pool)) {
+            return true;
+        }
+
+        if (getAmplificationFactor(pool) != AMPLIFICATION_FACTOR_BASE) {
```
dong77, line 63: How about we change `if (getAmplificationFactor(pool) != AMPLIFICATION_FACTOR_BASE) ` to `if (amplificationFactors[pool] != 0) ` so we only give special logic to uninitialized pools instead of AF=1 pools?

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -0,0 +1,140 @@
...
+    using MathUint          for uint;
+    using MathUint96        for uint96;
+    using SafeCast          for uint;
+
+    uint public constant AMPLIFICATION_FACTOR_BASE = (10 ** 18);
+
+    uint public constant CURVE_CHANGE_MIN_DELAY   = 7 days;
+    uint public constant CURVE_CHANGE_AUTH_WINDOW = 7 days;
+
+    mapping (address => uint) public amplificationFactors;
```
dong77, line 26: remove `public` since we have `getAmplificationFactor`?

dong77 (approved):
very minor change suggestions.

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
Brechtpd: Sure.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: I think that works for how we want to use it. But if the AF = 1 then the virtual balances should already match the actual balances so in practice it shouldn't make any difference.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: That makes sense.

dong77 (approved):

Issue #2438, opened 2021-05-26
[Loopring 3] exchange add method to withdraw unregistered tokens


dong77:
The exchange contract shall provide a function for the exchange owner to withdraw a specific token from the default deposit contract if the token is not registered in Loopring. This can help us handle L1-airdrop to people on our L2.

PR #2439, opened 2021-05-26
[Loopring 3] allow withdrawal of unregistered tokens

dong77:
https://github.com/Loopring/protocols/issues/2438


packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -166,6 +167,25 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
...
+        override
+        nonReentrant
+        onlyOwner
+    {
+        require(target != address(0), "INVALID_ADDRESS");
+        try state.getTokenID(token) {
+            revert("TOKEN_REGISTERED");
+        } catch {
+            uint _amount = amount == 0 ? ERC20(token).balanceOf(state.depositContract) : amount;
+            state.depositContract.transfer(address(0), to, token, _amount);
```
dong77, line 185: I think `address(0)` should be changed to `state.depositContract`?

packages/loopring_v3/contracts/core/impl/ExchangeV3.sol
```diff
@@ -662,6 +647,40 @@ contract ExchangeV3 is IExchangeV3, ReentrancyGuard
...
+        override
+        nonReentrant
+        onlyOwner
+    {
+        try this.getTokenID(token) {
+            revert("TOKEN_REGISTERED");
+        } catch {
+            require(to != address(0), "INVALID_ADDRESS");
+            require(amount > 0, "INVALID_AMOUNT");
+            state.depositContract.transfer(address(0), to, token, amount);
```
dong77, line 680: @Brechtpd shall I change `address(0)` to `address(state.depositContract)`?

Brechtpd: Instead of using `transfer` I think it's best to just use `withdraw` so we don't have to worry about token approvals:
```
state.depositContract.withdraw(address(state.depositContract), to, token, amount, new bytes(0));
```

dong77:
Please take a look again.

Brechtpd (approved):

PR #2445, opened 2021-05-31
Amm v2 - AssetManager fix

Brechtpd:
#2432

Okay, I implemented this in a way that doesn't work twice now, so hopefully the third time is the charm.

So the main problem is that joins/exits need to work on the actual balances in the pool, which (if an AssetManager is set) could be both on L2 and L1. 

The other problem is to correctly implement this, because L1 transactions like deposit/withdraw and L2 transactions do not automatically sync up correctly because of a number of reasons. For example, let's say we process 3 pool transactions in a block:
1. An exit
1. A withdrawal
1. Another exit

What are the pool balances? For 1) it would be the balances as reported on L2. For 3) however it would be the balances on L2 + the newly withdrawn funds in 2). But the funds withdrawn in 2) could either be already in the pool contract (when the block is submitted before the callback) or not yet in the pool contract (the block is submitted after the callback). So only depending on the actual token balances on L1 is not reliable because we do not know what the balances are for each L2 transaction because it depends on when withdrawals/deposits are handled on L2. The same problem is true for deposits (and could be even worse because the deposit may keep on pending).

And so I think the only way to do this correctly is to keep track of the balances deposited/withdrawn to/from the pool in exactly the same order they are done on L2. For that we use `mapping (address => uint96) balancesL1;` and so we just need to add this value to the L2 balances to get the correct balance in the pool. Because these values need to be updated next to their L2 counterparts deposits are now also required to be done with the `PoolTxType.DEPOSIT` pool instruction.

The pool contracts will only watch the `balancesL1` values. The AssetManager is allowed to update these values when required. It's not really possible to have fully dynamic L1 balances (e.g. always up to date with latest interest) because these values need to be deterministic when the operator processes joins/exits. And so by just updating the `balancesL1` at certain expected intervals also makes sure the relayer can count on them being the same when needed.

To make the deposits work as other L2 transactions on L2 (do a deposit and also see the transaction the block in the pool contract) I have made it possible again to do the `txReceiverCallbacks` before the block is submitted again.

I _think_ this works correctly now, but I'll have to think about it some more to make sure.

### DIFF
--- a/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
+++ b/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
@@ -107,6 +107,7 @@ contract LoopringAmmPool is
         external
         onlyFromController
     {
+        require(state.exitMode != enabled, "INVALID_STATE");
         state.exitMode = enabled;
     }
 
--- a/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
+++ b/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
@@ -7,12 +7,13 @@ import "../aux/access/ITransactionReceiver.sol";
 import "../core/iface/IAgentRegistry.sol";
 import "../lib/ReentrancyGuard.sol";
 import "../lib/TransferUtil.sol";
-import "./libamm/AmmTransactionReceiver.sol";
+import "./libamm/AmmAssetManagement.sol";
 import "./libamm/AmmData.sol";
 import "./libamm/AmmExitRequest.sol";
 import "./libamm/AmmJoinRequest.sol";
 import "./libamm/AmmPoolToken.sol";
 import "./libamm/AmmStatus.sol";
+import "./libamm/AmmTransactionReceiver.sol";
 import "./libamm/AmmWithdrawal.sol";
 import "./PoolToken.sol";
 
@@ -24,11 +25,12 @@ contract LoopringAmmPool is
     ITransactionReceiver,
     ReentrancyGuard
 {
-    using AmmTransactionReceiver for AmmData.State;
+    using AmmAssetManagement     for AmmData.State;
     using AmmJoinRequest         for AmmData.State;
     using AmmExitRequest         for AmmData.State;
     using AmmPoolToken           for AmmData.State;
     using AmmStatus              for AmmData.State;
+    using AmmTransactionReceiver for AmmData.State;
     using AmmWithdrawal          for AmmData.State;
     using TransferUtil           for address;
 
@@ -38,7 +40,7 @@ contract LoopringAmmPool is
     event Shutdown(uint timestamp);
 
     IAmmController public immutable controller;
-    address        public immutable assetManager;
+    IAssetManager  public immutable assetManager;
     bool           public immutable joinsDisabled;
 
     modifier onlyFromExchangeOwner()
@@ -49,7 +51,7 @@ contract LoopringAmmPool is
 
     modifier onlyFromAssetManager()
     {
-        require(msg.sender == assetManager, "UNAUTHORIZED");
+        require(msg.sender == address(assetManager), "UNAUTHORIZED");
         _;
     }
 
@@ -73,7 +75,7 @@ contract LoopringAmmPool is
 
     constructor(
         IAmmController _controller,
-        address        _assetManager,
+        IAssetManager  _assetManager,
         bool           _joinsDisabled
     )
     {
@@ -183,20 +185,6 @@ contract LoopringAmmPool is
         state.withdrawWhenOffline();
     }
 
-    function depositAssetsToL2(
-        uint96[] memory amounts
-        )
-        external
-        nonReentrant
-        onlyWhenOnline
-        onlyFromExchangeOwner
-    {
-        for(uint i = 0; i < state.tokens.length; i++) {
-            require(amounts[i] != 0, "INVALID_DEPOSIT_AMOUNT");
-            state.deposit(state.tokens[i].addr, amounts[i]);
-        }
-    }
-
     function transferOut(
         address to,
         address token,
@@ -206,6 +194,27 @@ contract LoopringAmmPool is
         nonReentrant
         onlyFromAssetManager
     {
-        token.transferOut(to, amount);
+        state.transferOut(to, token, amount);
+    }
+
+    function setBalanceL1(
+        address token,
+        uint96  balance
+        )
+        external
+        nonReentrant
+        onlyFromAssetManager
+    {
+        state.balancesL1[token] = balance;
+    }
+
+    function getBalanceL1(
+        address token
+        )
+        external
+        view
+        returns (uint96)
+    {
+        return state.balancesL1[token];
     }
 }
--- a/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
+++ b/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
@@ -79,6 +79,7 @@ contract LoopringAmmPool is
         bool           _joinsDisabled
     )
     {
+        require(_controller != IAmmController(0), "ZERO_ADDRESS");
         controller = _controller;
         assetManager = _assetManager;
         joinsDisabled = _joinsDisabled;
@@ -98,7 +99,15 @@ contract LoopringAmmPool is
         external
         nonReentrant
     {
+        require(state.accountID == 0 || msg.sender == address(controller), "UNAUTHORIZED");
         state.setupPool(config);
+    }
+
+    function enterExitMode(bool enabled)
+        external
+        onlyFromController
+    {
+        state.exitMode = enabled;
     }
 
     // Anyone is able to shut down the pool when requests aren't being processed any more.
@@ -113,7 +122,6 @@ contract LoopringAmmPool is
 
     function shutdownByController()
         external
-        payable
         onlyWhenOnline
         nonReentrant
         onlyFromController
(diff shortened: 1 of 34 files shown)


dong77:
>  But the funds withdrawn in 2) could either be already in the pool contract (when the block is submitted before the callback) or not yet in the pool contract (the block is submitted after the callback). 

Can we remove this flexibility to simplify things?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -183,29 +185,36 @@ contract LoopringAmmPool is
...
         onlyFromAssetManager
     {
-        token.transferOut(to, amount);
+        state.balancesL1[token] = balance;
+    }
+
+    function getBalanceL1(
+        address token
+        )
+        external
```
dong77, line 214: public

dong77 (commented):
Is the new idea keeping track of tokens that have been successfully deposited back to L2 (substracting `balancesL1`) and withdrawn to L1 (adding to `balancesL1`) and also allowing the AssetManager to refresh `balancesL1` based on L1 DeFi data?

It seems the new bulk work is to processing pool deposits and withdrawals, correct?

packages/loopring_v3/contracts/amm/libamm/AmmDepositProcess.sol
```diff
@@ -0,0 +1,74 @@
...
+        AmmData.State       storage S,
+        AmmData.Context     memory ctx,
+        AmmData.PoolDeposit memory poolDeposit
+        )
+        internal
+    {
+        require(poolDeposit.amounts.length == ctx.tokens.length, "INVALID_DEPOSIT_DATA");
+        for (uint i = 0; i < ctx.tokens.length; i++) {
+            verifyDepositTx(ctx, ctx.tokens[i].tokenID, poolDeposit.amounts[i]);
+            S.deposit(ctx.tokens[i].addr, poolDeposit.amounts[i]);
```
dong77, line 34: use local variables for `ctx.tokens[i].addr` and `poolDeposit.amounts[i]`

packages/loopring_v3/contracts/amm/libamm/AmmDepositProcess.sol
```diff
@@ -0,0 +1,74 @@
...
+        uint96 txAmount;
+        assembly {
+            packedData := calldataload(txsDataPtr)
+            txAmount := calldataload(add(txsDataPtr, 12))
+        }
+
+        require(
+            // txType == ExchangeData.TransactionType.DEPOSIT &&
+            // owner == address(this) &&
+            // accountID == crx.accountID &&
```
dong77, line 65: `crx` => `ctx`

packages/loopring_v3/contracts/aux/access/LoopringIOExchangeOwner.sol
```diff
@@ -143,11 +144,18 @@ contract LoopringIOExchangeOwner is SelectorBasedAccessManager, ERC1271, Drainab
...
+            // Do transaction verifying blocks callbacks
+            _verifyTransactions(blocks, txReceiverCallbacks);
+        }
+
         // Submit blocks
         target.fastCallAndVerify(gasleft(), 0, decompressed);
 
-        // Do transaction verifying blocks callbacks
-        _verifyTransactions(blocks, txReceiverCallbacks);
+        if (!txReceiverCallbacks.before) {
```
dong77, line 155: before => beforeBlockSubmission

packages/loopring_v3/contracts/test/TestAssetManager.sol
```diff
@@ -6,15 +6,31 @@ import "../lib/Claimable.sol";
...
     using MathUint         for uint;
     using TransferUtil     for address;
 
     mapping(address => mapping(address => uint)) public poolBalances;
 
+    function getBalances(
+        address          pool,
+        address[] memory tokens
+        )
+        external
```
dong77, line 23: public

packages/loopring_v3/contracts/amm/libamm/AmmUpdateProcess.sol
```diff
@@ -72,6 +77,13 @@ library AmmUpdateProcess
             txsDataPtr += ExchangeData.TX_DATA_AVAILABILITY_SIZE;
         }
 
+        if (start && ctx.settings.assetManager != IAssetManager(0)) {
```
dong77, line 80: what if `assetManager` is non-zero, then `S.balancesL1[...] ` is non-zero, then we set `assetManager` to zero, now `S.balancesL1[...] ` will still be non-zero, shall we still do the `add` operation？

Brechtpd:
> Is the new idea keeping track of tokens that have been successfully deposited back to L2 (substracting balancesL1) and withdrawn to L1 (adding to balancesL1) and also allowing the AssetManager to refresh balancesL1 based on L1 DeFi data?

Yes exactly.  Keeping track of those balances on L1 in exactly the same way the balances change on L2. And tI guess it should be even easier for the relayer to keep track of the funds on L1 because it only depends on `balancesL1` instead of also some external dependency (which would either be ERC20 token balances in the contracts or some stats from an external dapp).

> It seems the new bulk work is to processing pool deposits and withdrawals, correct?

Yes, only the implementation of those pool operations have changed, but on not that different  on a high level than before.

packages/loopring_v3/contracts/amm/libamm/AmmUpdateProcess.sol
Brechtpd: It's currently impossible to update the `assetManager` address in the pool, it's an `immutable` value hardcoded in the implementation contract.

If we would allow setting the `assetManager` back to zero, I would say we should make that impossible if `balancesL1` are non-zero.

dong77: Keep it as immutable is fine.

packages/loopring_v3/contracts/amm/libamm/AmmAssetManagement.sol
```diff
@@ -0,0 +1,52 @@
...
+        public
+    {
+        if (amount == 0) {
+            return;
+        }
+        uint ethValue = 0;
+        if (token == address(0)) {
+            ethValue = amount;
+        } else {
+            ERC20(token).approve(address(S.exchange.getDepositContract()), amount);
```
dong77, line 31: maybe we can try `S.exchange.deposit` and if it fails, do `approve` then ``S.exchange.deposit` for saving gas.

packages/loopring_v3/contracts/amm/libamm/AmmDepositProcess.sol
```diff
@@ -0,0 +1,76 @@
...
+        AmmData.State       storage S,
+        AmmData.Context     memory ctx,
+        AmmData.PoolDeposit memory poolDeposit
+        )
+        internal
+    {
+        require(poolDeposit.amounts.length == ctx.tokens.length, "INVALID_DEPOSIT_DATA");
+        for (uint i = 0; i < ctx.tokens.length; i++) {
+            address token = ctx.tokens[i].addr;
+            uint96 amount = poolDeposit.amounts[i];
```
dong77, line 34: how about we check `if (amount > 0)` then do the deposit and update of `balancesL1`, otherwise just skip.

packages/loopring_v3/contracts/amm/libamm/AmmDepositProcess.sol
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        require(
+            // txType == ExchangeData.TransactionType.DEPOSIT &&
+            // owner == address(this) &&
+            // accountID == ctx.accountID &&
+            // tokenID == tokenID &&
+            packedData & 0xffffffffffffffffffffffffffffffffffffffffffffffffffffff == (uint(ExchangeData.TransactionType.DEPOSIT) << 208) | (uint(address(this)) << 48) | (uint(ctx.accountID) << 16) | uint(tokenID) &&
+            amount == txAmount,
+            "INVALID_DEPOSIT_TX_DATA"
```
dong77, line 71: INVALID_AMM_DEPOSIT_TX_DATA

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawProcess.sol
```diff
@@ -0,0 +1,84 @@
...
+        AmmData.Context        memory ctx,
+        AmmData.PoolWithdrawal memory poolWithdrawal
+        )
+        internal
+    {
+        require(ctx.settings.assetManager != IAssetManager(0), "CANNOT_WITHDRAW_FROM_POOL");
+        require(poolWithdrawal.amounts.length == ctx.tokens.length, "INVALID_WITHDRAWAL_AMOUNTS");
+        for (uint i = 0; i < ctx.tokens.length; i++) {
+            address token = ctx.tokens[i].addr;
+            uint96 amount = poolWithdrawal.amounts[i];
```
dong77, line 32: same feedback: check `if (amount >0)` here

dong77 (approved):

packages/loopring_v3/contracts/amm/libamm/AmmDepositProcess.sol
Brechtpd: Deposits/withdrawals transactions were already skipped when `amount == 0` but you're right some other things can be skipped as well. Updated.

packages/loopring_v3/contracts/amm/libamm/AmmAssetManagement.sol
Brechtpd: Now with Berlin live I think that may actually end up using more gas than just always approving exactly what is needed. The approval amount is immediately consumed and reset to 0 by the deposit so there are no high storage costs. The token contract needs to be called already for the `transferFrom` so the call to `approve` will also be cheap.

Issue #2449, opened 2021-06-02
[Protocol 3] Allow AMM pool owner to force exits 


dong77:
We want to expire some AMM pools because the liquidity there is really small. But to move them off our relayer system, we still need to wait long enough for the LPs to exit. I propose that we can allow the owner of the pool to authorize exits with zero exit fees.

Once an AMM pool is expired, it will be great if we can re-config it with different tokens and parameters to reuse the pool account and its LP token on L2 for a new trading pair.

PR #2450, opened 2021-06-02
Amm v2 - Exit mode and pool reuse

Brechtpd:
#2449

Allows the controller to enter exit mode, which allows the operator to exit users from the pool if no fee needs to be paid.
Also allows the controller to re-use the pool with different tokens when the supply is `0`.

I used the controller as the authorized address because pools currently don't have an owner and exit mode/pool re-using a pool even with no supply may not work for all types of pools.

Made `permit` public to make the Pool contract smaller because the contract size was too large.

### DIFF
--- a/packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
+++ b/packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
@@ -20,7 +20,13 @@ contract AmplifiedAmmController is IAmmController, Claimable
 
     uint public constant AMPLIFICATION_FACTOR_BASE = (10 ** 18);
 
+    uint public constant MIN_CURVE_CHANGE_DELAY   = 7 days;
+    uint public constant CURVE_CHANGE_AUTH_WINDOW = 7 days;
+
     mapping(address => uint) public amplificationFactors;
+
+    mapping(address => uint) public curveChangeAuthorization;
+
 
     function getInitialVirtualBalances(
         uint96[] memory joinAmounts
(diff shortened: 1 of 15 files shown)


packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -77,4 +78,21 @@ contract AmplifiedAmmController is IAmmController, Claimable
...
+        LoopringAmmPool pool,
+        AmmData.PoolConfig calldata config
+        )
+        external
+        onlyOwner
+    {
+        pool.setupPool(config);
+    }
+
+    function enableExitMode(LoopringAmmPool pool)
```
dong77, line 92: enterExitMode?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -98,9 +98,17 @@ contract LoopringAmmPool is
         external
         nonReentrant
     {
+        require(state.accountID == 0 || msg.sender == address(controller), "UNAUTHORIZED");
```
dong77, line 101: even when state.accountID == 0, should we still only allow **only** the controller can set up the pool?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -98,9 +98,17 @@ contract LoopringAmmPool is
...
         state.setupPool(config);
     }
 
+    function enableExitMode()
+        external
+        onlyFromController
```
dong77, line 107: now with  exit mode, maybe we should check `controller` is not 0x0 in the constructor.

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
```diff
@@ -71,16 +71,26 @@ library AmmStatus
...
-        S.balanceOf[address(this)] = AmmData.POOL_TOKEN_MINTED_SUPPLY;
-        S.allowance[address(this)][address(exchange.getDepositContract())] = type(uint256).max;
-        exchange.deposit(
-            address(this), // from
-            address(this), // to
-            address(this), // token
-            uint96(AmmData.POOL_TOKEN_MINTED_SUPPLY),
-            new bytes(0)
-        );
+        bool newPool = (S.accountID == 0);
```
dong77, line 74: when/where is this `S.accountID` set to 0?

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
```diff
@@ -71,16 +71,26 @@ library AmmStatus
...
+            S.balanceOf[address(this)] = AmmData.POOL_TOKEN_MINTED_SUPPLY;
+            S.allowance[address(this)][address(exchange.getDepositContract())] = type(uint256).max;
+            exchange.deposit(
+                address(this), // from
+                address(this), // to
+                address(this), // token
+                uint96(AmmData.POOL_TOKEN_MINTED_SUPPLY),
+                new bytes(0)
+            );
+        }
```
dong77, line 93: in method `_shutdown`, shall we assert `S.shutdownTimestamp == 0` ?

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
```diff
@@ -71,16 +71,26 @@ library AmmStatus
...
-        S.allowance[address(this)][address(exchange.getDepositContract())] = type(uint256).max;
-        exchange.deposit(
-            address(this), // from
-            address(this), // to
-            address(this), // token
-            uint96(AmmData.POOL_TOKEN_MINTED_SUPPLY),
-            new bytes(0)
-        );
+        bool newPool = (S.accountID == 0);
+        if (newPool) {
```
dong77, line 75: This logic seems incorrect to me. No matter if this is a new pool or an existing pool to be reused, we just need to check that the LP token balance is 0. For reusable pools, we may need to force the pool to use the same accountID, as the pool's address is the same, so its accountID should stay unchanged.


```solidity
require(S.poolTokenBalance == 0),
if (S.accountID == 0) { // new pool
  require (config.accountID > 0);
  S.accountID = config.accountID;
} else { // reused pool
  require(S.shutdownTimestamp > 0);
  require(config.accountID == 0 || config.accountID == S.accountID)
  S.shutdownTimestamp = 0;
  S.exitMode = false;
}
// all other parameter setup
```

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
dong77: should we do enterExitMode(bool enter) so we can also exit from Exit mode?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
Brechtpd: Currently `setupPool` is called directly after the proxy contract is created in `LoopringAmmFactory`, so that would have to be changed. But I think it should be fine, I don't think there's any risk with allowing anyone to call setupPool because it's called in the same Ethereum tx as the proxy is created.

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
Brechtpd: It's only set in the first call to `setupPool` here, it's value can never be changed even if the pool gets reused.

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
Brechtpd: `shutdown` and `shutdownByController` are already `onlyWhenOnline`.

packages/loopring_v3/contracts/amm/libamm/AmmStatus.sol
Brechtpd: > This logic seems incorrect to me. No matter if this is a new pool or an existing pool to be reused, we just need to check that the LP token balance is 0.

I think that's what my code also does?

> For reusable pools, we may need to force the pool to use the same accountID, as the pool's address is the same, so its accountID should stay unchanged.

The current code only sets the accountID for new pools, it ignores the passed in value in `config` when the pool is reused.

I don't think a requirement for pool reuse is that the pool _needs_ to have been shutdown. But otherwise your code and the current code seems the same or did I miss anything?

dong77: I'll propose a change with a question in another PR

dong77 (approved):

PR #2454, opened 2021-06-03
Amm v2 - More flexible virtual balances

Brechtpd:
> The flexibility to use SET_VIRTUAL_BALANCE on L2 to change the virtual balances after pools are initialized, so we can re-use the same pool contract and the LP token. We can use a time lock to prevent ourselves from being evil.

Changed the interface to allow updating the virtual balances more easily. Also implemented a very simple mechanism in `AmplifiedAmmController.sol` but could easily be updated to support other methods if we need it.

> The relayer can decide to use the virtual balances as-is, or use smaller values (while keeping the ratio/price) that are >= actual balances. With this approach, any pool with a >1 initial AF will fall back to the full spectrum of the curve. In otherwords, the virtual balances are the maximum virtual balances. we can use any values between the actual balances and the max virtual balances.

Did some testing and this works. For example swaps can be settled as if the AF = 1 if the virtual balances are configured for AF = 2. Note that if this is done the price as enforced by the pool will lag the enforced price, which can be seen as a buffer in which we can more freely decide on the prices offered to swappers while reducing IL. 

I'm also trying to make it possible to do the first join in a more flexible way. Currently the first joiner always needs to add liquidity in the center of the supported range. It would be nice if that's not needed and the first joiner can add liquidity off-center and add liquidity at the price he wants, but just in the predefined range (e.g. the price range for ETH/USDC is between $5k and $10k, so the first joiner could just add ETH at $5k without needing any USDC). It's of course also possible to just add a tiny bit of liquidity first and move the curve to the correct position but would be nicer if something like that's not needed. But I haven't figured out the formula yet to easily allow this.

Moved ECDSA signature checking code (which isn't often used as far as I know) to an external library to reduce the contract size.

### DIFF
--- a/packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
+++ b/packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
@@ -20,7 +20,8 @@ contract AmplifiedAmmController is IAmmController, Claimable
 
     uint public constant AMPLIFICATION_FACTOR_BASE = (10 ** 18);
 
-    uint public constant MIN_CURVE_CHANGE_DELAY = 3 days;
+    uint public constant MIN_CURVE_CHANGE_DELAY   = 7 days;
+    uint public constant CURVE_CHANGE_AUTH_WINDOW = 7 days;
 
     mapping(address => uint) public amplificationFactors;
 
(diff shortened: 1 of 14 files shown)


packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -38,24 +43,41 @@ contract AmplifiedAmmController is IAmmController, Claimable
...
+
+        // Remove authorization
+        delete curveChangeAuthorization[msg.sender];
 
-        // Just set the virtual balances to the actual balances
-        uint96[] memory vTokenBalancesL2 = new uint96[](tokenBalancesL2.length);
-        for (uint i = 0; i < tokenBalancesL2.length; i++) {
-            vTokenBalancesL2[i] = tokenBalancesL2[i];
+        // Special case: Always allow updating the virtual balances if the AF = 1
+        if (!authorized && getAmplificationFactor(msg.sender) == AMPLIFICATION_FACTOR_BASE) {
```
dong77, line 63: if `authorized==true`, this method will return false, which doesn't seem right to me.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -38,24 +43,41 @@ contract AmplifiedAmmController is IAmmController, Claimable
...
+            for (uint i = 0; i < balances.length; i++) {
+                if(vBalancesNew[i] != balances[i]) {
```
dong77, line 65: I don't understand the logic here. Please describe when new virtual balances are allowed.

Here is my understanding:

```solidity
if (!authorized) return false;
if (vBalancesNew.length != balances.length) return false;
for (uint i = 0; i < balances.length; i++) {
  if (vBalancesNew[i] < balances[i])  return false;
}

return true;
```

packages/loopring_v3/contracts/amm/libamm/AmmTransactionReceiver.sol
```diff
@@ -138,4 +142,24 @@ library AmmTransactionReceiver
             revert("INVALID_POOL_TX_TYPE");
         }
     }
+
+    function processSetVirtualBalances(
+        AmmData.State               storage /*S*/,
```
dong77, line 147: remove the first parameter?

Brechtpd:
Added all Feedback to this PR to make future merging of different PRs easier.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: Yes I think there was a mistake in the logic. I simplified the logic similarly to how you did it here.

Updating virtual balances is only allowed when:
- `authorizeCurveChange` was called and the current timestamp is in the supported range
- `AF == 1` and the virtual balances are set to the current balances

dong77 (approved):

PR #2460, opened 2021-06-05
AMM v2- some minor improvements

dong77:
- Add checking that virtual balance is always non-zero.
- Removed AmmData.Settings

### DIFF
--- a/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
+++ b/packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
@@ -178,12 +178,7 @@ contract LoopringAmmPool is
         // nonReentrant     // Not needed, does not do any external calls
                             // and can only be called by the exchange owner.
     {
-        AmmData.Settings memory settings = AmmData.Settings({
-            controller: controller,
-            assetManager: assetManager,
-            joinsDisabled: joinsDisabled
-        });
-        state.onReceiveTransactions(txsData, callbackData, settings);
+        state.onReceiveTransactions(txsData, callbackData);
     }
 
     function withdrawWhenOffline()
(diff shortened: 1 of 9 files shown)


packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -80,9 +76,9 @@ contract LoopringAmmPool is
     )
     {
         require(_controller != IAmmController(0), "ZERO_ADDRESS");
-        controller = _controller;
-        assetManager = _assetManager;
-        joinsDisabled = _joinsDisabled;
+        state.controller = _controller;
+        state.assetManager = _assetManager;
+        state.joinsDisabled = _joinsDisabled;
```
Brechtpd, line 81: Because we always use the pools after a proxy, this will set the data inside the implementation contract (which will never get used) and so these values will still be `0` when actually used.

But do we want to set this data for each pool separately? Doesn't seem that useful to me and increases cost quite a bit both for pool creation (more data to store when creating each pool) and when using the pool (reading this data from storage for each join/exit, currently even sometimes when not used).

My thought process was that most pools will use the same values so may as well just hard-code them in the implementation contract so this data automatically set for each pool for free. They are also constants so free to read. If we would like to update these for the pools we can just deploy a new implementation contract with updated values and update the proxy.

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -205,6 +205,7 @@ library AmmExitProcess
         uint newTotalSupply = ctx.totalSupply.sub(exit.burnAmount);
         for (uint i = 0; i < ctx.tokens.length; i++) {
             ctx.vTokenBalancesL2[i] = (uint(ctx.vTokenBalancesL2[i]).mul(newTotalSupply) / ctx.totalSupply).toUint96();
+            require(ctx.vTokenBalancesL2[i] > 0, "ZERO_VIRTUAL_BALANCE");
```
Brechtpd, line 208: If the pool is empty after the exit, pool supply will be `0` so the virtual balances will also be `0`.

dong77: I see

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
dong77: Alright.

PR #2485, opened 2021-06-16
[AMM] improve virtual balances update check

dong77:


packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
```diff
@@ -41,6 +41,7 @@ library AmmJoinProcess
     {
         require(!ctx.settings.joinsDisabled, "JOINS_DISABLED");
         require(join.validUntil >= block.timestamp, "EXPIRED");
+        // require(!S.exitMode, "JOIN_DISALLOWED_IN_EXIT_MODE")
```
dong77, line 44: I think skipping the check should be fine?

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
```diff
@@ -51,6 +51,9 @@ library AmmExitProcess
         if (signature.length == 0) {
             if (S.exitMode) {
                 require(exit.fee == 0, "INVALID_FEE");
+                // for (uint i = 0; i < S.tokens.length; i++) {
```
dong77, line 54: Do we need this?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
```diff
@@ -212,8 +212,10 @@ contract LoopringAmmPool is
         )
         external
         nonReentrant
+        onlyWhenOnline
         onlyFromAssetManager
     {
+        require(!state.exitMode, "DISALLOWED_IN_EXIT_MODE");
```
dong77, line 218: do we need these changes？

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawProcess.sol
```diff
@@ -27,11 +27,14 @@ library AmmWithdrawProcess
...
         require(ctx.settings.assetManager != IAssetManager(0), "CANNOT_WITHDRAW_FROM_POOL");
         require(poolWithdrawal.amounts.length == ctx.tokens.length, "INVALID_WITHDRAWAL_AMOUNTS");
+        // require(!S.exitMode, "WITHDRAWAL_DISALLOWED_IN_EXIT_MODE")
+
         for (uint i = 0; i < ctx.tokens.length; i++) {
             uint96 amount = poolWithdrawal.amounts[i];
             if (amount > 0) {
                 address token = ctx.tokens[i].addr;
                 verifyWithdrawalTx(ctx, ctx.tokens[i].tokenID, amount);
+                // token.transferOut(..., ...) where is the trasfer out?
```
dong77, line 37: ?

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
```diff
@@ -55,8 +55,17 @@ contract AmplifiedAmmController is IAmmController, Claimable
...
+            return false;
+        }
+
         // Check if a curve change was explicitly authorized
         if (consumeCurveChangeAuthorized(pool)) {
+            for (uint i = 0; i < balances.length; i++) {
+                if (vBalancesNew[i] < balances[i]) {
+                    return false;
+                }
+            }
```
Brechtpd, line 68: Any reason to disallow AF < 1?

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
Brechtpd: Shouldn't be needed because `tokenBalancesL2` contains the `balancesL1` values to calculate the exit amounts so if there are enough funds in the pool on L2 it should work correctly.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
Brechtpd: Yes I think we can skip it. I guess in most cases it's not smart to join a pool in exit mode but it should be valid.

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawProcess.sol
Brechtpd: This code is to authorize the withdrawal from the L2 account owned by the pool, so the actual token transfer will be done on the deposit contract.

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
Brechtpd: I don't think we can do these checks because the pool may be offline while the asset manager still needs to return funds to the pool and so `balancesL1` still needs updating. Same for when the pool is in exit mode, funds still may need to be returned to the pool.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
dong77: In theory we can, but shall we do that?

packages/loopring_v3/contracts/amm/LoopringAmmPool.sol
dong77: OK

packages/loopring_v3/contracts/amm/libamm/AmmExitProcess.sol
dong77: OK

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: OK

packages/loopring_v3/contracts/amm/libamm/AmmWithdrawProcess.sol
dong77: But where did we call the deposit contract to transfers the fund out to the AMM pool contract, I didn't find the code.

Brechtpd: That's done by the withdrawal transaction in the exchange contract like any other withdrawal.

packages/loopring_v3/contracts/amm/AmplifiedAmmController.sol
Brechtpd: With or without the check it kind of depends on us. If we don't want to do it we just don't set any AF < 1.

packages/loopring_v3/contracts/amm/libamm/AmmJoinProcess.sol
dong77: I'll leave the code commented out then.

PR #2500, opened 2021-07-01
[hebaov2]fixes + more tests

kongliangzhong:


packages/hebao_v2/contracts/base/libwallet/GuardianLib.sol
```diff
@@ -42,8 +42,10 @@ library GuardianLib
         external
     {
         address guardian = address(0);
+        address owner = wallet.owner;
         for (uint i = 0; i < _guardians.length; i++) {
             require(_guardians[i] > guardian, "INVALID_ORDERING");
+            require(_guardians[i] != owner, "GUARDIAN_CAN_NOT_BE_OWNER");
```
Brechtpd, line 48: Add this check in `_addGuardian` because otherwise the owner could still be added using one of the other methods to add guardians?

Brechtpd: I guess best to remove the check here now because it's not needed anymore.

kongliangzhong: Yes, It's hard to say which one is the better choice because the relayer should do the check. Do this check in the contract can reduce possibilities so prevent potential bugs.

Brechtpd: Oh yeah I can see the benefit. But I meant that this function also calls `_addGuardian` which will already do the check that the guardian isn't the owner now so doing this check in `addGuardiansImmediately` as well will just end up doing the same check twice. So it would make sense to remove the check here in `addGuardiansImmediately`.

kongliangzhong: Oh, yes, I see.

packages/hebao_v2/contracts/base/libwallet/QuotaLib.sol
```diff
@@ -98,9 +101,9 @@ library QuotaLib
         }
 
         Quota storage quota = wallet.quota;
-        quota.currentQuota = currentQuota(wallet).toUint128();
         quota.pendingQuota = newQuota.toUint128();
         quota.pendingUntil = effectiveTime.toUint64();
+        quota.currentQuota = currentQuota(wallet).toUint128();
```
Brechtpd, line 106: Sorry for asking this just now, I didn't think this change had any effect but on second thought it does change the behavior. 

It does seem like we first have to update `quota.currentQuota` in the case the current `quota.pendingQuota` is the actual current quota:  `q.pendingUntil <= block.timestamp ? q.pendingQuota : q.currentQuota`. So if with the current values `q.pendingUntil <= block.timestamp` is true than we have to first update `quota.currentQuota` because otherwise the correct current quota will be lost after the update.

For example:
- `currentQuota := 10`, `pendingQuota := 20` and `pendingUntil := 1000`.
- `block.timestamp := 2000` so `currentQuota(wallet) := 20`.
- `setQuota` is called with `newQuota := 50` and `effectiveTime := 4000`
- Correct behavior:  `currentQuota := 20`, `pendingQuota := 50`, and `pendingUntil := 4000`
- Behavior after this change: `currentQuota := 10`, `pendingQuota := 50`, and `pendingUntil := 4000`

kongliangzhong: I did this change because when changeDailyQuotaWA is invoked, the currentQuota is expected to change immediately.
This issue is caused by the currentQuota is not updated when block.timestamp >= pendingUntil. So add quota.currentQuota = currentQuota(wallet).toUint128() at the begining of setQuota function can fix it.

Brechtpd: But the quota does change immediately no? `currentQuota()` will immediately return the new quota because `pendingUntil := 0` (`0 <= block.timestamp` so `pendingQuota` is the actual current quota). Is there a reason why you want the state variable `wallet.quota.currentQuota` to also be updated immediately? Like you said, the state variable for `currentQuota` is not updated elsewhere so we cannot depend on it being the actual current quota so we we still need to depend on the `currentQuota()` function everywhere.

I think adding the extra `quota.currentQuota = currentQuota(wallet).toUint128();` is okay, but I think the code works just as well without it so I don't really see the benefit of adding it.

dong77 (approved):

packages/hebao_v2/contracts/base/libwallet/QuotaLib.sol
kongliangzhong: Makes sense

Brechtpd (approved):

PR #2518, opened 2021-07-16
[Hebaov2] metatx transfer test

kongliangzhong:


dong77 (approved):

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -61,7 +61,6 @@ library MetaTxLib
     {
         bytes memory encoded = abi.encode(
             META_TX_TYPEHASH,
-            msg.sender,
```
Brechtpd, line 64: There could be a problem with people front-running the meta transactions because the fee is also sent to `msg.sender`. Anyone could just monitor the mempool and execute the meta txs themselves with slightly higher gas price, receiving all the fees after a bidding way while making the all other txs fail while wasting gas. Or it could also just be the miner that does all the meta txs because it is MEV.

If the relayer doesn't know the future `msg.sender` we'll have to find a way to solve this problem differently ideally without increasing the gas costs. One way would be to make the fee recipient a fixed address in the contract so the value cannot be extracted by someone else. This does however lock the user into using a single relayer service which is probably fine but not ideal from the user's perspective. Perhaps there are better ways to do it.

Brechtpd: Hmm I guess just putting the fee recipient address instead of `msg.sender` in the data signed by the user should work just as well:
```
bytes memory encoded = abi.encode(
    META_TX_TYPEHASH,
    metaTx.feeRecipient,
    metaTx.to,
    metaTx.nonce,
    metaTx.gasToken,
    metaTx.gasPrice,
    metaTx.gasLimit,
    metaTx.gasOverhead,
    metaTx.requiresSuccess,
    keccak256(metaTx.data)
);

ERC20Lib.transfer(metaTx.gasToken, metaTx.feeRecipient, gasCost);
```

kongliangzhong: Great!, better than a fixed address.

kongliangzhong:
@Brechtpd , Add a feeReceipt to MetaTx add extra complexity in the relayer system. The wallets will have to get feeReceipt address from backend service each time they sign a meta tx, so I still prefer msg.sender. If we don't care about MEV, can we just remove it from the signature?

Brechtpd (approved):

PR #2522, opened 2021-07-20
[hebao] Add creation timestamp

Brechtpd:
Small changes to allows supporting Agents that allows L2 authorization without deployment of the wallet. Deployment gas cost remains the same.

The actual Agent logic will be implemented in a separate contract by inheriting from `WalletFactory` instead of doing the logic directly in `WalletFactory` because the Agent logic is much more likely to change more frequently (and there are already changes between 3.6.2 and 3.7/3.8).


packages/hebao_v2/contracts/base/SmartWallet.sol
```diff
@@ -142,6 +146,24 @@ contract SmartWallet is ERC1271
...
+    function getOwner()
+        public
+        view
+        override
+        returns (address)
+    {
+        return wallet.owner;
+    }
+
+    function getCreationTimestamp()
```
dong77, line 158: how about `owner()` and `creationTimestamp()`?

packages/hebao_v2/contracts/lib/EIP712.sol
```diff
@@ -37,15 +38,19 @@ library EIP712
     }
 
     function hashPacked(
-        bytes32 domainSeperator,
-        bytes   memory encodedData
+        bytes32 domainHash,
```
dong77, line 41: maybe stick to `domainSeperator` because when you call this method, you passed in `DOMAIN_SEPARATOR`

packages/hebao_v2/contracts/base/WalletFactory.sol
```diff
@@ -105,27 +107,29 @@ contract WalletFactory
...
-            CREATE_WALLET_TYPEHASH,
-            config.owner,
-            keccak256(abi.encodePacked(config.guardians)),
-            config.quota,
-            config.inheritor,
-            config.feeRecipient,
-            config.feeToken,
-            config.feeAmount,
-            salt
+        bytes32 encodedRequestHash = keccak256(
```
dong77, line 118: replace globally `encodedRequestHash` with `dataHash`?

dong77 (dismissed):

packages/hebao_v2/contracts/base/SmartWallet.sol
Brechtpd: If I do this I get these shadows warning:
```
contracts/iface/ILoopringWalletV2.sol:19:9: Warning: This declaration shadows an existing declaration.
        address             owner,
        ^-----------------------^
contracts/iface/ILoopringWalletV2.sol:40:5: The shadowed declaration is here:
    function owner()
    ^ (Relevant source part starts here and spans across multiple lines).

contracts/base/SmartWallet.sol:110:9: Warning: This declaration shadows an existing declaration.
        address             owner,
        ^-----------------------^
contracts/base/SmartWallet.sol:149:5: The shadowed declaration is here:
    function owner()
    ^ (Relevant source part starts here and spans across multiple lines).
```

Probably because these kinds of getters get automatically generated when there are state variable that are called like that, so even though there is no problem it seems like it's better not to do this to not run into these kinds of automatic checks.

dong77 (approved):

PR #2525, opened 2021-07-22
[discussion only] Is this a bug?

dong77:
I noticed we may have more inconsistency in computing hashes. For example, in MetaTxLib.sol, we have:

```solidity
    function validateMetaTx(
        Wallet  storage wallet,
        bytes32 DOMAIN_SEPARATOR,
        MetaTx  memory  metaTx
        )
        public
        view
        returns (bytes32)
    {
        bytes memory encoded = abi.encode(
            META_TX_TYPEHASH,
            metaTx.to,
            metaTx.nonce,
            metaTx.gasToken,
            metaTx.gasPrice,
            metaTx.gasLimit,
            metaTx.gasOverhead,
            metaTx.feeReceipt,
            metaTx.requiresSuccess,
            keccak256(metaTx.data)
        );
        bytes32 metaTxHash = EIP712.hashPacked(DOMAIN_SEPARATOR, encoded);
        require(metaTxHash.verifySignature(wallet.owner, metaTx.signature), "METATX_INVALID_SIGNATURE");
        return metaTxHash;
    }
```
For other social-approval based transactions, we also do not have an `keccak256` call wrapping `abi.encode`.

Which approach is the correct one?

### DIFF
--- a/packages/hebao_v2/contracts/base/WalletFactory.sol
+++ b/packages/hebao_v2/contracts/base/WalletFactory.sol
@@ -113,19 +113,21 @@ contract WalletFactory
     {
         require(config.owner != address(0), "INVALID_OWNER");
 
-        bytes memory encodedRequest = abi.encode(
-            CREATE_WALLET_TYPEHASH,
-            config.owner,
-            keccak256(abi.encodePacked(config.guardians)),
-            config.quota,
-            config.inheritor,
-            config.feeRecipient,
-            config.feeToken,
-            config.feeAmount,
-            salt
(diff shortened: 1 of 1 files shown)


Brechtpd:
Essentially the same thing, but there was a difference in implementation in `EIP712.sol` between the protocols version and the smart wallet version. I removed the inconsistency in #2522.

dong77:
> Essentially the same thing, but there was a difference in implementation in `EIP712.sol` between the protocols version and the smart wallet version. I removed the inconsistency in #2522.

Thank you. Just noticed the change in EIP712.sol.

PR #2526, opened 2021-07-22
[hebao_v2] Support dataless metaTxs when calling *-WA functions

kongliangzhong:


packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -106,11 +106,15 @@ library MetaTxLib
...
-        );
+        // No need to verify metaTx signature when invoke *-WA functions.
+        bytes32 metaTxHash;
+        if (metaTx.nonce != 0) {
+            metaTxHash = validateMetaTx(
+                wallet,
+                DOMAIN_SEPARATOR,
+                metaTx
+            );
+        }
```
dong77, line 117:
```
bytes32 metaTxHash = metaTx.nonce == 0 ?
    bytes32(0): validateMetaTx(wallet, DOMAIN_SEPARATOR, metaTx);
```

Brechtpd:
I don't think we can do this. The meta-tx signature is necessary to allow the wallet to pay a fee for the meta-tx, and this fee payment is not authorized by the *-WA variants because the meta-tx data is not related to the data necessary to do the actual operation. So not requiring the meta-tx signature would mean any fee could be charged by anyone. There's a good reason not to mix the data signed up for the meta-tx and the operations because they don't really have anything to do with each other and WA functions can be called from other wallets (well any address) as well.

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
kongliangzhong: 后台用metaTxHash作为唯一索引，因此不管是不是要验签这个值都计算出来了。在后面的update中。

kongliangzhong:
This PR is mostly interaction-related. In the App side they don't want to sign a metaTx twice when calling a *-WA function.

kongliangzhong:
@Brechtpd , If we don't care about MEV, can we do this? we don't suppose to make money from this kind of txs, As long as security is guaranteed, anyone can do the job.

kongliangzhong:
OK, I know your point, others can overcharge from the metaTx.

Brechtpd:
Added support for a dataless meta-tx where only the function signature is signed while still allowing no nonce to be used so that meta-tx can be executed whenever necessary even when other meta txs are being done in the meantime (which I assume is also something the relayer wants).

So these dataless meta-txs can only be used when
1. The nonce is 0
2. requiresSuccess is True
3. An operation with the approval of a majority of guardians is called

Point 3 is important. The meta-tx has no real replay protection when the nonce is 0 and not all the calldata is signed. It would be possible to reuse the meta-tx with other calldata, however this is generally not possible because of 3. because to make another successful meta-tx there would be a need for new approved data signed by a majority of guardians (previously used data is already marked as used so cannot be reused). If there is a new approval for the same operation it could be possible to reuse with a previously known meta-tx by replacing the calldata. However, this is now made impossible after some time by introducing a `validUntil` value in the meta-tx that limits how long a meta-tx can be used, so the fee payment for the meta-txs should be close in value which makes this not an attracive thing to do (especially because only the relayer trusted by the user would be able to do it).

Another way replay protection  can be done is of course to use the meta-tx hash as the replay protection method when `nonce == 0`. This requires storing the hash onchain for these kinds of meta-txs so the cost will be higher, but these should be fairly uncommon operations to this should also be fine. I didn't do this to save gas and because we did support this in the V1 contracts once and it was removed so I don't think the relayer team liked this. In any case, if we decide to `validUntil` would still be useful as an easy way to make the meta-tx unique.

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -60,17 +61,22 @@ library MetaTxLib
         view
         returns (bytes32)
     {
+        // If this is a dataless meta-tx the user only signs the function selector,
+        // not the full function calldata.
+        bytes memory data = metaTx.nonce == 0 ? metaTx.data.slice(0, 4) : metaTx.data;
```
dong77, line 66: I still think there is a security issue here. Assuming two different recovery requests have been signed by the majority of all guardians, now the owner doesn't really know which recovery will be executed and it's totally up to the operator to decide what payload should be injected into the meta transaction.

dong77: Can we do this instead:
1. inside xxxWA methods, we return the hash signed by guardians
2. in executeMetaTx, somewhere around `(success, ) = metaTx.to.call{gas: metaTx.gasLimit}(metaTx.data);` we try to extract this **txAwareHash**, if we find one, then we pass it into validateMetaTx and request the owner also sigh the hash together with other fee related parameters. If this hash is not found, we use 0x0.

Brechtpd: I don't believe that's a problem. If the owner wants to pay a certain fee for an operation, than it doesn't really matter if this is done to A, B or both. It's only really about what fee is acceptable by the owner for a certain operation for some period of time, and that's it. And for all these *WA operations (except `recover`, but the guardians can call this directly to get all the funds in the wallet directly) the requirement is `MAJORITY_OWNER_REQUIRED` which means the owner actually does have to approve the use of the meta-tx indirectly as well. So without the owner directly involved it's impossible to actually create new valid transactions using the same meta-tx.

Getting the hash as the return value is a good idea, if you thing this is a problem that's a good cheap solution as well.

packages/hebao_v2/contracts/base/SmartWallet.sol
```diff
@@ -235,8 +238,9 @@ contract SmartWallet is ERC1271
         address           guardian
         )
         external
+        returns (bytes32 approvedHash)
     {
-        wallet.removeGuardianWA(DOMAIN_SEPARATOR, approval, guardian);
+        approvedHash = wallet.removeGuardianWA(DOMAIN_SEPARATOR, approval, guardian);
```
dong77, line 243: the second `resetGuardiansWA` should also return 'approvedHash'?

packages/hebao_v2/contracts/base/libwallet/ERC20Lib.sol
```diff
@@ -115,9 +116,9 @@ library ERC20Lib
         bytes    calldata data
         )
         external
-        returns (bytes memory returnData)
+        returns (bytes32 approvedHash)
```
dong77, line 119: Brecht, I think in callContractWA, we should also return the `returnData` as well:

```
returns (
    bytes32 approvedHash,
    bytes memory returnData
)
```

packages/hebao_v2/contracts/base/libwallet/ERC20Lib.sol
```diff
@@ -212,7 +214,7 @@ library ERC20Lib
...
```
dong77, line 217: Same comment here.

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -154,27 +172,40 @@ library MetaTxLib
...
+        // We don't require any data in the meta tx when
+        // - the meta-tx has no nonce
+        // - the meta-tx needs to be successful
+        // - a function is called that requires a majority of guardians and fails when replayed
+        bytes4 methodId = metaTx.data.toBytes4(0);
+        return metaTx.nonce == 0 &&
+               metaTx.requiresSuccess &&
+               (methodId == SmartWallet.changeMasterCopy.selector ||
+                methodId == SmartWallet.addGuardianWA.selector ||
+                methodId == SmartWallet.removeGuardianWA.selector ||
```
dong77, line 201: `resetGuardiansWA` missing from the list

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
```diff
@@ -88,28 +101,33 @@ library MetaTxLib
...
+        require(isNonceValid(wallet, metaTx), "INVALID_NONCE");
         if (metaTx.nonce != 0) {
             wallet.nonce = metaTx.nonce;
-        } else {
-            require(metaTx.requiresSuccess, "META_TX_WITHOUT_NONCE_REQUIRES_SUCCESS");
         }
 
-        (success, ) = metaTx.to.call{gas: metaTx.gasLimit}(metaTx.data);
+        // Check if the meta-tx is still valid
+        require(block.timestamp <= metaTx.validUntil, "METATX_EXPIRED");
```
dong77, line 117: move this to the top to save gas in case this is false.

dong77 (changes requested):

packages/hebao_v2/contracts/base/libwallet/MetaTxLib.sol
Brechtpd: This is why I don't like these kinds of lists, so easy to forget to update them. And everything would work fine without this list...

dong77 (approved):

PR #2533, opened 2021-07-27
[hebaoV2] WalletFactory refactor

Brechtpd:


packages/hebao_v2/contracts/base/WalletDeployer.sol
```diff
@@ -0,0 +1,82 @@
...
+        returns (address)
+    {
+        return Create2.computeAddress(
+            _computeWalletSalt(owner, salt),
+            _getWalletCode(),
+            deployer
+        );
+    }
+
+    function _getWalletCode()
```
kongliangzhong, line 55: Can we make this public? we want this code to compute wallet address off-chain (in a script, without query the contract).

Brechtpd: Done.

kongliangzhong (approved):

PR #2540, opened 2021-07-29
[hebaov2] Allow operator to charge less when creating a new wallet

kongliangzhong:


packages/hebao_v2/contracts/base/WalletFactory.sol
```diff
@@ -32,6 +32,7 @@ contract WalletFactory is WalletDeployer
         address   inheritor;
         address   feeRecipient;
         address   feeToken;
+        uint      maxFeeAmount;
```
dong77, line 35: I suggest that we remove `feeAmount` from this struct. And pass in feeAmount to `createWallet` as a seperate parameter directly:

```solidity
 function createWallet(
        WalletConfig calldata config,
        uint                  salt,
        uint                  feeAmount
        )

```

dong77 (changes requested):

Brechtpd:
@dong77 I was wrong about the front-running/MEV risk. The `feeRecipient` is signed by the wallet owner so there is no reason why anyone else would want to deploy the wallet instead (even with the fee always set to maxFee).

Brechtpd (dismissed):

packages/hebao_v2/contracts/base/WalletFactory.sol
```diff
@@ -96,20 +98,21 @@ contract WalletFactory is WalletDeployer
...
     function _validateRequest(
         WalletConfig calldata config,
-        uint                  salt
+        uint feeAmount
         )
         private
         view
     {
         require(config.owner != address(0), "INVALID_OWNER");
+        require(feeAmount <= config.maxFeeAmount, "INVALID_FEE_AMOUNT");
```
dong77, line 115: Please move this check out of this function so we can remove  the`feeAmount` parameter.

dong77 (approved):

PR #2561, opened 2021-08-19
Loopring 3.9.0 with NFT support

dong77:
Same as https://github.com/Loopring/protocols-internal/tree/master-public


Brechtpd:
Quite some changes were still missing, so I added the missing changes and verified that the PR diff is as expected against the internal repo PR.

Brechtpd (approved):

Issue #2564, opened 2021-08-26
[hebao v2] encode initial wallet config into counterfactual wallet address


dong77:
Currently, we only use the initial owner and salt as the input together with the wallet contract code to generate the wallet's address. I think we should improve it by hashing the entire wallet configuration into the address, especially initial guardians.

First of all, it's not hard to do at all. Secondly, this can prevent malicious operators, Loopring included, from colluding with a hacker who gained access to the initial owner's private keys. With this change, they cannot change the initial guardians.

PR #2565, opened 2021-08-26
[hebaoV2] Hash more data in wallet address

Brechtpd:
See #2564 for discussion if this is what we want or not.

Have left the fee payment options in the signature because at wallet creation the owner won't always know what funds will be in the wallet and the gas fee when the wallet for deploying the wallet can also vary a lot over the months/years. Putting the fee data in the wallet address would of course also be possible.

Have not yet updated the tests (or agent dependencies).


kongliangzhong:
Close this PR because it's not necessary.

PR #2579, opened 2021-11-23
Counterfactual NFTs

Brechtpd:
Adds the counterfactual NFT contracts in this repo.

These contracts get their own subfolder because they are independent but I also made simplified test versions directly in the protocol subfolder for ease of testing (and to minimize version clashing because the `pragma` mode doesn't seem supported with our currently used packages in the protocols subfolder).

### DIFF
--- a/packages/counterfactual_nft/README.md
+++ b/packages/counterfactual_nft/README.md
@@ -25,6 +25,6 @@ Should be simple to understand for users:
 
 The NFT is now fully functional on L2 and the only cost was the NFT mint price on L2.
 
-If `baseURI != 0` it's a bit more complicated because that data is not publicly known by default before the NFT contract is created. So to be able to know how to show the NFT on L2 this data need to be made public somehow (well, at least for them to work on a general block explorer, could be part of the offchain data field in a block, but this of course also costs some gas). But the option is there if minters don't want to use the IPFS hash as the NFT ID and still have the full flexibility to store whatever data they want in the NFT ID like they can in normal NFT contracts.
+If `baseURI != 0` it's a bit more complicated because that data is not publicly known by default before the NFT contract is created. So to be able to know how to show the NFT on L2 this data needs to be made public somehow (well, at least for them to work on a general block explorer, could be part of the offchain data field in a block, but this of course also costs some gas). But the option is there if minters don't want to use the IPFS hash as the NFT ID and still have the full flexibility to store whatever data they want in the NFT ID like they can in normal NFT contracts.
 
 This is really just mainly for people that want to mint a couple of NFTs for as cheap as possible. Creators doing serious collections with thousands of NFTs will probably want to use their own contract anyway.
--- a/packages/counterfactual_nft/README.md
+++ b/packages/counterfactual_nft/README.md
@@ -1,0 +1,30 @@
+# CounterfactualNFT
+
+## Build instructions
+
+```
+yarn install
+yarn compile
+```
+
+## Test instructions
+
+```
+yarn run ganache-cli
+yarn test
+```
+
+## Usage
+
+Everybody on L2 automatically has his/her own NFT contract for free that can be used to mint NFTs. The address of this NFT contract is `NFTFactory.computeNftContractAddress(owner, baseURI)`. In the simplest case `baseURI == ""` so there is no data that needs to be shared publicly to be able to create the NFT contract by calling `NFTFactory.createNftContract(owner, baseURI)`. **Anybody** can call this function to create an NFT contract for an account. When this contract is created is not important, it just needs to be created when somebody wants to withdraw an NFT associated with this counterfactual NFT contract to L1 because it acts like a bridge between L1 and L2.
+
+Should be simple to understand for users:
+- The user puts his NFT on IPFS and gets an IPFS hash in return (per NFT!)
+- The UI calculates the counterfactual NFT token contract address and shows it as the default option to mint NFTs to for his account
+- The NFT is minted with the counterfactual NFT token contract as the token address and the IPFS hash as the NFT ID
+
+The NFT is now fully functional on L2 and the only cost was the NFT mint price on L2.
+
+If `baseURI != 0` it's a bit more complicated because that data is not publicly known by default before the NFT contract is created. So to be able to know how to show the NFT on L2 this data need to be made public somehow (well, at least for them to work on a general block explorer, could be part of the offchain data field in a block, but this of course also costs some gas). But the option is there if minters don't want to use the IPFS hash as the NFT ID and still have the full flexibility to store whatever data they want in the NFT ID like they can in normal NFT contracts.
+
+This is really just mainly for people that want to mint a couple of NFTs for as cheap as possible. Creators doing serious collections with thousands of NFTs will probably want to use their own contract anyway.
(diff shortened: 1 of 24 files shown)


packages/counterfactual_nft/migrations/2_deploy_contracts.js
```diff
@@ -0,0 +1,26 @@
...
+  const loopringExchangeAddress = "0x0BABA1Ad5bE3a5C0a66E7ac838a129Bf948f1eA4";
+
+  let owner;
+  if (network === 'rinkeby') {
+    owner = addresses[0];
+  } else if (network === 'development') {
+    owner = addresses[0];
+  } else {
+    owner = addresses[0];
+  }
```
isaacaskew, line 16: `owner` is always the same?

packages/counterfactual_nft/contracts/AddressSet.sol
```diff
@@ -0,0 +1,96 @@
...
+        bytes32 key,
+        address addr,
+        bool    maintainList
+        ) internal
+    {
+        Set storage set = sets[key];
+        require(set.positions[addr] == 0, "ALREADY_IN_SET");
+
+        if (maintainList) {
+            require(set.addresses.length == set.count, "PREVIOUSLY_NOT_MAINTAILED");
```
isaacaskew, line 28: `PREVIOUSLY_NOT_MAINTAILED` -> `PREVIOUSLY_NOT_MAINTAINED`

packages/counterfactual_nft/migrations/2_deploy_contracts.js
Brechtpd: Ah good point. Left it there to make it easier to add different addresses again on different networks but that was for the deployment of different contracts and it's not even useful here. Removed, thanks!

dong77 (approved):

kongliangzhong (approved):

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
```diff
@@ -0,0 +1,202 @@
...
+        if (bytes(_uri).length != 0) {
+            _setURI(_uri);
+        }
+    }
+
+    function setURI(string memory _uri)
+        public
+        onlyOwner
+    {
+        _setURI(_uri);
```
kongliangzhong, line 70: I think the _uri should be immutable, If some NFTs had been minted in the IPFS way, and the owner changed the _uri, their metadata will be lost.

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    function mint(
```
kongliangzhong, line 73: If baseURI is not empty, we should not allow people to mint from L1, and only one NFT will be allowed to mint on L2, 
because they share the same metadata uri

packages/counterfactual_nft/contracts/external/IPFS.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+pragma experimental ABIEncoderV2;
+
+
+/// @title IPFS
+/// @author Brecht Devos - <brecht@loopring.org>
+library IPFS
+{
+    bytes constant ALPHABET = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';
```
paaskus, line 10: I found a few missing characters in the alphabet (uppercase `I`, uppercase `O`, lowercase `l`). Is that on purpose?
```suggestion
    bytes constant ALPHABET = '123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
```

PostCapone: Makes sense in reducing common typographical errors, if human transcription were an expected method of input

packages/counterfactual_nft/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Should be simple to understand for users:
+- The user puts his NFT on IPFS and gets an IPFS hash in return (per NFT!)
+- The UI calculates the counterfactual NFT token contract address and shows it as the default option to mint NFTs to for his account
+- The NFT is minted with the counterfactual NFT token contract as the token address and the IPFS hash as the NFT ID
+
+The NFT is now fully functional on L2 and the only cost was the NFT mint price on L2.
+
+If `baseURI != 0` it's a bit more complicated because that data is not publicly known by default before the NFT contract is created. So to be able to know how to show the NFT on L2 this data need to be made public somehow (well, at least for them to work on a general block explorer, could be part of the offchain data field in a block, but this of course also costs some gas). But the option is there if minters don't want to use the IPFS hash as the NFT ID and still have the full flexibility to store whatever data they want in the NFT ID like they can in normal NFT contracts.
```
sudoforge, line 28:
```suggestion
If `baseURI != 0` it's a bit more complicated because that data is not publicly known by default before the NFT contract is created. So to be able to know how to show the NFT on L2 this data needs to be made public somehow (well, at least for them to work on a general block explorer, could be part of the offchain data field in a block, but this of course also costs some gas). But the option is there if minters don't want to use the IPFS hash as the NFT ID and still have the full flexibility to store whatever data they want in the NFT ID like they can in normal NFT contracts.
```

nit; just a small typo.

packages/counterfactual_nft/contracts/external/IPFS.sol
Brechtpd: That is indeed on of the the intended goals of the [base58 alphabet](https://en.bitcoin.it/wiki/Base58Check_encoding) used by IPFS CIDs.

packages/counterfactual_nft/README.md
Brechtpd: Fixed, thanks!

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
kongliangzhong, line 70: should be:
```
if (bytes(_uri).length != 0) {
            _setURI(_uri);
 }
```
because some NFTs maybe have been minted already in the `IPFS way`, so the owner should not overwrite _uri if it is empty.

packages/loopring_v3/contracts/core/impl/libexchange/ExchangeWithdrawals.sol
```diff
@@ -480,31 +480,35 @@ library ExchangeWithdrawals
...
-                from,
-                to,
-                nft.nftType,
-                nft.token,
-                nft.nftID,
-                amount,
-                extraData,
-                gasLimit
-            );
+        if (gasLimit > 0) {
```
dong77, line 483: maybe cleaner as:
```solidity
if (gasLimit == 0) return false;
...

Brechtpd: The function shouldn't immediately return because the checks + event at the end of the function should still be done (why I tend to not do early outs often in functions and just do if/else).

dong77:
@Brechtpd Maybe we should have a release_loopring_3.6.3 branch and merge the new changes to it. All release_loopring_* shall freeze.

dong77 (approved):

PR #2581, opened 2021-12-06
Make quota calculation more robust

dong77:
With the current implementation, when the amount is `~uint(0)`, the calculation in this price oracle will throw and cause the whole transaction to fail.

### DIFF
--- a/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
+++ b/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
@@ -33,7 +33,7 @@ contract UniswapV2PriceOracle is PriceOracle
         override
         returns (uint)
     {
-        if (amount == 0) return 0;
+        if (amount == 0 || amount == ~uint(0)) return 0;
         if (token == address(0) || token == wethAddress) return amount;
 
         address pair = factory.getPair(token, wethAddress);
--- a/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
+++ b/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
@@ -33,7 +33,7 @@ contract UniswapV2PriceOracle is PriceOracle
         override
         returns (uint)
     {
-        if (amount == 0 || amount == ~uint(0)) return 0;
+        if (amount == 0 || amount == type(uint).max) return 0;
         if (token == address(0) || token == wethAddress) return amount;
 
         address pair = factory.getPair(token, wethAddress);
--- a/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
+++ b/packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
@@ -48,9 +48,9 @@ contract UniswapV2PriceOracle is PriceOracle
         }
 
         if (token < wethAddress) {
-            return amount.mul(reserve1 / reserve0);
+            return amount.mul(reserve1.mul(1000) / reserve0) / 1000;
         } else {
-            return amount.mul(reserve0 / reserve1);
+            return amount.mul(reserve0.mul(1000) / reserve1) / 1000;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kongliangzhong:
please merge to release_loopring_3.6.3

kongliangzhong (approved):

kongliangzhong (approved):

packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
```diff
@@ -33,7 +33,7 @@ contract UniswapV2PriceOracle is PriceOracle
         override
         returns (uint)
     {
-        if (amount == 0) return 0;
+        if (amount == 0 || amount == ~uint(0)) return 0;
```
Brechtpd, line 36: Could use `type(uint).max` as well.

Brechtpd:
I don't think returning `0` is the correct behavior, that would mean that the quota wouldn't have any impact on infinite approvals, which means the quota is basically rendered useless (just always approve infinite and the wallet owner can do whatever he wants with all the wallet funds).

Failing here with an overflow seems reasonable enough to me, because the tx should fail regardless so it would just happen somewhere else. If the quota is disabled the price oracle isn't called as far as I can see so the current behavior looks okay to me. If users want to do infinite approvals they have to disable the daily quota, I don't think there's any way around that.

dong77:
> I don't think returning `0` is the correct behavior, that would mean that the quota wouldn't have any impact on infinite approvals, which means the quota is basically rendered useless (just always approve infinite and the wallet owner can do whatever he wants with all the wallet funds).
> 
> Failing here with an overflow seems reasonable enough to me, because the tx should fail regardless so it would just happen somewhere else. If the quota is disabled the price oracle isn't called as far as I can see so the current behavior looks okay to me. If users want to do infinite approvals they have to disable the daily quota, I don't think there's any way around that.

I agree.

packages/hebao_v2/contracts/price/UniswapV2PriceOracle.sol
```diff
@@ -48,9 +49,9 @@ contract UniswapV2PriceOracle is PriceOracle
         }
 
         if (token < wethAddress) {
-            return amount.mul(reserve1) / reserve0;
+            return amount.mul(reserve1.mul(PRICE_SCALE) / reserve0) / PRICE_SCALE;
```
Brechtpd, line 52: Hmmm... Doesn't this behave the same as with without the price scale? (except that this will overflow earlier).

kongliangzhong: How about this:
```
    if (amount > 2 ** 128) {
        return (amount / reserve0).mul(reserve1);
    } else {
        return amount.mul(reserve1) / reserve0; 
    }
```

Brechtpd: Is there a specific problem we're trying to solve with changing the math?  I don't think there's a realistic scenario where `amount > 2 ** 128` that will result in a valid transaction.

kongliangzhong: see this tx: https://dashboard.tenderly.co/tx/mainnet/0x61e413faa4ae2a768d91742aab8106f0b4c424d72c913e792da83cd2befa2584

Brechtpd: But this is with quota enabled right? If so, a failing tx is unavoidable like I explained above: https://github.com/Loopring/protocols/pull/2581#issuecomment-986970889

You can fix the overflow but the tx would just fail at a later point. Infinite approvals cannot be used together with a daily quota. The wallet owner needs to disable the daily quota if he wants to use infinite approvals.

kongliangzhong: Yes, I agree. For some reason, we use the uint.max for all approvals.

PR #2592, opened 2022-01-14
[loopringv3] improve ETH deposit handling logics in DepositContract

kongliangzhong:
ETH not returned to deposit account when msg.value > 0 and deposit amount == 0.

### DIFF
--- a/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
+++ b/packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
@@ -85,23 +85,15 @@ contract DefaultDepositContract is IDepositContract, Claimable
     function deposit(
         address          from,
         address          token,
-        uint96           amount,
+        uint96           amount, // 0-value supported
         bytes   calldata /*extraData*/
         )
         external
         override
         payable
         onlyExchange
-        // ifNotZero(amount)
         returns (uint96 amountReceived)
     {
-        if (amount == 0) {
-            if (msg.value > 0) {
-                from.sendETHAndVerify(msg.value, gasleft());
-            }
-            return 0;
-        }
-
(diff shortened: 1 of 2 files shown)


dong77:
isETHInternal is now called twice. Better to call at most once.

kongliangzhong:
> isETHInternal is now called twice. Better to call at most once.

Yes, it is used twice in the function, but will be called once at most in a tx.

kongliangzhong:
> isETHInternal is now called twice. Better to call at most once.

fixed.

packages/loopring_v3/contracts/core/impl/DefaultDepositContract.sol
```diff
@@ -92,9 +92,16 @@ contract DefaultDepositContract is IDepositContract, Claimable
         override
         payable
         onlyExchange
-        ifNotZero(amount)
+        // ifNotZero(amount)
```
dong77, line 95: Maybe we just need to remove this line without adding the other code. @Brechtpd ?

Brechtpd: Yeah that seems to be sufficient to me.

Also, because we can't update the deposit contract, we can perhaps add a small additional check in `deposit` in `Exchangedeposits.sol`, but perhaps not that important:
```
uint16 tokenID = S.getTokenID(tokenAddress);
if (tokenID == 0 && amount == 0) {
     require(msg.value == 0), "INVALID_ETH_DEPOSIT");
}
```

dong77: I added this logic

dong77 (approved):

Brechtpd (approved):

PR #2620, opened 2022-04-29
[counterfactual_nft] use baseURI as uri prefix

kongliangzhong:


PeterHeng666 (approved):

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
```diff
@@ -136,12 +136,12 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
...
-            // If no base URI is set we interpret the tokenId as an IPFS hash
-            return string(abi.encodePacked("ipfs://", IPFS.encode(tokenId)));
-        } else {
-            return baseURI;
-        }
+
+	if (bytes(baseURI).length == 0 || bytes4("ipfs") == bytes4(bytes(baseURI))) {
+	    return string(abi.encodePacked("ipfs://", IPFS.encode(tokenId)));
+	} else {
+	    return string(abi.encodePacked(baseURI, uint2str(tokenId)));
```
Brechtpd, line 143: What's the benefit of doing this over using `{id}` in the baseURI? Seems less flexible to always have to have it add the end.

kongliangzhong: Seems opensea use baseURI as prefix, so we choose the same way.

kongliangzhong: In this way the actual URL is returned from the contract directly, no further processing is needed.

Brechtpd: It does make it impossible to do something like having the IPFS directory as the base URI and then still nicely point to normal filenames as `{id}.json`. Not the biggest downside but just a bit cleaner I guess.

`{id}` is part of EIP-1155 so everybody already has to support it and do any processing on the returned URI when necessary so we may as well use it as well I think.

kongliangzhong: We already support {id} in the old version. we can use both.

PR #2627, opened 2022-05-24
[hebao-v2] Support signature without signatureType

kongliangzhong:


packages/hebao_v2/contracts/lib/SignatureUtil.sol
```diff
@@ -115,31 +115,29 @@ library SignatureUtil
...
-            success = false;
-        }
-
-        // Restore the signature length
-        assembly {
-            mstore(signature, add(signatureTypeOffset, 1))
         }
 
-        return success;
+	if (signature.length == 66) {
```
Brechtpd, line 135: Length will always be `65` here I think, because `mstore(signature, 65)` forces `signature.length` to `65`.

packages/hebao_v2/contracts/lib/SignatureUtil.sol
```diff
@@ -115,31 +115,29 @@ library SignatureUtil
...
+	    // Strip off the last byte of the signature by updating the length
+	    assembly {
+		mstore(signature, 65)
+	    }
+	}
 
-        if (signatureType == SignatureType.EIP_712) {
-            success = (signer == recoverECDSASigner(signHash, signature));
-        } else if (signatureType == SignatureType.ETH_SIGN) {
+	success = (signer == recoverECDSASigner(signHash, signature));
```
Brechtpd, line 127: This does increase the gas cost of verifying `ETH_SIGN` signatures by ~3000 gas if each time it needs to be checked if it's not a valid `EIP_712` signature. I guess not too bad if most times signatures are `EIP_712`.

Brechtpd (dismissed):

packages/hebao_v2/contracts/lib/ReentrancyGuard.sol
```diff
@@ -16,9 +16,9 @@ contract ReentrancyGuard
 
     modifier nonReentrant()
     {
-        require(_guardValue == 0, "REENTRANCY");
-        _guardValue = 1;
+        require(_guardValue == 1, "REENTRANCY");
```
dong77, line 19: where is _guardValue initialized to 1?

dong77 (dismissed):

packages/hebao_v2/contracts/lib/ReentrancyGuard.sol
```diff
@@ -11,14 +11,14 @@ pragma solidity ^0.7.0;
 ///      the number of times the value is changed
 contract ReentrancyGuard
 {
-    //The default value must be 0 in order to work behind a proxy.
-    uint private _guardValue;
+    //The default value must be 1 in order to work behind a proxy.
+    uint private _guardValue = 1;
```
Brechtpd, line 15: I don't think this will work because the storage variable in the proxy contract will not be initialized to 1, only the storage variable in the implementation contract will have it set but that won't be used.

kongliangzhong: Sorry, I didn't pay enough attention to the comment line above. all changes in ReentrancyGuard had been reverted.

dong77 (approved):

Brechtpd (approved):

PR #2629, opened 2022-05-31
[counterfactual_nft] Disable minting on L1 and add creator

kongliangzhong:
Changes:
1. disable minting on L1
2. add creator info for token id (in order to keep compatible with opensea)

### DIFF
--- a/packages/counterfactual_nft/contracts/CounterfactualNFT.sol
+++ b/packages/counterfactual_nft/contracts/CounterfactualNFT.sol
@@ -11,7 +11,6 @@ import "./AddressSet.sol";
 import "./external/IL2MintableNFT.sol";
 import "./external/IPFS.sol";
 import "./external/OwnableUpgradeable.sol";
-import "./WithCreator.sol";
 
 
 /**
@@ -29,7 +28,6 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
     bytes32 internal constant MINTERS = keccak256("__MINTERS__");
     bytes32 internal constant DEPRECATED_MINTERS = keccak256("__DEPRECATED_MINTERS__");
     address public immutable layer2Address;
-
 
     modifier onlyFromLayer2
     {
@@ -180,7 +178,6 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
         for (uint i = 0; i < minterAddresses.length; i++) {
             mintersAndOwner[idx++] = minterAddresses[i];
         }
-
         for (uint i = 0; i < deprecatedAddresses.length; i++) {
             mintersAndOwner[idx++] = deprecatedAddresses[i];
         }
@@ -197,7 +194,6 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
         // Also allow the owner to mint NFTs to save on gas (no additional minter needs to be set)
         return addr == owner() || isAddressInSet(MINTERS, addr);
     }
-
 
     function uint2str(uint _i) internal pure returns (string memory _uintAsString) {
         if (_i == 0) {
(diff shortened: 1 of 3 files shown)


packages/counterfactual_nft/contracts/CounterfactualNFT.sol
```diff
@@ -69,29 +75,30 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
...
-        uint256       id,
-        uint256       amount,
-        bytes  memory data
-        )
-        external
-        onlyFromMinter
-    {
-        _mint(account, id, amount, data);
-    }
+    /// update at 2022-05-31: Do not allow minting on Layer 1
```
Brechtpd, line 78: Any particular reason to disable minting on L1?

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
```diff
@@ -159,6 +166,7 @@ contract CounterfactualNFT is ICounterfactualNFT, Initializable, ERC1155Upgradea
     {
         require(isMinter(minter) || isAddressInSet(DEPRECATED_MINTERS, minter), "invalid minter");
 
+        creators[id] = minter; // set minter as creator
```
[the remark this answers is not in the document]
kongliangzhong: we will make sure each tokenId can be mint only once on L2, so we disabled minting on L1. creator is useful when users deposit their NFT to opensea, the creator can edit collection info in opensea.

packages/counterfactual_nft/contracts/CounterfactualNFT.sol
kongliangzhong: > Any particular reason to disable minting on L1?

This functionality is never used and is not expected to be used in the future.

dong77: Is there any downside of keeping it?

kongliangzhong: > 

The backend will make sure a tokenId can be minted only once, this method can break it.

Brechtpd:
Looks good to me. But maybe best if we move this to a different contract instead of modifying the current general purpose `CounterfactualNFT` contract? The new logic is kind of special and isn't really a general NFT contract anymore so a new separate contract living next to the current one makes sense to me.

kongliangzhong:
> Looks good to me. But maybe best if we move this to a different contract instead of modifying the current general purpose `CounterfactualNFT` contract? The new logic is kind of special and isn't really a general NFT contract anymore so a new separate contract living next to the current one makes sense to me.

You mean a contract that extends CounterfactualNFT contract?

Brechtpd:
Yeah I think that should work.

kongliangzhong:
Ok, makes sence.

packages/counterfactual_nft/contracts/WithCreator.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2017 Loopring Technology Limited.
+pragma solidity ^0.8.2;
+
+
+/// @title AddressSet
+/// @author Freeman Zhong
+contract WithCreator {
```
Brechtpd, line 8: Instead of using this to inherit from in `CounterfactualNFT` I think it makes more sense to inherit `CounterfactualNFT` on this contract. You can disable minting here by overriding the mint function and just throw an error. You can also override `mintFromL2` to call `_setCreator` and then just pass it through to `CounterfactualNFT.mintFromL2` (which you'll have to change from `external` to `public` I think to be able to do so). This way `CounterfactualNFT` is left unchanged and still implements the default NFT logic (which includes minting on L1).

Brechtpd (approved):

PR #2634, opened 2022-06-06
[counterfactual-nft] Allow open minting mode

kongliangzhong:


packages/counterfactual_nft/contracts/CounterfactualNftExt.sol
```diff
@@ -51,7 +54,9 @@ contract CounterfactualNftExt is WithCreator, CounterfactualNFT
         override
         onlyFromLayer2
     {
-        require(isMinter(minter) || isAddressInSet(DEPRECATED_MINTERS, minter), "invalid minter");
+        if (!openMinting) {
+            require(isMinter(minter) || isAddressInSet(DEPRECATED_MINTERS, minter), "invalid minter");
```
Brechtpd, line 58: Maybe instead `isMinter` can be overridden to always return `true` when open minting is enabled, and else pass through to `super.isMinter()`? This way the existing public functions will return the correct info. Well, `minters()` is still a bit unclear. Maybe we can return `0x0` when open minting is enabled as a way to differentiate from closed minter behavior.

kongliangzhong: In open minting mode, we don't care about isMinter and minters, so I think it is fine to keep them unchanged.

Brechtpd: We don't care directly, but when writing things that integrate these NFT contracts it's still nice if things work as expected on the public functions. For example, the explorer shows all approved minters and can show if an NFT was minted by an approved minter. For this to work correctly these functions need to return data that is correct/usable.

kongliangzhong: OK, that makes sence.

PR #2635, opened 2022-06-07
[counterfactualNFT] Add contractURI support

kongliangzhong:


packages/counterfactual_nft/contracts/CounterfactualNftExt.sol
```diff
@@ -63,6 +63,20 @@ contract CounterfactualNftExt is WithCreator, CounterfactualNFT
         emit MintFromL2(to, id, amount, minter);
     }
 
+    function setContractURI(string memory contractURI_)
+        external
+        override
+    {
+        if (bytes(contractURI_).length > 0) {
+            if (_owner == address(0)) {             // every on can set when owner not initialized
```
Brechtpd, line 71: Can you explain a bit more how this would be used? Seems a bit strange to allow anyone to set this value.

packages/counterfactual_nft/contracts/NFTFactory.sol
```diff
@@ -44,9 +46,10 @@ contract NFTFactory
         );
 
         // Initialize
+        IOpenseaSupport(nftContract).setContractURI(contractURI);
         ICounterfactualNFT(nftContract).initialize(owner, baseURI);
```
Brechtpd, line 50: Maybe better to add `contractURI` to `initialize` as well so all values are set together?

packages/counterfactual_nft/contracts/CounterfactualNftExt.sol
kongliangzhong: Allow NFTFactory to set this value before the owner is set. After owner is set, only owner can set this value.

packages/counterfactual_nft/contracts/NFTFactory.sol
kongliangzhong: Because I don't want to change the behavior of CounterfactualNFT, I added the field contractURI to OpenseaSupport contract, otherwise I will have to add it to CounterfactualNFT

packages/counterfactual_nft/contracts/CounterfactualNftExt.sol
```diff
@@ -63,6 +63,20 @@ contract CounterfactualNftExt is WithCreator, CounterfactualNFT
         emit MintFromL2(to, id, amount, minter);
     }
 
+    function setContractURI(string memory contractURI_)
+        external
+        override
+    {
+        if (bytes(contractURI_).length > 0) {
```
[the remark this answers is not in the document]
kongliangzhong: Yes, makes sense.

packages/counterfactual_nft/contracts/NFTFactory.sol
Brechtpd: Fair enough, though you could also just override `initialize` in `CounterfactualNftExt` which may still be a bit cleaner so there's no special logic with the owner in `setContractURI`. But I'm okay with how it is now. If you keep the different calls a small gas cost improvement is probably to only call `setContractURI` if `contractURI.length > 0`.

packages/counterfactual_nft/contracts/NFTFactory.sol
```diff
@@ -30,7 +31,8 @@ contract NFTFactory
     /// @return nftContract The new NFT contract address
     function createNftContract(
         address            owner,
-        string    calldata baseURI
+        string    calldata baseURI,
+        string    calldata contractURI
```
Brechtpd, line 35: Oh didn't think of this yesterday but `contractURI` isn't part of the counterfactual address data. This means that whoever creates the contract can set it to whatever value he wants. Maybe not the biggest issue because the owner can change it back afterwards but still seems like something that shouldn't be possible. 

Easiest solution here would be to make it part of the counterfactual address but I think that probably doesn't make that much sense. Some other possible solutions:
1. if `contractURI.length > 0` an additional signature is required signing that data from `owner`
2. Simple don't allow setting `contractURI` in `createNftContract`, the owner simply has to do it aftewards.

Of these solutions (maybe there are better ones) I would prefer 2 because it's the easiest. It would also mean we could simply add `contractURI` support in the base `CounterfactualNFT` NFT contract because it just adds some general functionality that doesn't change anything.

Brechtpd: I personally don't think it's problem because it can be a very bad thing for content creators. Imagine an attacker deploying the contract of somebody's NFTs with NSFW/illegal content in the contract metadata. Because this data will be shown on the Opensea store front and center it could get the content creator in trouble or maybe even could get banned from market places.

The fact that the owner can change it afterwards I think is not good enough because
- the damage is already done (bad first impression/banned/...)
- now the owner of an undeployed counterfactual NFT contract needs to monitor the contract to be sure no harmful content is shown which is not realistic

Because like you said this is just a convenient way to set the value and is not important at all I would definitely choose not to make it possible because the change of potential problems for content creators and even users is just too big.

One other possible solution that still allows setting everything together is to do something like this:
```
// ...
ICounterfactualNFT(nftContract).initialize(owner, baseURI);
if(contractURI.length > 0) {
     IOpenseaSupport(nftContract).setContractURI(contractURI);
}

// ...
function setContractURI(string memory contractURI_)
        external
        override
{
     require(_owner == msg.sender, "OWNER_REQUIRED");
      _setContractURI(contractURI_);
}
````

This will only allow setting the `contractURI` if the person deploying the contract is also the owner. If the deployer is not the owner then it cannot be set and so it cannot be abused.

Comment (Brechtpd) [reply packages/counterfactual_nft/contracts/NFTFactory.sol:35]: Oh wait `require(_owner == msg.sender, "OWNER_REQUIRED");` in `setContractURI` won't work of course when called from the factory contract... I guess it needs to be done like this for example to keep things easy: 
```
if(contractURI.length > 0) {
     require(owner == msg.sender, "OWNER_REQUIRED");
     IOpenseaSupport(nftContract).setContractURI(contractURI);
}
ICounterfactualNFT(nftContract).initialize(owner, baseURI);

// setContractURI remains unchanged
```

kongliangzhong: The NFT contracts are always deployed by one of Loopring's operator account, so this won't be work.

kongliangzhong: Maybe we should add a signature of the owner to the parameter list and validate it in NFTFactory.createNftContract.

kongliangzhong: We'd like to add a contractURI shared across all Loopring's NFT contracts, so I will return a default nftContract value if it is not set. and the owner can still change it afterward.

Brechtpd (approved):

packages/counterfactual_nft/contracts/NFTFactory.sol
Brechtpd: > The NFT contracts are always deployed by one of Loopring's operator accounts, so this won't work.

Everybody can deploy the contracts right? There's not check on a deployer address onchain (which wouldn't work because we need everybody to be able to deploy the contract at any time when necessary).

> Maybe we should add a signature of the owner to the parameter list and validate it in NFTFactory.createNftContract. But others can get the bytecode of NFTContract directly and deploy it. So still does not work.

Not sure what you mean with that? Only the NFTFactory contract will deploy the NFT contract to the expected contract address.

kirataik: >  Not sure what you mean with that? Only the NFTFactory contract will deploy the NFT contract to the expected contract address.

Ah yes, I thought anyone can deploy to the same contract address.

kirataik: But the latest approach should be enough for our requirements.

kongliangzhong: > Not sure what you mean with that? Only the NFTFactory contract will deploy the NFT contract to the expected contract address.

Ah, sorry, I thought from different addresses can deploy to the same contract address.

PR #2652, opened 2022-09-10
[protocol] update DestroyableWalletAgent.sol

kongliangzhong:


Brechtpd:
Any reason for this change? A `DestroyableWallet` shouldn't really have a deployed wallet at any time because then it would keep being able to authorize transactions on its own even if the wallet is marked as destroyed (bypassing the correct checks here).

kongliangzhong:
DestroyableWalletAgent was used as LoopringWalletAgent. I think this is wrong, we should use LoopringWalletAgent directly.

PR #2735, opened 2024-09-11
add delayed upgrade

kl456123:


packages/loopring_v3/contracts/core/impl/BlockVerifier.sol
```diff
@@ -77,6 +80,25 @@ contract BlockVerifier is ReentrancyGuard, IBlockVerifier
...
+        require(circuit.enabled == false, "ALREADY_ENABLED");
+        require(
+            block.timestamp >= circuit.registeredTime + CIRCUIT_DELAYED_EFFECT,
+            "NOT_EFFECT_YET"
+        );
+
+        // Enable the circuit
+        circuit.enabled = true;
+
+        emit CircuitDisabled(blockType, blockSize, blockVersion);
```
PeterHeng666, line 99: this is enabled ，can add an event

packages/loopring_v3/contracts/core/impl/LoopringV3.sol
```diff
@@ -159,17 +160,50 @@ contract LoopringV3 is ILoopringV3, ReentrancyGuard
...
+            // delayed effect
+            cachedSettings = CachedSettings({
+                protocolFeeVault: _protocolFeeVault,
+                blockVerifierAddress: _blockVerifierAddress,
+                forcedWithdrawalFee: _forcedWithdrawalFee
+            });
+            nextEffectiveTime = _nextEffectiveTime;
+        }
+
+        emit SettingsUpdated(block.timestamp, _nextEffectiveTime);
```
PeterHeng666, line 190: should not modify current event, can add another event which only update but not effect

packages/loopring_v3/contracts/core/iface/IBlockVerifier.sol
```diff
@@ -55,6 +55,16 @@ abstract contract IBlockVerifier is Claimable
         external
         virtual;
 
+    /// @dev Enables the use of the specified circuit.
```
dantaik, line 58:
```suggestion
    /// @dev Enables the use of a specified circuit.
```

packages/loopring_v3/contracts/core/iface/ILoopringV3.sol
```diff
@@ -14,11 +14,14 @@ abstract contract ILoopringV3 is Claimable
...
     event ExchangeStakeBurned(address exchangeAddr, uint amount);
-    event SettingsUpdated(uint time);
+    event SettingsUpdated(uint time, uint effectTime);
+    event SettingsApplied(uint time);
 
     // == Public Variables ==
+    uint256 internal constant SETTING_UPDATE_DELAY = 7 days;
     mapping (address => uint) internal exchangeStake;
 
+    uint public nextEffectiveTime;
```
dantaik, line 24: This will change the data storage layout table and break upgradeability!!!!!

dantaik: Since you have `contract LoopringV3 is ILoopringV3, ReentrancyGuard`, changing any storage layout in IloopringV3 will break upgradabilty.

packages/loopring_v3/contracts/core/impl/BlockVerifier.sol
```diff
@@ -17,11 +17,13 @@ contract BlockVerifier is ReentrancyGuard, IBlockVerifier
     struct Circuit
     {
         bool registered;
+        uint registeredTime;
```
dantaik, line 20: will you deploy a new BlockVerifier or upgrade an existing deployment?

dantaik (changes requested):

packages/loopring_v3/contracts/core/iface/ILoopringV3.sol
kl456123: loopringv3 is not deployed using proxy mode

kl456123: because it has no `initialize` function

packages/loopring_v3/contracts/core/impl/BlockVerifier.sol
kl456123: blockverifier is the same case

dantaik: But you are modifying the code, so what did you mean "is the same"?

dantaik:
@letsgoustc @kl456123 I think "delayed upgrade" is a general feature to all contracts. It's supposed to be contract-agnostic, you probably just need a wrapping "owner contract" that implements a delay internally without any change to the contracts being owned.

dantaik:
I think we should schedule a call for discussion.

kl456123:
> @letsgoustc @kl456123 I think "delayed upgrade" is a general feature to all contracts. It's supposed to be contract-agnostic, you probably just need a wrapping "owner contract" that implements a delay internally without any change to the contracts being owned.

Yes I also think it is absolutely feasible and might be a better way to reach the same goal

packages/loopring_v3/contracts/core/impl/BlockVerifier.sol
kl456123: sorry for my confused statement. I mean that block verifier also cannot be deployed by using proxy mode like in loopringv3 so that only way to upgrade smart contract is to deploy a new one. so the disorder of storage of it doesn't matter.

